PR #45, opened 2025-04-02
XP-1497: Prevent expensive regex generation for the profanity detection function in the toxicity rule

ntatsumi:
Where '*' is occuring, '-' should be substituted for the detect_profanity function in order to prevent very expensive regex processing.


cheriexu (approved):
LGTM!

PR #64, opened 2025-04-16
XP-1543: Create docker-compose that will start both services

Kacpersoftberry:


Makefile
ntatsumi, line 1: This is a good idea but let's wait to introduce it until the facade can provide enough value to be something we want to maintain.

PR #77, opened 2025-04-23
New env vars

jdbarthur:
https://arthurai.atlassian.net/browse/XP-1582


deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -38,6 +39,9 @@ services:
     stop_grace_period: "60s"
     environment:
       - ARTHUR_API_HOST=https://platform.arthur.ai
+      - SHIELD_ENGINE_INTERNAL_API_KEY=${GENAI_ENGINE_ADMIN_KEY}
+      - SHIELD_ENGINE_INTERNAL_HOST=${GENAI_ENGINE_INGRESS_URI}
```
mcgrawia, line 43: why is this not `http://genai-engine`?

deployment/cloudformation/genai-engine/arthur-genai-engine-ecs-task-definition.yml
```diff
@@ -166,6 +166,13 @@ Parameters:
     Type: Number
     Default: 60
     Description: 'Task rules cache TTL in seconds'
+  GenaiEngineAllowAdminKeyGeneralAccess:
+    Type: String
+    Description: 'Allow admin key general access'
```
ntatsumi, line 171: Add: "(Must be enabled for the Arthur Platform. Disabling is recommended for the guardrails-only standalone deployment)"

deployment/cloudformation/ml-engine/arthur-ml-engine-ecs-task-definition.yml
```diff
@@ -47,6 +47,16 @@ Parameters:
     Type: String
     Description: 'The ARN of the Secret containing Client Credentials for the ML Engine'
     NoEcho: true
+  ShieldEngineInternalAPIKeySecretARN:
+    Type: String
+    Description: 'The ARN of the Secret containing the Shield Engine Internal API Key'
+    NoEcho: true
+  ShieldEngineInternalHost:
```
ntatsumi, line 54: Remove this parameter. It will be the same as ShieldEngineInternalIngressHost for ECS deploy.

deployment/cloudformation/ml-engine/arthur-ml-engine-ecs-task-definition.yml
```diff
@@ -90,11 +100,17 @@ Resources:
           Environment:
             - Name: 'ARTHUR_API_HOST'
               Value: !Ref MLEngineArthurApiHost
+            - Name: 'SHIELD_ENGINE_INTERNAL_HOST'
+              Value: !Ref ShieldEngineInternalHost
```
ntatsumi, line 104: !Ref ShieldEngineInternalIngressHost

deployment/cloudformation/root-arthur-engine-cpu.yml
```diff
@@ -724,6 +724,9 @@ Resources:
         ContainerRepositoryCredentialsSecretARN: !GetAtt [ CoreSecretsStack, Outputs.ContainerRepositoryCredentialsSecretOutput ]
         MLEngineArthurApiHost: !Ref ArthurApiHost
         MLEngineClientCredentialsSecretARN: !GetAtt [ MLEngineSecretsStack, Outputs.MLEngineClientCredentialsSecretOutput ]
+        ShieldEngineInternalAPIKeySecretARN: !GetAtt [ GenaiEngineSecretsStack, Outputs.GenaiEngineAPIKeySecretOutput ]
+        ShieldEngineInternalHost: !Ref GenaiEngineIngressURL
```
ntatsumi, line 728: Remove

deployment/cloudformation/root-arthur-engine-gpu.yml
```diff
@@ -882,6 +882,9 @@ Resources:
         ContainerRepositoryCredentialsSecretARN: !GetAtt [ CoreSecretsStack, Outputs.ContainerRepositoryCredentialsSecretOutput ]
         MLEngineArthurApiHost: !Ref ArthurApiHost
         MLEngineClientCredentialsSecretARN: !GetAtt [ MLEngineSecretsStack, Outputs.MLEngineClientCredentialsSecretOutput ]
+        ShieldEngineInternalAPIKeySecretARN: !GetAtt [ GenaiEngineSecretsStack, Outputs.GenaiEngineAPIKeySecretOutput ]
+        ShieldEngineInternalHost: !Ref GenaiEngineIngressURL
```
ntatsumi, line 886: Remove

deployment/docker-compose/arthur-engine/docker-compose.yml
jdbarthur: I was reusing what seems to be a leftover var from our env template. Switched it to the service name.

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -38,6 +39,9 @@ services:
     stop_grace_period: "60s"
     environment:
       - ARTHUR_API_HOST=https://platform.arthur.ai
+      - SHIELD_ENGINE_INTERNAL_API_KEY=${GENAI_ENGINE_ADMIN_KEY}
+      - SHIELD_ENGINE_INTERNAL_HOST=http://genai-engine:3000
+      - SHIELD_ENGINE_INTERNAL_INGRESS_HOST=http://genai-engine:3000
```
mcgrawia, line 44: wait this isn't right, users from outside docker won't be able to hit the shield API on the docker internal address. What port are we exposing the shield API on? The ingress host should be localhost in docker compose

mcgrawia: another question, why are we exposing shield on port 3000? That seems like a very common frontend port, and likely to conflict if users have other things running on their machine. Can we pick another port?

ntatsumi: I think you have to use "host.docker.internal". @jdbarthur is this tested to make sure the host address works from the ml-engine to genai-engine?

deployment/helm/ml-engine/values.schema.json
```diff
@@ -98,6 +98,18 @@
                         "mlEnginePodTolerations": {
                             "type": "object",
                             "description": "Tolerations for the ml engine pods."
+                        },
+                        "shieldEngineInternalAPIKeySecretName": {
+                            "type": "string",
+                            "description": "The name of the secret containing the internal API key for Shield Engine."
+                        },
+                        "shieldEngineInternalHost": {
```
ntatsumi, line 106: We should be able to use Kubernetes DNS instead.

deployment/helm/ml-engine/values.schema.json
```diff
@@ -109,7 +121,10 @@
                         "containerMemoryLimits",
                         "containerCPULimits",
                         "appPlaneUrl",
-                        "secretsName"
+                        "secretsName",
+                        "shieldEngineInternalAPIKeySecretName",
+                        "shieldEngineInternalHost",
```
ntatsumi, line 126: Remove

https://github.com/arthur-ai/arthur-engine/pull/77/files#r2057187067

deployment/docker-compose/arthur-engine/docker-compose.yml
mcgrawia: I don't think we should use `host.docker.internal` as that references the user's localhost outside docker. The internal host should use the docker internal service address for shield, and the external address should be localhost. I don't think there's a need for the data plane to connect outside docker to the users localhost ports IIUC?

deployment/helm/ml-engine/values.schema.json
jdbarthur: I am not understanding this one.

deployment/docker-compose/arthur-engine/docker-compose.yml
jdbarthur: I was using one of the other vars that was already in the template, but it was confusing. 

This is updated now and tested for the separate docker compose up commands in genai-engine, ml-engine, and arthur-engine.

deployment/docker-compose/arthur-engine/.env.template
```diff
@@ -17,8 +17,18 @@ GENAI_ENGINE_ADMIN_KEY=changeme_genai_engine_admin_key
...
+GENAI_ENGINE_INGRESS_URI=http://localhost:3030
+
+# This variable is used by the ml-engine to create genai-engine tasks
+SHIELD_ENGINE_INTERNAL_API_KEY=${GENAI_ENGINE_ADMIN_KEY}
+
+# This variable is used by the ml-engine to create genai-engine tasks
+SHIELD_ENGINE_INTERNAL_HOST=http://genai-engine:3030
+
+# This variable is used from the host to access the genai-engine and /docs
+SHIELD_ENGINE_INTERNAL_INGRESS_HOST=http://localhost:3030
```
mcgrawia, line 29: are these used in the docker compose file? I think most of these are hard coded now

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -21,13 +21,15 @@ services:
     depends_on:
       - db
     ports:
-      - "3000:3000"
+      - "3030:3030"
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_URL=host.docker.internal
```
mcgrawia, line 27:
```suggestion
      - POSTGRES_URL=db
```
this should also use the docker internal address

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -21,13 +21,15 @@ services:
     depends_on:
```
mcgrawia, line 21: I can't seem to leave a comment on line 7, but I don't think we should expose the postgres port on the user's localhost. It's only needed inside docker compose, and exposing it will only add trouble if the user already has another container using that port.

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -37,7 +39,10 @@ services:
     restart: unless-stopped
     stop_grace_period: "60s"
     environment:
-      - ARTHUR_API_HOST=https://platform.arthur.ai
+      - SHIELD_ENGINE_INTERNAL_API_KEY=${GENAI_ENGINE_ADMIN_KEY}
+      - SHIELD_ENGINE_INTERNAL_HOST=http://genai-engine:3030
+      - SHIELD_ENGINE_INTERNAL_INGRESS_HOST=http://localhost:3030
+      - ALLOW_ADMIN_KEY_GENERAL_ACCESS=true
```
mcgrawia, line 45: `ALLOW_ADMIN_KEY_GENERAL_ACCESS` is not needed on the ml engine

deployment/docker-compose/genai-engine/docker-compose.yml
```diff
@@ -21,13 +21,15 @@ services:
     depends_on:
       - db
     ports:
-      - "3000:3000"
+      - "3030:3030"
     environment:
       - POSTGRES_USER=postgres
       - POSTGRES_URL=host.docker.internal
```
mcgrawia, line 27: same comments in this file about making this use the `db` hostname and not exposing the 5432 port outside docker for postgres

deployment/helm/ml-engine/values.schema.json
mcgrawia: this shouldn't be a configured value, it should automatically use the service DNS for the GenAI engine's Kubernetes service

deployment/helm/ml-engine/values.yaml.template
```diff
@@ -25,6 +25,10 @@ mlEngine:
     mlEnginePodAffinity: {}
     mlEnginePodTolerations: {}
 
+    shieldEngineInternalAPIKeySecretName: "genai-engine-secret-admin-key"
+    shieldEngineInternalHost: "arthur-genai-engine-service"
+    shieldEngineInternalIngressHost: ""
```
mcgrawia, line 30: couple comments on this, is this a separate chart just for running the ml-engine by itself? If so, these should not be needed. These only are set when deployed together.

On that note, when deployed together, a couple things:
1. the internal host should not be a config value, it should just use the gan-ai's service address
2. the ingress host should use the same config as the gen-ai engine, it shouldn't have a separate one
3. are the secrets managed outside this helm chart? if so, where is the config value for passing the admin-key to the gen-ai engine? can this use the same config instead of making an ml-engine specific one?

deployment/docker-compose/arthur-engine/.env.template
jdbarthur: You are correct. Thanks.

deployment/docker-compose/arthur-engine/docker-compose.yml
jdbarthur: I thought this was the situation where we needed to use host.docker.internal, but I will test it again with this change.

deployment/docker-compose/arthur-engine/.env.template
```diff
@@ -17,8 +17,12 @@ GENAI_ENGINE_ADMIN_KEY=changeme_genai_engine_admin_key
 APP_SECRET_KEY=changeme_app_secret_key
 
 # GenAI Engine Ingress URL
-GENAI_ENGINE_INGRESS_URI=http://localhost:3000
+GENAI_ENGINE_INGRESS_URI=http://localhost:3030
```
ntatsumi, line 20: This should be set in docker-compose.yml. It's not something you can configure without fiddling the docker-compose.yaml itself.

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -37,7 +39,10 @@ services:
     restart: unless-stopped
     stop_grace_period: "60s"
     environment:
-      - ARTHUR_API_HOST=https://platform.arthur.ai
+      - SHIELD_ENGINE_INTERNAL_API_KEY=${GENAI_ENGINE_ADMIN_KEY}
+      - SHIELD_ENGINE_INTERNAL_HOST=http://genai-engine:3030
+      - SHIELD_ENGINE_INTERNAL_INGRESS_HOST=http://localhost:3030
```
ntatsumi, line 44: @mcgrawia we need to call these envars `GENAI_ENGINE_*`

mcgrawia: is this a requirement or just a convention? we've already got code shipped that uses the existing ones, but can see how much effort it would be to rename

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -21,13 +21,15 @@ services:
     depends_on:
       - db
     ports:
-      - "3000:3000"
+      - "3030:3030"
```
ntatsumi, line 24: I don't think this would work because the container itself doesn't expose this port

ntatsumi (approved):

PR #107, opened 2025-04-29
Postgres deployment optimizations in Docker Compose

ntatsumi:
* Drop pgvector and migrate to using a regular postgres; drop the usage of init-db.sh script since we no longer have to create more than 1 database.
* Provide developers with docker for Postgres since the engine deployment version no longer exposes the port to the host. Also make adminer available for easy GUI access.


genai-engine/docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+      - 5432:5432
+    shm_size: 128mb
+    environment:
+      POSTGRES_PASSWORD: changeme_pg_password
+      POSTGRES_DB: arthur_genai_engine
+
+  adminer:
+    image: adminer
+    ports:
+      - 8080:8080
```
fryz, line 15: @nori-arthur - what is this?

ntatsumi: This docker compose is for the genai engine developers only. It exposes the Postgres port on the host so that your app running on the IDE can connect to it. All engine deployments moved to `/deployment`. Adminer is a DB UI tool for developer productivity.

PR #116, opened 2025-05-01
Simplify the Docker Compose script by optimizing for localhost deploys

ntatsumi:


deployment/docker-compose/arthur-engine/start.sh.template
```diff
@@ -1,9 +1,6 @@
 #!/bin/bash
 
 cat <<EOF > .env
-GENAI_ENGINE_INGRESS_URI=http://localhost:3030
-POSTGRES_PASSWORD=changeme_pg_password
-APP_SECRET_KEY=changeme_app_secret_key
 GENAI_ENGINE_OPENAI_PROVIDER=Azure
 GENAI_ENGINE_OPENAI_GPT_NAMES_ENDPOINTS_KEYS=model_name::https://my_service.openai.azure.com/::my_api_key
 GENAI_ENGINE_ADMIN_KEY=changeme_genai_engine_admin_key
```
fryz, line 6: just a note that we'll probably want to default to OpenAI instead of Azure (esp. in the UI) as this is going to be far more likely the API provider that users will be using

deployment/docker-compose/genai-engine/.env.template
```diff
@@ -1,12 +1,3 @@
-# GenAI Engine Ingress URL (the address of the proxy or the load balancer if you have one)
```
fryz, line 1: Is there a reason to have 2 of the same files for the different engines?

ntatsumi: This one is for the genai-engine stand alone users. They don't need the additional three envars required by ml-engine.

fryz (approved):
Looks like you need to rebase but +1 for now

PR #125, opened 2025-05-06
API host defaults to SaaS in ml-engine

ntatsumi:


mcgrawia:
@ntatsumi LGTM but I don't think I have access to approve

ntatsumi:
> @ntatsumi LGTM but I don't think I have access to approve

@mcgrawia is the "Review changes" button available to you under the "Files changed" tab? Is the approve option greyed out?

mcgrawia (approved):

mcgrawia:
@ntatsumi found it. I was used to GitHub having a review button on this page. Did they get rid of that?

mcgrawia (changes requested):
@ntatsumi testing this with @Pierniki it doesn't seem to work. A couple things we found:

1. it looks like the ml-engine is not using the same network as the other containers, so the connection fails when the ML engine tries to speak to the genAI engine
2. it doesn't look like postgres is using a volume to persist the data, can we add that so it isn't lost on a restart?

mcgrawia:
@ntatsumi another thing we're noticing, the genai-engine takes a few minutes to start because it needs to download the models.. can you add a health check to that service so docker compose waits for it to become healthy before completing? Right now the command finishes, but the `genai-engine` isn't usable for a few minutes, which leads to errors during first model onboarding

ntatsumi:
@mcgrawia Thanks for the very helpful feedback! I've addressed everything except for the health check. Will address in another PR.

PR #146, opened 2025-05-08
XP-1604: Initial commit for Arthur Common module

Kacpersoftberry:
Add Arthur Common module


arthur-common/src/arthur_common/aggregations/functions/README.md
```diff
@@ -0,0 +1,26 @@
...
+| ShieldInferenceRuleToxicityScoreAggregation                                  | 00000000-0000-0000-0000-000000000004 | Toxicity Distribution                                                                   |
+| ShieldInferenceHallucinationCountAggregation                                 | 00000000-0000-0000-0000-000000000003 | Hallucination Count                                                                     |
+| ShieldInferenceRuleCountAggregation                                          | 00000000-0000-0000-0000-000000000002 | Rule Result Count                                                                       |
+| ShieldInferencePassFailCountAggregation                                      | 00000000-0000-0000-0000-000000000001 | Inference Count                                                                         |
+| ShieldInferenceTokenCountAggregation                                         | 00000000-0000-0000-0000-000000000021 | Token Count                                                                             |
```
ntatsumi, line 24: Rename Shield to GenAI Engine

arthur-common/src/arthur_common/models/datasets.py
```diff
@@ -4,7 +4,7 @@
 class ModelProblemType(str, Enum):
     REGRESSION = "regression"
     BINARY_CLASSIFICATION = "binary_classification"
-    ARTHUR_SHIELD = "arthur_shield"
+    ARTHUR_GENAI_ENGINE = "arthur_genai_engine"
```
mcgrawia, line 7: this is a breaking change, these enum values are saved in the BE database, if we're changing the value we need to write a migration to update the state. I think you could just change the name of the ENUM value and leave the value as is more easily?

mcgrawia:
@Kacpersoftberry what's the plan with the unit tests? If I understand correctly, it doesn't seem like they're running on this MR? I think we should make sure the tests are run in CI before merging so we can be sure we haven't broken the package/code from these refactors

arthur-common/src/arthur_common/models/datasets.py
Kacpersoftberry: We want to migrate all shield mention so this is something that we would need to do @fryz how we should proceed with that?

Kacpersoftberry:
> @Kacpersoftberry what's the plan with the unit tests? If I understand correctly, it doesn't seem like they're running on this MR? I think we should make sure the tests are run in CI before merging so we can be sure we haven't broken the package/code from these refactors

We will introduce the pipelines that will do everything in later on PRs.

fryz:
> We will introduce the pipelines that will do everything in later on PRs.

Why not this PR?

arthur-common/src/arthur_common/models/datasets.py
fryz: We should not migrate this - as Ian mentioned, it's an API breaking change and would require a lot more lift. 

For now, you can leave a comment indicating this on this line

arthur-common/src/arthur_common/aggregations/functions/README.md
fryz: CC @Kacpersoftberry - per the convo below, think you should just revert the commit you made to do the rename

arthur-common/src/arthur_common/models/datasets.py
fryz: We should similarly leave comments in all the other files as well where shield is mentioned - we can queue this up for a breaking api change in our backlog but don't think it's super high priority.

.pre-commit-config.yaml
```diff
@@ -65,3 +65,11 @@ repos:
         language: system
         always_run: false
         files: ^genai-engine/src/schemas/
+      - id: arthur-common-pytest-check
+        name: "arthur-common: pytest unit tests and coverage"
+        entry: poetry -C arthur-common run pytest tests/ --cov=arthur_common --cov-fail-under=45
```
mcgrawia, line 70: just for my understanding, does this run the whole unit test suite before every commit?

mcgrawia (approved):
lgtm otherwise

PR #149, opened 2025-05-12
Update CORS to allow for local UI dev + Arthur Platform reqs

fryz:
This allows for the UI to make direct requests to the GenAI Engine (eg: for validating prompts/responses)


genai-engine/src/server.py
```diff
@@ -279,19 +279,23 @@ def get_base_app(
         "http://0.0.0.0:8000",
         "http://localhost:3030",
         "http://localhost:8080",
+        "http://localhost:3023",
```
fryz, line 282: this is the URL when running our FE locally

jdbarthur (approved):

PR #153, opened 2025-05-14
[no-ref] fix docker compose deploy so it starts in daemon mode

mcgrawia:
currently using the https://engine.arthur.ai installation script, the docker images start in the foreground. Adding a the `-d` flag so it runs in the background.


Kacpersoftberry (approved):
LGTM

PR #157, opened 2025-05-15
Metrics management new

videetparekh:
# Spans and Metrics API - Functionality & Endpoints

The Spans and Metrics API provides comprehensive OpenTelemetry trace processing and GenAI metrics computation for LLM application monitoring. It handles protobuf trace ingestion, automatic span extraction with version control, and on-demand metrics evaluation with caching.

## Core Functionality:

**Trace Ingestion**: Processes OpenTelemetry protobuf traces, extracts spans, validates versions, and stores with metadata
**Span Querying**: Flexible filtering by task/trace IDs and time ranges with trace-level pagination
Metrics Computation: On-demand LLM span evaluation with automatic feature extraction (system prompts, user queries, responses)
**Data Integrity**: Version validation, span type checking, and robust error handling

## API Endpoints:
**POST /v1/traces**
Input: Protobuf trace bytes
Output: Processing statistics (total/accepted/rejected spans, status)
Purpose: Ingest OpenInference trace data

**GET /v1/traces/query**
Input: task_ids (required), trace_ids, start_time, end_time, pagination
Output: Spans with existing metrics (no computation)
Purpose: Query spans with filters

**GET /v1/traces/metrics/**
Input: task_ids (required), start_time, end_time, pagination
Output: Spans with computed metrics for LLM spans
Purpose: Query and compute metrics for traces

**GET /v1/span/{span_id}/metrics**
Input: span_id (path parameter)
Output: Single span with computed metrics
Purpose: Compute metrics for specific LLM span


genai-engine/src/scorer/metrics/relevance/relevance.py
```diff
@@ -0,0 +1,354 @@
...
+    pt = PromptTemplate(
+        input_variables=["system_prompt", "user_query", "response"],
+        partial_variables={
+            "format_instructions": parser.get_format_instructions(),
+        },
+        template=RESPONSE_RELEVANCE_PROMPT_TEMPLATE,
+    )
+
+    evaluation_chain = pt | model | parser
+    return evaluation_chain
```
notthattal, line 116: It seems like this is trying to enforce structured outputs. Is it possible to enforce them similar to how the hallucination check is now using structured outputs? Maybe using the check to see if a model supports structured outputs and using this JsonParser if but using the langchain structured outputs if it is?

genai-engine/src/scorer/metrics/tool_selection/tool_selection.py
```diff
@@ -0,0 +1,182 @@
...
+    pt = PromptTemplate(
+        input_variables=["system_prompt", "user_query", "context"],
+        partial_variables={
+            "format_instructions": parser.get_format_instructions(),
+        },
+        template=TOOL_USAGE_PROMPT_TEMPLATE,
+    )
+
+    evaluation_chain = pt | model | parser
+    return evaluation_chain
```
notthattal, line 73: Same here

genai-engine/src/metrics_engine.py
```diff
@@ -0,0 +1,105 @@
...
+# )
+# MAX_HALLUCINATION_TOKEN_LIMIT = int(
+#     get_env_var(constants.GENAI_ENGINE_HALLUCINATION_CHECK_MAX_TOKEN_LIMIT_ENV_VAR),
+# )
+# MAX_TOXICITY_TOKEN_LIMIT = int(
+#     get_env_var(constants.GENAI_ENGINE_TOXICITY_CHECK_MAX_TOKEN_LIMIT_ENV_VAR, True)
+#     or 1200,
+# )
+
+# ZERO_TOKEN_CONSUMPTION = LLMTokenConsumption(prompt_tokens=0, completion_tokens=0)
```
notthattal, line 31: If these aren't being used is it possible to just delete these comments?

genai-engine/src/routers/v2/task_management_routes.py
```diff
@@ -316,14 +376,119 @@ def archive_task_rule(
...
+    try:
+        metric_repo = MetricRepository(db_session)
+        task_repo = TaskRepository(
+            db_session,
+            RuleRepository(db_session),
+            metric_repo,
+            application_config,
+        )
+        task = task_repo.get_task_by_id(str(task_id))
+        metric = Metric._from_request_model(request)
```
fryz, line 410: i see in the comment that only agent tasks can have metrics - should we do the validation here to confirm this?

genai-engine/src/schemas/response_schemas.py
```diff
@@ -419,6 +441,9 @@ class TaskResponse(BaseModel):
     )
     is_agentic: bool = Field(description="Whether the task is agentic or not")
     rules: List[RuleResponse] = Field(description="List of all the rule for the task.")
+    metrics: List[MetricResponse] = Field(
```
fryz, line 444: believe this also needs to be an optional field to maintain backwards compatibility

genai-engine/src/api_changelog.md
```diff
@@ -2,6 +2,18 @@ The intention of this changelog is to document API changes as they happen to eff
 
 ---
 
+# 07/16/2025
+- **CHANGE** for **URL**: /api/v2/tasks  added the required property '/items/metrics' to the response with the '200' status
+- **CHANGE** for **URL**: /api/v2/tasks  added the required property 'metrics' to the response with the '200' status
+- **CHANGE** for **URL**: /api/v2/tasks/search  added the required property 'tasks/items/metrics' to the response with the '200' status
```
fryz, line 8: not 100% sure im following these changes specifically, but if we are adding something that requires clients to parse then this represents an API break.

genai-engine/src/config/database_config.py
```diff
@@ -17,8 +18,8 @@ class DatabaseConfig(BaseSettings):
     POSTGRES_PORT: int | str | None = None
     POSTGRES_USER: str | None = None
     POSTGRES_USE_SSL: bool = False
-    POSTGRES_CLIENT_CONNECTION_POOL_SIZE: int = 5
-    POSTGRES_CLIENT_CONNECTION_POOL_MAX_OVERFLOW: int = 15
+    POSTGRES_CLIENT_CONNECTION_POOL_SIZE: int = 10
+    POSTGRES_CLIENT_CONNECTION_POOL_MAX_OVERFLOW: int = 20
```
fryz, line 22: why the change here?

genai-engine/src/metrics_engine.py
```diff
@@ -0,0 +1,92 @@
...
+        self.scorer_client = scorer_client
+
+    def evaluate(
+        self,
+        request: MetricRequest,
+        metrics: List[Metric],
+    ) -> List[MetricResult]:
+        if not metrics:
+            return []
+        logger.info(f"Metric Request: {request.model_dump_json()}")
```
fryz, line 34: we shouldn't log this, as it container user data

genai-engine/src/metrics_engine.py
```diff
@@ -0,0 +1,92 @@
...
+        with TracedThreadPoolExecutor(tracer) as executor:
+            for metric in metrics:
+                future = executor.submit(self.run_metric, request, metric)
+                thread_futures.append((metric, future))
+        metric_results: list[MetricResult] = []
+        for metric, future in thread_futures:
+            exc = future.exception()
+            if exc is not None:
+                logger.error(
+                    "Metric evaluation failed. Metric: %s" % (metric.model_dump_json()),
```
fryz, line 55: please confirm that this doesn't have user data in it - we can't log things that contain user data

fryz (commented):
a few comments, mainly around making sure we're sanitizing/protecting log inputs/outputs and confirming that we are not breaking the API for the /v2 routes

otherwise, this PR is quite large and I scanned over the logic w/o following it in super detail. I have tested this out on my local machine and can confirm that I've got the tracing stuff working (not the metrics stuff), so think it's probably ok to move forward and fix any bugs that come out of it instead of refactoring this gigantic PR into composable parts.

genai-engine/src/config/database_config.py
videetparekh: This was an attempt to improve load testing results. I'll revert it.

genai-engine/src/routers/v2/task_management_routes.py
videetparekh: We have validations for it in the task_repository. See
https://github.com/arthur-ai/arthur-engine/blob/2f139d1daea69a5e19571e1e2e079cb1974753ae/genai-engine/src/repositories/tasks_repository.py#L253

genai-engine/src/schemas/response_schemas.py
videetparekh: Should it be optional or default to `[]`?

genai-engine/src/api_changelog.md
videetparekh: This adds metrics to the task response, similar to rules. I don't think this qualifies as a breaking change as the schema is only being expanded not reduced.

fryz (approved):

PR #161, opened 2025-05-19
UP-2514: Single script install for gen ai and ml engine

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-2514

I'm going to wait to make the FE updates until this gets merged. Let me know if this is how you were all thinking this script would be run to include the ml-engine as well!


alexshanabrook:
Sidenote, once this is ready to be merged, I'm going to need some help figuring out how to get this file from just being in this directory to being hosted somewhere like https://engine.arthur.ai, where the current genai-engine script is hosted, before adding it to the FE—let me know if those steps are documented anywhere!

deployment/docker-compose/arthur-engine/setup-and-start.sh
```diff
@@ -0,0 +1,147 @@
...
+echo "└───────────────────────────────────────────────────┘"
+
+check_docker_compose
+
+# parse command line arguments
+parse_script_vars "$@"
+
+# create necessary directories if not already present
+root_dir="$HOME/.arthur-engine-install"
+engine_subdir="$root_dir/arthur-engine"
```
mcgrawia, line 85: just thinking a bit into the future, I think it may make more sense to use a path like this?
```shell
~/.arthur-engine/local-stack/arthur-engine`  # for combined stack
~/.arthur-engine/local-stack/genai-engine`  # for gen ai only stack
```

deployment/docker-compose/arthur-engine/setup-and-start.sh
```diff
@@ -0,0 +1,147 @@
...
+env_file=".env"
+create_directory_if_not_present "$engine_subdir"
+
+if [[ -f "$engine_subdir/$env_file" ]]; then
+    echo "The $engine_subdir/$env_file file already exists."
+    echo "Please review the file and press any key to proceed to Docker Compose up..."
+    echo "Any flags that were set to update environment variables will be ignored."
+    read -n 1 -s
```
mcgrawia, line 93: I think this script should function like the AWS CLI in that if someone sets a command line argument like `--arthur-client-id` it will take precedence over a setting that may or may not already be in the `.env` file.

If the script is started without CLI args, it should default to the values already in the `.env` file. This will allow users to re-run the script later, without having to specify the credentials again, which is important because the UI will never be able to retrieve them after the initial installation.

deployment/docker-compose/arthur-engine/setup-and-start.sh
```diff
@@ -0,0 +1,147 @@
...
+        echo "Skipping OpenAI configuration..."
+    fi
+
+    echo "$all_env_vars" > "$engine_subdir/$env_file"
+    echo "Created combined .env file at $engine_subdir/$env_file"
+fi
+
+sleep 1
+cd "$engine_subdir"
+curl -s https://raw.githubusercontent.com/arthur-ai/arthur-engine/refs/heads/main/deployment/docker-compose/arthur-engine/docker-compose.yml | docker compose -f - up -d --pull always
```
mcgrawia, line 147: To ease debugging an troubleshooting locally, I think we should actually save the docker-compose file to the local folder. That way users can shut down/restart the stack easily without having to use a script. That said, I think the script should still download the latest compose file each time, but just persist it as well.

mcgrawia (commented):
@jdbarthur @Kacpersoftberry do either of you know how to update the script behind the engine.arthur.ai domain?

deployment/docker-compose/arthur-engine/setup-and-start.sh
alexshanabrook: Definitely—it does currently default to values in the `.env` file if there are no CLI args, it only requires them if there is no `.env` file in the first place. But I can update with the override—would you expect those CLI args to replace existing `.env` file variables or just to passed through to the docker command & leave the `.env` file untouched?

deployment/docker-compose/arthur-engine/setup-and-start.sh
alexshanabrook: Sure, makes sense! I mostly created those scripts because the local folder is hidden (`.arthur-engine-install`) so I wasn't sure that users will know where to find the docker-compose file to shut down/restart the stack

deployment/docker-compose/arthur-engine/setup-and-start.sh
mcgrawia: I think I would expect any CLI arg to overwrite the file. For example, if a customer wants to disable raw data retrieval, we should be able to give them a sample command like 

`./this-script --fetch-raw-data-enabled=false` 

which would update the `.env` file and restart the stack with the latest settings, but leave the credentials in the ENV file untouched. Does that make sense?

deployment/docker-compose/arthur-engine/setup-and-start.sh
mcgrawia: yeah they likely won't, but it will be helpful for troubleshooting with a customer if we can give them direct docker compose commands if we need to

deployment/docker-compose/arthur-engine/setup-and-start.sh
```diff
@@ -112,6 +147,8 @@ FETCH_RAW_DATA_ENABLED=$FETCH_RAW_DATA_ENABLED
 ########################################################
 "
 
+# prompt for OpenAI configuration if it's not already set
+if [[ -z "$GENAI_ENGINE_OPENAI_PROVIDER" ]]; then
```
alexshanabrook, line 151: Right now this will prompt for the OpenAI configuration any time it's not in the `.env` file yet. Previously it was only prompted for if the `.env` file didn't exist—let me know if that seems preferable & I can change it back, I haven't worked with shield enough to know what users would probably want

deployment/docker-compose/arthur-engine/setup-and-start.sh
```diff
@@ -0,0 +1,189 @@
...
+      # remove any quotes from the value
+      value=$(echo "$value" | tr -d '"'"'")
+      # export the variable to env
+      export "$key=$value"
+    done < "$env_file"
+  fi
+}
+
+missing_ml_engine_vars() {
+  local required_vars=("ARTHUR_API_HOST" "ARTHUR_CLIENT_ID" "ARTHUR_CLIENT_SECRET" "FETCH_RAW_DATA_ENABLED")
```
mcgrawia, line 90: one last comment, just tried this out locally and it works well but I think the ARTHUR_API_HOST and the FETCH_RAW_DATA_ENABLED should be optional args. If the host isn't set, it should just default to the one in the container (platform.arthur.ai) and fetch raw data should default to true

mcgrawia (approved):
lgtm!

PR #162, opened 2025-05-19
[no-ref] fix docker compose

mcgrawia:
1. remove container_name
2. update DB hostnames to match service name since container names are removed

Running the latest docker compose I get this error:
```
genai-engine-1  | sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) could not translate host name "arthur-engine-db" to address: Name or service not known
genai-engine-1  |
genai-engine-1  | (Background on this error at: https://sqlalche.me/e/20/e3q8)
```


jdbarthur (approved):
lgtm

PR #163, opened 2025-05-19
UP-2513: Install lsof in genai-engine image

alexshanabrook:
This is related to the PR here: https://github.com/arthur-ai/arthur-engine/pull/154/files

I'm separating the dockerfile changes from the docker-compose changes so that `lsof` will be in the `latest` image once this is merged to main so the changes to the docker-compose don't break anything

### DIFF
--- a/genai-engine/dockerfile
+++ b/genai-engine/dockerfile
@@ -23,6 +23,9 @@ WORKDIR /app
 ENV PYTHONPATH="$PYTHONPATH:/app/src"
 RUN poetry install --without dev,performance --no-ansi
 
+# install lsof for healthchecks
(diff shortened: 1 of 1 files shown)


mcgrawia (approved):
just confirming, does the `lsof` health check properly detect the service hasn't started until the models finish downloading? I just want to make sure FastAPI doesn't bind the port immediately but wait to serve traffic until the models download or something strange like that

alexshanabrook:
> just confirming, does the `lsof` health check properly detect the service hasn't started until the models finish downloading? I just want to make sure FastAPI doesn't bind the port immediately, causing `lsof` to pass, but then wait to serve traffic until the models download or something strange like that

Yep, I tested this & FastAPI does wait until the models download to bind to the port so this health check will fail until the models finish downloading

PR #164, opened 2025-05-19
UP-2521: Fix failing OTel tests

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-2521

Investigation results:
These breaking tests were introduced in the following PR: https://github.com/arthur-ai/arthur-engine/pull/140/files

The failure seems intentional, & the PR somehow got merged without removing the failing tests. The PR:
1. Removes `get_instance_id`—this is because it was only used by `send_telemetry_event` to set the user ID, but the PR added support for inferring the user's ID differently:
Screenshot 2025-05-19 at 3 25 04 PM
Screenshot 2025-05-19 at 3 25 17 PM

The tests for `get_instance_id` were not removed, so they've been failing.

2. Removes the `CONFIG_FILE_PATH` from the `TelemetryConfig` class. This is because it seems to have been only used to persist the `instance_id` removed in `get_instance_id`, so it's no longer needed. The tests for the file path were not removed, so they've been failing:
Screenshot 2025-05-19 at 3 26 25 PM

It seems to me these tests should have just been removed in the initial PR, so I'm solving the breaking tests by removing them now.


mcgrawia:
curious, if the tests were failing, why wasn't the MR merge blocked? @Kacpersoftberry do you know if we're blocking MRs on failing tests?

alexshanabrook:
> curious, if the tests were failing, why wasn't the MR merge blocked? @Kacpersoftberry do you know if we're blocking MRs on failing tests?

Yeah, it seems like an issue with the repo settings. I flagged it here: https://arthur-ai.slack.com/archives/C06FDBLFDEZ/p1747681748833779

mcgrawia (approved):
not an expert in this code but I think lgtm!

PR #187, opened 2025-06-02
UP-2490: Don't run pypi-publish on dev

alexshanabrook:
Did some research & it seems like pre-release versions are not ideally meant to be pushed to pypi. This means anyone trying to work on a version of arthur-common that hasn't been released yet will have to install it from the source code, but that seems reasonable.

Let me know if anyone feels strongly that we ignore the guidelines & push pre-release versions to pypi anyway!

Also @jdbarthur not sure how this affects our ability to validate the pipeline since it means it's not supposed to be run on a version increase to dev. let me know what you think!


jdbarthur:
Okay... @Kacperek0 for thoughts on how this might affect pipeline development

jdbarthur (approved):

PR #196, opened 2025-06-12
UP-2502: Migrate ML Engine code

Kacpersoftberry:


Kacpersoftberry:
> @Kacpersoftberry a couple of questions
> 
> 1. I notice the genAI client code is not present, what is the plan for that? Are users expected to generate it when running locally? If so, where are the instructions for how to do that?
> 2. It doesn't look like there is a readme, can we bring that over too?
> 3. Did any of the files need to be changed? or is this just a straight copy from the internal repo?
> 4. What's the plan for the Dockerfile? will that be added later?

1. Yes, I've expect that user will generate the genAI client by themself
2. I focused on moving the code, planed to add README.md later in next PR
3. Some of the files needed to be changed, because of the library/modules renaming. Some tests needed to be updated. 2 tests are marked as skipped for now, that we would need to fix up later
4. Adding Dockerfile and CI/CD would be a next step (different PR)

mcgrawia:
@Kacpersoftberry ok sounds good. For 2, 3, and 4, are those changes coming shortly? or will they be in separate tickets/sprints? Just want to make sure we don't merge this and move on without those

Kacpersoftberry:
> @Kacpersoftberry ok sounds good. For 2, 3, and 4, are those changes coming shortly? or will they be in separate tickets/sprints? Just want to make sure we don't merge this and move on without those

Once we merge this, next thing to do will be point 2,3 and 4. I hope we will done it in this/next week

mcgrawia (approved):

PR #197, opened 2025-06-12
Change hallucination max token limit to be set dynamically

notthattal:
### genai-engine/.env
- Bump up the GENAI_ENGINE_HALLUCINATION_CHECK_MAX_TOKEN_LIMIT env var to 7000 as the default

### llm_client.py
- Create get_gpt_model_token_limit() in the LLMExecutor class which retrieves the token limit (i.e. context window) for the gpt model being used

### rules_engine.py
- Create hallucination_token_limit variable for the RuleEngine class
- Create set_hallucination_token_limit() which either uses the new get_gpt_model_token_limit() function for the LLMExecutor or defaults to GENAI_ENGINE_HALLUCINATION_CHECK_MAX_TOKEN_LIMIT if it fails to find a gpt model context window

### utils/constants.py
- Created OPENAI_MODEL_CONTEXT_WINDOW_LENGTHS which is a dictionary relating models offered by OpenAI to their context window sizes
- Created AZURE_OPENAI_MODEL_CONTEXT_WINDOW_LENGTHS which is a dictionary relating models offered by Azure OpenAI to their context window sizes


genai-engine/src/scorer/llm_client.py
```diff
@@ -148,6 +148,19 @@ def get_gpt_model(
...
+            
+    def get_gpt_model_token_limit(self) -> int:
+        model = self.get_gpt_model()
+        
+        if model is None:
+            return -1
+        
+        if self.azure_openai_enabled and model.deployment_name in constants.AZURE_OPENAI_MODEL_CONTEXT_WINDOW_LENGTHS:
+            return constants.AZURE_OPENAI_MODEL_CONTEXT_WINDOW_LENGTHS[model.deployment_name]
+        elif self.openai_enable and model.model_name in constants.OPENAI_MODEL_CONTEXT_WINDOW_LENGTHS:
```
videetparekh, line 160: Can you rename this flag and it's calls to `openai_enabled` just to make sure we're being uniform?

videetparekh (approved):

PR #201, opened 2025-06-18
UP-2733: Add prompt version id dimension to default aggregations

alexshanabrook:


arthur-common/tests/unit/aggregation_functions/test_categorical_count.py
```diff
@@ -38,3 +38,41 @@ def test_string_categorical_count(
         "Highly Negative",
         "null",
     }
+
+
+def test_cv_categorical_count(
```
alexshanabrook, line 43: It may be overkill to include this for every aggregation, so for most of them I just made sure calling aggregate didn't result in errors because it's executing a different SQL in the case of the prompt_version_id columns for peace of mind.

mcgrawia:
@alexshanabrook @fryz didn't look super closely at this MR but assuming if it's passing the tests it should work fine. My only question/callout would be if it's worth hard-coding this one of case vs. trying to make this generic and re-usable later for segmentation, since it's essentially segmentation. 

I wonder if we could instead make this an input to the aggregation as a list type field, and allow the user to set it in the UI? That way for this demo, you could add `prompt_version_id` to the metric configs, and the metrics would handle it as a generic `segmentation_columns` list or something that we could build on later? 

If this is mainly just to get something quick and dirty working that's fine too, just a thought.

alexshanabrook:
> @alexshanabrook @fryz didn't look super closely at this MR but assuming if it's passing the tests it should work fine. My only question/callout would be if it's worth hard-coding this one case vs. trying to make this generic and re-usable later for segmentation, since it's essentially segmentation.
> 
> I wonder if we could instead support this use case by adding a new, optional `segment_by` field to the aggregation. It would be a list type field, and allow the user to set the columns to segment by? That way for this demo, you could add `prompt_version_id` to the metric's segmentation config, and the metrics would handle it as a generic `segmentation_columns` list or something that we could build on later?
> 
> If this is mainly just to get something quick and dirty working that's fine too, just a thought.

Yeah, completely agree that there are better solutions to this. I think the plan we made was just to get something working asap for the demo & later this use case will be actually supported by custom aggregations, since that work is planned to begin soon anyway—this way we don't have to coordinate FE & BE changes & Zach can start building the demo sooner

fryz (approved):

PR #209, opened 2025-06-23
UP-2738: Update workflows to allow publish arthur-common independently

Kacpersoftberry:


.github/workflows/arthur-common-release.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     # written from guide: https://johnfraney.ca/blog/how-to-publish-a-python-package-with-poetry-and-github-actions/
     name: Build release for PyPI
     if: |
-      contains(github.event.head_commit.message, 'Increment version to') &&
+      contains(github.event.head_commit.message, 'Increment arthur-common version to') &&
```
fryz, line 18: @Kacpersoftberry - it smells that we're using commit messages as the API for determining what files have changed 

Is there no way for us to determine and/or flag that the packages have changed that we can check based on the files?

PR #229, opened 2025-06-30
convert hallucination llm call to structured output

notthattal:
# Convert hallucination LLM Pipeline to use structured outputs

## Overview
The overarching goal of this PR is to convert the current hallucination llm pipeline from having to use 2 llm calls (1 for determining which claims are hallucinations and a second call to provide an explanation for these) to 1 llm call that can do both by utilizing GPT's structured outputs. 

*Note: Using structured outputs increases the token count by about 2.5x on average. This is due to how structured outputs are enforced by OpenAI* 

## hallucination_prompts.py
- Refactored the hallucination system prompts into their own file to de-clutter and better organize the main v2.py file
- Modified the hallucination prompts to have a proper sytem/ai/user structure. I found that the way we were setting up the prompts in v2.py incorrectly assigned what was meant to be the llm system prompt as a user message. So, remedied that. 
- Created a new system prompt/examples for the LLM call with structured outputs

## v2.py
- Create the pydantic classes to be used for structured outputs
- Create the new validate_claim_batch function which calls GPT with structured outputs only once
- Kept the old pipeline in-tact for users that still use GPT models which don't support structured outputs

## llm_client.py
- Create a function within the LLMExecutor to return a boolean for whether the GPT model being used supports structured outputs

## constants.py
- added two new sets which list which OpenAI and Azure OpenAI gpt models support structured outputs. Currently, neither platforms offer an automated way of determining which models support structured outputs.


genai-engine/src/utils/constants.py
```diff
@@ -287,4 +287,63 @@
     'gpt-35-turbo-16k': 16385,
 }
 
+##################################################################
+
+##################################################################
+# MODELS THAT SUPPORT STRUCTURED OUTPUTS
```
fryz, line 293: is there not an API we can hit that allows us to determine this?

fryz:
Just read through your description 

To confirm, instead of a 2-stage process (eval for hallucination -> explanation), we're now batching that into a single call? 
If so, that might explain the 2x increase in token consumption? I believe that for non-hallucinations, we're not sending those for explanations. 

Also - given the change, have you run benchmarks against this? How does it perform?

notthattal:
> To confirm, instead of a 2-stage process (eval for hallucination -> explanation), we're now batching that into a single call? If so, that might explain the 2x increase in token consumption? I believe that for non-hallucinations, we're not sending those for explanations.

Yeah on the notes when we went over implementing structured outputs my understanding was the goal of this was to implement structured outputs and reduce the calls to the LLM from 2 calls to 1. I can change it back to 2 if we'd prefer.

The token increase may have to do with a reduction in call partially. But, I tested a simple system prompt and user query example in a separate workspace and found even with just that that using structured outputs used ~5x more tokens on that small example.

notthattal:
> 
> Also - given the change, have you run benchmarks against this? How does it perform?

I ran approximately 30 test-cases against the old and new method to see if there was a difference in the model determining hallucination/explanation and found no differences (except obv slight differences in explanation due to the stochastic nature of responses by LLMs). I didn't benchmark it in databricks I can check to see if we have datasets to use a precedent in their to get metrics for the LLM call? I know we test the claims classifier there but haven't checked if we benchmark the LLM calls there

genai-engine/src/scorer/checks/hallucination/hallucination_prompts.py
```diff
@@ -0,0 +1,373 @@
...
+            "text_list_str": "- BERT came out in 2019\n- BERT is an NLP model\n- LLMs are also NLP models\n- ChatGPT is a great new NLP model",
+            "output": """[
+                LLMClaimResult(claim_text="BERT came out in 2019", is_hallucination=False, explanation="The context references BERT as the most advanced model as of 2019."),
+                LLMClaimResult(claim_text="BERT is an NLP model", is_hallucination=False, explanation="The context directly states this."),
+                LLMClaimResult(claim_text="LLMs are also NLP models", is_hallucination=True, explanation="LLMs are not mentioned in the context."),
+                LLMClaimResult(claim_text="ChatGPT is a great new NLP model", is_hallucination=True, explanation="ChatGPT is not referenced in the context.")
+            ]""",
+        },
+        {
+            "context": "This software library is only usable in the context of NLP, and the most advanced NLP model to date so far (2019) is BERT from Google.",
```
videetparekh, line 98: Both examples here are not claims. Will they not be automatically filtered out by the claims classifier?

genai-engine/src/scorer/checks/hallucination/hallucination_prompts.py
```diff
@@ -0,0 +1,373 @@
...
+            "output": """[
+                LLMClaimResult(
+                    claim_text="BERT is a solid model.",
+                    is_hallucination=False,
+                    explanation="The claim is supported because it references BERT, the advanced NLP model mentioned in the context."
+                )
+            ]"""
+        },
+        {
+            "context": "This software library is only usable in the context of NLP, and the most advanced NLP model to date so far (2019) is BERT from Google.",
```
videetparekh, line 167: Not a claim.

genai-engine/src/scorer/checks/hallucination/hallucination_prompts.py
```diff
@@ -0,0 +1,373 @@
...
+            "num_texts": "3",
+            "text_list_str": "- the actor is terrible\n- the actor didnt deserve an award\n- the actor is terrible because they have released no movies",
+            "output": """[
+                LLMClaimResult(claim_text="the actor is terrible", is_hallucination=False, explanation="The context states the performer is terrible."),
+                LLMClaimResult(claim_text="the actor didnt deserve an award", is_hallucination=False, explanation="The word 'inexplicably' implies the award was undeserved."),
+                LLMClaimResult(claim_text="the actor is terrible because they have released no movies", is_hallucination=True, explanation="The context says they released a movie, contradicting the claim.")
+            ]""",
+        },
+        {
+            "context": "Furthermore see our blog post https www arthur ai blog automating data drift thresholding in systems for an overview of data on how Arthur automates the choice of thresholding for drift metrics",
```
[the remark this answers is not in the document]
notthattal: fixed [here](https://github.com/arthur-ai/arthur-engine/pull/229/commits/4580f1cfdc2eecdfe8f3d7126caf58b986f94d7b)

genai-engine/src/scorer/checks/hallucination/hallucination_prompts.py
```diff
@@ -0,0 +1,373 @@
...
+    end_instruction = "=Now the real thing=\nMake sure that you include only {num_texts} labels or else you will fail.\nContext: {context}\n{num_texts} Texts\n{text_list_str}\n{num_texts} Labels: "
+
+    return get_flagging_prompt(flagging_examples, flagging_instruction, end_instruction)
+
+def get_flagged_claim_explanation_prompt():
+    explanation_examples = [
+        {
+            "context": "There are two dogs eating my shoes and it hurts",
+            "flagged_claim": "it does not hurt",
+            "output": "the claim is unspported because the context mentions that it hurts, but the claim mentions it does not, which is a contradiction",
```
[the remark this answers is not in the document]
notthattal: fixed [here](https://github.com/arthur-ai/arthur-engine/pull/229/commits/4580f1cfdc2eecdfe8f3d7126caf58b986f94d7b)

genai-engine/src/scorer/checks/hallucination/v2_prompts.py
```diff
@@ -0,0 +1,170 @@
...
+            "num_texts": "3",
+            "text_list_str": "- the actor is terrible\n- the actor didnt deserve an award\n- the actor is terrible because they have released no movies",
+            "output": """[
+                LLMClaimResult(claim_text="the actor is terrible", is_hallucination=False, explanation="The context states the performer is terrible."),
+                LLMClaimResult(claim_text="the actor didnt deserve an award", is_hallucination=False, explanation="The word 'inexplicably' implies the award was undeserved."),
+                LLMClaimResult(claim_text="the actor is terrible because they have released no movies", is_hallucination=True, explanation="The context says they released a movie, contradicting the claim.")
+            ]""",
+        },
+        {
+            "context": "What are common data drift metrics?",
```
videetparekh, line 43: The context is a question with no information about KL Divergence and P/R/F1, and so the model should detect the fact that the response is provided without any sort of context and flag them as hallucinations.

genai-engine/src/scorer/checks/hallucination/v2_prompts.py
```diff
@@ -0,0 +1,170 @@
...
+            "text_list_str": "- BERT came out in 2019\n- BERT is an NLP model\n- LLMs are also NLP models\n- ChatGPT is a great new NLP model",
+            "output": """[
+                LLMClaimResult(claim_text="BERT came out in 2019", is_hallucination=False, explanation="The context references BERT as the most advanced model as of 2019."),
+                LLMClaimResult(claim_text="BERT is an NLP model", is_hallucination=False, explanation="The context directly states this."),
+                LLMClaimResult(claim_text="LLMs are also NLP models", is_hallucination=True, explanation="LLMs are not mentioned in the context."),
+                LLMClaimResult(claim_text="ChatGPT is a great new NLP model", is_hallucination=True, explanation="ChatGPT is not referenced in the context.")
+            ]""",
+        },
+        {
+            "context": "There are two dogs eating my shoes and it hurts",
```
videetparekh, line 64: Lines 64-108 seem like duplicates

genai-engine/src/scorer/checks/hallucination/v2_prompts.py
notthattal: Fixed [here](https://github.com/arthur-ai/arthur-engine/pull/229/commits/b0bfb908a001b260af13650957e0e833a30e1933)

videetparekh (approved):

PR #261, opened 2025-07-21
Add is_agentic flag to tasks

videetparekh:
Adds a new flag to tasks: `is_agentic: bool`. This allows us to know at execution time whether the task is agentic or not, and allows us to iterate over our data models in the future without blowing up the existing workflow.


notthattal (approved):
Looks good to me

PR #299, opened 2025-08-01
Update dependencies in poetry.lock and pyproject.toml; add cx-oracle,…

iabouhashish:
… psycopg, pymysql, pyodbc, and sqlalchemy


iabouhashish:
@alexshanabrook Made some changes, the script can be used for local installations (hence why it supports 2 distributions) The docker now just runs the code in it.

Waiting on some understanding for the integration testing and then will be good to go :)

ml-engine/Dockerfile
```diff
@@ -3,6 +3,35 @@ FROM python:3.13-slim AS install
 # install poetry
 RUN pip install poetry==2.1.3
 
+# install system dependencies for database drivers
+RUN apt-get update && apt-get install -y \
```
alexshanabrook, line 7: Could we just call the `install_db_dependencies.sh` script from here? It looks like they do the same thing, so we only have to maintain it in one place

alexshanabrook:
otherwise I think we can merge, assuming this builds—I wanted to be able to confirm the integration tests are passing against it, but I think we can merge & fix the integration tests later if not. Right now the integration tests already test the functionality in this MR, so they'll fail until we merge anyway—so there's not much point in holding off.

alexshanabrook:
^ the above force push was just me rebasing–I'm going to try to run the integration tests against this to see if we can merge while Ibrah's out

PR #301, opened 2025-08-01
port toxicity changes from shield to engine

notthattal:
[Jira Ticket](https://arthurai.atlassian.net/browse/UP-2888)

# Porting toxicity changes from shield to engine
- Added the new profanity model to the engine
- Added text chunking iterator and sliding window iterator
- Updated Env vars to include new batching/chunk size env vars and updated the toxicity token limit
- Added expanded profanity blacklist
- Altered profanity check to use stricter Regex and fallback to using the new profanity model
- Updated template files to default to new toxicity token limit
- Downloaded profanity model locally and uploaded it to repo
- Updated the loading mechanism for the toxicity model/tokenizer to user AutoModel/Tokenizer from hf so we can use the fast tokenizer to retrieve offset mappings for chunking
- Only run toxicity and harmful requests if we pass profanity since if any profanity is detected we return "Toxicity Detected" with a score of 0.999. This improves latency for runs flagged as profanity
- Updated token limits in the CFT stack templates

# TODO:
Move the profanity model to the container image instead


genai-engine/src/scorer/checks/toxicity/toxicity.py
```diff
@@ -15,21 +20,45 @@
...
+    or 32,
+)
+
+TOXICITY_MODEL_BATCH_SIZE = int(
+    get_env_var(constants.GENAI_ENGINE_TOXICITY_MODEL_BATCH_SIZE_ENV_VAR, True) or 64,
+)
+
+
+# Global singletons
+_profanity_classifier = None
```
videetparekh, line 56: Shouldn't the toxicity classifier also be a global singleton?

notthattal: ah, you're right

notthattal: yes, will do

videetparekh (approved):

PR #309, opened 2025-08-05
port shield prompt injection changes to engine

notthattal:
# Porting Recent Prompt Injection Changes in Shield to Engine
- converted the prompt injection classifier to use the new SlidingWindowChunkIterator to split the text up into chunks of max length = 512, and a stride of 256. Now, the prompt injection classifier is no longer limited to 2000 characters and will not cut out the middle section of texts. Instead the entire text is now evaluated for prompt injection
- Removed the warning that the middle of text will be cut out if a text is longer than 512 tokens long
- Moved the creation of the prompt injection classifier to utils/model_load.py to be in parity with other metrics


genai-engine/src/rules_engine.py
```diff
@@ -283,17 +282,6 @@ def run_prompt_injection_rule(
         )
 
         rule_score = self.scorer.score(score_request)
-        total_tokens = self.token_counter.count(request.prompt)
```
videetparekh, line 286: Same as above, I'm seeing this log in the shield repos as well

notthattal: will move to shield

PR #322, opened 2025-08-07
Fix UP-2918: 400 err on NewMetricRequest

notthattal:
# Fix UP-2918: 400 err on NewMetricRequest
- Fixes the bug where a 400 error would be returned if relevance_threshold was explicitly set to None. We were previously checking if "relevance_threshold" was in the dictionary not if relevance_threshold was actually set to a value. Now we check both.
- Similarly when checking if neither relevance_threshold nor use_llm_judge were set, we now check also if the values weren't set not just if they were set or not.


alexshanabrook:
otherwise lgtm!

genai-engine/src/schemas/request_schemas.py
```diff
@@ -504,10 +504,9 @@ def set_config_type(cls, values):
                     config_values["use_llm_judge"] = False
 
                 # If neither is set, default to use_llm_judge=True
-                if (
-                    "relevance_threshold" not in config_values
-                    and "use_llm_judge" not in config_values
-                ):
+                if config_values.get(
+                    "relevance_threshold",
+                ) is None and not config_values.get("use_llm_judge", False):
```
videetparekh, line 509:
```
if isinstance(config_values, dict):
    relevance_threshold = config_values.get("relevance_threshold")
    use_llm_judge = config_values.get("use_llm_judge")

    # 1. Conflict check: both set
    if relevance_threshold is not None and use_llm_judge:
        raise HTTPException(
            status_code=400,
            detail="relevance_threshold and use_llm_judge=true are mutually exclusive. Set use_llm_judge=false when using relevance_threshold.",
        )

    # 2. If relevance_threshold is set, enforce use_llm_judge = False
    if relevance_threshold is not None and use_llm_judge is None:
        config_values["use_llm_judge"] = False

    # 3. If neither is set, default to use_llm_judge = True
    if relevance_threshold is None and use_llm_judge is None:
        config_values["use_llm_judge"] = True   
```
We can simplify the checks

alexshanabrook (approved):

PR #354, opened 2025-08-19
Fix port conflicts between Arthur services

videetparekh:
## Problem:
Multiple Arthur services were using conflicting ports, preventing simultaneous local development across arthur-auth, arthur-scope, and genai-engine.

## Solution:
Database: 5432 → 5435 (avoid arthur-auth conflict)
GenAI Server: 8000 → 3030 (avoid arthur-scope conflict)
Adminer: 8080 → 8081 (avoid arthur-auth conflict)

### DIFF
--- a/ml-engine/Dockerfile
+++ b/ml-engine/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13-slim AS install
+FROM python:3.13-slim-bookworm AS install
 
 # install poetry and wget
 RUN pip install poetry==2.1.3 && \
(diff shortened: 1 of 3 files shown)


ntatsumi (dismissed):

genai-engine/src/server.py
```diff
@@ -408,7 +408,7 @@ async def redirect_to_docs():
 def start():
     send_telemetry_event(TelemetryEventTypes.SERVER_START_INITIATED)
     app = get_app()
-    uvicorn.run(app, host="0.0.0.0", port=8000)
```
fryz, line 411: this will require a change to the helm + cfts 

Can we leave this port the same and update port mappings in our docker compose?

videetparekh: In local dev mode, we presently run the genai-engine (non-docker) on 8000 which conflicts with the app-plane

videetparekh: @fryz Spoke to Nori about this. At a CFT/Helm level, the server is run via gunicorn (already set to port 3030 in `gunicorn.conf.py`) via `docker-entrypoint.sh`. This change should have no effect on helm+cfts.

fryz (approved):

PR #355, opened 2025-08-19
[NO-REF]: Pin python images for ML engine to debian 12 for compatibil…

alexshanabrook:
Previously, with the new release of `python:3.13-slim`, we were seeing the following errors when running the container:

```2025-08-18 18:50:02 /usr/local/bin/python3.13: /lib/aarch64-linux-gnu/libm.so.6: version `GLIBC_2.38' not found (required by /usr/local/bin/../lib/libpython3.13.so.1.0)```

This is an incompatibility issue; the new release of python 3.13-slim had an incompatible version of GLIBC with the distroless image we're using in the build step.

The distroless image we're using uses debian-12 as its base distribution: https://github.com/GoogleContainerTools/distroless?tab=readme-ov-file#base-operating-system

This pins both images to use debian-12 for compatibility.

Passing scope integration tests against this build:
Screenshot 2025-08-19 at 9 40 28 AM

### DIFF
--- a/ml-engine/Dockerfile
+++ b/ml-engine/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.13-slim AS install
+FROM python:3.13-bookworm AS install
 
 # install poetry and wget
 RUN pip install poetry==2.1.3 && \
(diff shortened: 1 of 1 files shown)


alexshanabrook:
included arthur-common bump from this PR going into arthur-scope soon: https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1290

jdbarthur (approved):

PR #358, opened 2025-08-19
Im/fix pagination in get spans

mcgrawia:
Fixing two bugs the query spans API:
1. previously, the API did a full table scan to fetch all trace IDs for a task
2. previously, the API paginated the trace IDs based on the random order returned from the DB

Now, the API only retrieves the most recent N trace IDs and paginates them using the same filters and sorts as the traces.

### DIFF
--- a/.github/workflows/arthur-engine-workflow.yml
+++ b/.github/workflows/arthur-engine-workflow.yml
@@ -255,13 +255,11 @@ jobs:
         run: |
           cd ml-engine
           docker build -t arthurplatform/ml-engine:${{ env.VERSION }} .
-          echo "Pushing paused temporarily..."
-          exit 0
           docker push arthurplatform/ml-engine:${{ env.VERSION }}
-          if [ "${{ github.ref }}" = "refs/heads/main" ]; then
-            docker tag arthurplatform/ml-engine:${{ env.VERSION }} arthurplatform/ml-engine:latest
-            docker push arthurplatform/ml-engine:latest
-          fi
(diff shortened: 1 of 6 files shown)


genai-engine/src/repositories/span_repository.py
```diff
@@ -192,35 +198,85 @@ def query_spans_with_metrics_as_traces(
...
+        if end_time:
+            conditions.append(DatabaseSpan.created_at <= end_time)
 
-        return paginated_trace_ids if paginated_trace_ids else None
+        # Use a subquery to get the earliest span time for each trace
+        # This ensures we order traces by their start time (earliest span)
+        earliest_span_subquery = (
+            select(
+                DatabaseSpan.trace_id,
+                func.min(DatabaseSpan.created_at).label("earliest_time")
```
videetparekh, line 263: This should probably be `DatabaseSpan.start_time`

mcgrawia: I debated that as well, but saw the spans query already uses `created_at` for filtering on the final list: https://github.com/arthur-ai/arthur-engine/blob/cd0382fc0b5094d481082577675f69b70abc1797/genai-engine/src/repositories/span_repository.py#L559-L563

videetparekh: I would consider that a bug as well. The created at time is when the spans are received by GenE, not when they're created, so that might cause weirdness where you Traces get interchanged because one trace was delayed when sending, even though it was first.

mcgrawia: updated to use `start_time` but let me know

PR #363, opened 2025-08-20
UP-2945 Api Consolidation

martin-arthur-ai:


ml-engine/src/ml_engine/tools/api_client_type_converters.py
```diff
@@ -128,6 +135,15 @@ def new_rule_request_api_to_shield_client(
             config=Config(cls.rule_config_api_to_shield_client(api.config)),
         )
 
+    @classmethod
```
alexshanabrook, line 138: curious, why did these need to be added?

ml-engine/src/ml_engine/job_executors/task_management_job_executors.py
```diff
@@ -154,7 +156,7 @@ def upload_final_task_state(self, model_id: str, task: TaskResponse) -> None:
         self.tasks_client.put_task_state_cache(
             model_id=model_id,
             put_task_state_cache_request=PutTaskStateCacheRequest(
-                task=ScopeClientTypeConverter.task_response_api_to_scope_client(task),
+                task=task,
```
alexshanabrook, line 159: I'm still curious how this is working. It seems to me like even though arthur-common is shared, since the OpenAPI generation happens outside of arthur-common, we probably still have 2 different types for each shared object between arthur-client and the genai-client. I wonder if this is the cause of the failing model task management integratino test? Or maybe I'm missing something about how the change in arthur-common adds support for this

ml-engine/scripts/openapi_client_utils.sh
```diff
@@ -50,7 +50,19 @@ if [[ "$purpose" == "generate" ]]; then
     openapi-generator-cli generate -i ../../genai-engine/staging.openapi.json --skip-validate-spec -g "python" -o ../src/genai_client --package-name genai_client -p packageVersion=$version
 fi
 
+if [[ "$purpose" == "generate-common" ]]; then
```
alexshanabrook, line 53: do you mind documenting this usage in the README? it seems like you're adding new functionality to this script?

genai-engine/src/api_changelog.md
```diff
@@ -2,6 +2,12 @@ The intention of this changelog is to document API changes as they happen to eff
 
 ---
 
+# 09/03/2025
+- **CHANGE** for **URL**: /v1/span/{span_id}/metrics  added the optional property 'span_name' to the response with the '200' status
```
alexshanabrook, line 6: do these changes have to happen in this PR? Should we have Videet review them? I went through all of the code and totally missed where these are added so I might not be the right person

In general, it might have been easier to keep this as a separate MR—it makes it easier to review & iterate on

genai-engine/staging.openapi.json
```diff
@@ -1359,6 +1379,16 @@
                                     }
                                 },
                                 "example5": {
+                                    "summary": "Hallucination Rule V1 Example (Deprecated)",
```
alexshanabrook, line 1382: how did this get added back in? It seems like it might have not existed as an example before since it's been deprecated?

genai-engine/src/api_changelog.md
martin-arthur-ai: My pipeline fails without running the changelog and getting these additions. They're from some other branch, but they're coming up as required in mine. I would love to not have to do this.

ml-engine/src/ml_engine/job_executors/task_management_job_executors.py
martin-arthur-ai: genai engine imports taskresponse from athur-common, and it gets added to the generated spec. No conversion is necessary anymore.

genai-engine/src/api_changelog.md
videetparekh: This happened because the changes from the DB optimizations branch were copied over here, but the changelog hadn't been updated yet. The changelog is eternally finnicky, but I don't think this is a big deal.

genai-engine/staging.openapi.json
martin-arthur-ai: Figured this out. I used the NewMetricRequest version that was up-to-date code wise, but it had the outdated examples. This will have to be changed in arthur-common

genai-engine/changelog.py
```diff
@@ -83,7 +83,7 @@ def generate_new_openapi(path: str) -> None:
 
 
 def main():
-    path_directory = os.path.dirname(os.path.abspath(__name__))
+    path_directory = os.path.dirname(os.path.abspath(__file__))
```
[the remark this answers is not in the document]
martin-arthur-ai: I changed it back because it stopped working for me again.

ml-engine/src/ml_engine/job_executors/task_management_job_executors.py
alexshanabrook: Right, but this is my understanding (please let me know if it's not yours):

1. We still have the arthur-common models being embedded in control plane routes: https://gitlab.com/ArthurAI/arthur-scope/-/blob/main/scope/app_plane/app/api_models/tasks.py?ref_type=heads#L31 As a result, when we generate the arthur-client from the control plane, we will generate bindings for the arthur-common models
2. We also have the arthur-common models embedded in GenAI engine routes. As a result, when we generate the genai-client from the GenAI engine, we will generate different bindings for the arthur-common models.
3. In the ML engine, it first receives these task requests from the control plane in the form of task creation job requests. These will have the type of the arthur-client generated bindings.
4. The ML engine then makes requests to the GenAI engine using these arthur-client generated bindings. However, this is what I expect to be incompatible—the ML engine still needs to parse these from the arthur-client generated models to the genai-client generated models. They'll have all of the same fields, but since they're coming from different Python packages, the interpreter & code will think we're passing an incorrect type.

Is that different from your understanding? I think if the model_task_management test is passing with this configuration we can decide that (4) is untrue & the code can handle the implicit conversion since all of the expected fields are present, but it seems like from your screenshot that test is not passing so this may be a place to look for the bug

genai-engine/src/api_changelog.md
alexshanabrook: @videetparekh  how did that get merged without the pipeline blocking it but it's blocking now? I want to make sure if there's an issue with the pipeline here we're fixing it—but will defer to you whether these changes are fine to merge with the notes in this PR

genai-engine/staging.openapi.json
alexshanabrook: cool—if it's not important to fix immediately can we ticket it?

alexshanabrook:
> Able to run integration tests. Most passing, with the failures unrelated to imports.
> 
> Screenshot 2025-09-04 at 17 39 00

@videetparekh can you confirm if these failures are expected from your perspective?

ml-engine/pyproject.toml
```diff
@@ -32,7 +32,7 @@ types-requests = "2.32.4.20250809"
 psutil = "7.0.0"
 pyarrow = "21.0.0"
 arthur-client = "1.4.1319"
-arthur-common = "^2.1.60"
+arthur-common = "^2.1.61"
```
alexshanabrook, line 35: @martin-arthur-ai mind removing the `^` here too please? we'll see the same issue as in the GenAI engine since we're not semantically versioning arthur-common

alexshanabrook:
all GenAI integration tests are passing apart from one, which is failing on main—so I don't think it needs to be resolved in this PR:
Screenshot 2025-09-09 at 5 07 59 PM

genai-engine/staging.openapi.json
alexshanabrook: Ticketed here: https://arthurai.atlassian.net/browse/UP-3053 
@videetparekh as you review, please confirm that we can just merge this & clean it up later since it just changes the examples & they're noted to be deprecated anyway

genai-engine/src/scorer/metrics/tool_selection/tool_selection.py
```diff
@@ -4,14 +4,14 @@
 from langchain_core.output_parsers.json import JsonOutputParser
 from pydantic import BaseModel, Field
 
-from schemas.common_schemas import LLMTokenConsumption
-from schemas.enums import MetricType, ToolClassEnum
+from arthur_common.models.common_schemas import LLMTokenConsumption
+from arthur_common.models.enums import MetricType
+from schemas.enums import ToolClassEnum
```
videetparekh, line 9: Any reason this was left in genai-engine? This could be used in the ML Engine for filtering.

genai-engine/staging.openapi.json
videetparekh: Yeah, we can remove this downstream. This has little impact.

genai-engine/src/services/metrics_integration_service.py
```diff
@@ -9,8 +9,9 @@
 from dependencies import get_metrics_engine
 from repositories.metrics_repository import MetricRepository
 from repositories.tasks_metrics_repository import TasksMetricsRepository
-from schemas.internal_schemas import MetricResult, Span
-from schemas.metric_schemas import MetricRequest
+from schemas.internal_schemas import Span
+from arthur_common.models.metric_schemas import MetricRequest
+from schemas.internal_schemas import MetricResult
```
videetparekh, line 14: Why is this broken into two imports from the same file?

genai-engine/tests/unit/scorer/test_tool_selection.py
```diff
@@ -3,10 +3,11 @@
 
 import pytest
 
-from schemas.common_schemas import LLMTokenConsumption
-from schemas.enums import MetricType, ToolClassEnum
+from arthur_common.models.common_schemas import LLMTokenConsumption
+from arthur_common.models.enums import MetricType
 from schemas.internal_schemas import MetricResult
-from schemas.metric_schemas import MetricRequest
+from schemas.enums import ToolClassEnum
```
videetparekh, line 9: Same as above

genai-engine/src/schemas/metric_schemas.py
```diff
@@ -1,39 +1,14 @@
...
-
-class QueryRelevanceMetric(RelevanceMetric):
-    """Inherits from RelevanceMetric. This class is left empty so that the openapi response schema remains the same as before, but we have a single source of truth for the relevance metric details."""
-
-
-class ResponseRelevanceMetric(RelevanceMetric):
-    """Inherits from RelevanceMetric. This class is left empty so that the openapi response schema remains the same as before, but we have a single source of truth for the relevance metric details."""
-
-
 class ToolSelectionCorrectnessMetric(BaseModel):
```
videetparekh, line 12: Any reason to leave this here? It's comparable to the Relevance Metrics, so if they are in commons, this should be as well.

genai-engine/tests/unit/scorer/test_tool_selection.py
alexshanabrook: going to resolve as it's covered in ticket: https://arthurai.atlassian.net/browse/UP-3057

genai-engine/src/schemas/metric_schemas.py
alexshanabrook: also covered in this fast follow ticket: https://arthurai.atlassian.net/browse/UP-3057

videetparekh (approved):

PR #369, opened 2025-08-21
Update port for inter-container comms

videetparekh:


deployment/docker-compose/arthur-engine/docker-compose.yml
videetparekh, line 1: See that DB now runs on 5435, but since genai-engine is running on the same network, it communicates via `db:5432`

jdbarthur (approved):

PR #400, opened 2025-08-27
Bake the latest AWS RDS global cert bundle

ntatsumi:
AWS users are one of the primary Arthur Engine users. Some of them are in air-gapped environments where they do not have easy access to download the Postgres SSL certs. Instead of having no certs file as a start, this change provides AWS RDS global cert bundle as the default when you don't specify the `POSTGRES_SSL_CERT_DOWNLOAD_URL` envar.

### DIFF
--- a/genai-engine/dockerfile
+++ b/genai-engine/dockerfile
@@ -1,9 +1,8 @@
 ################################################################################
 #                                                                              #
-#             Build GenAI Engine on slim-bookworm images for cpu and gpu             #
+#       Build GenAI Engine on slim-bookworm images for cpu and gpu             #
 #                                                                              #
 ################################################################################
-
 
 # TORCH_DEVICE must be either "cpu" or "gpu"
 ARG TORCH_DEVICE=cpu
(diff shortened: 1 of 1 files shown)


genai-engine/dockerfile
```diff
@@ -52,18 +50,20 @@ COPY .env /app/
 ARG ENABLE_TELEMETRY=false
 RUN echo "TELEMETRY_ENABLED=${ENABLE_TELEMETRY}" >> /app/.env
 
+# Download AWS RDS global certificate bundle
+RUN apt-get update && apt-get install -y curl && \
+    curl -o /app/postgres-cert.pem https://truststore.pki.rds.amazonaws.com/global/global-bundle.pem
```
jdbarthur, line 55: This looks good. Do you want to add cleanup lines to the command? I think it's like:

&& \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

ntatsumi:
@jdbarthur The .pem file gets copied into the final layer with this command:
https://github.com/arthur-ai/arthur-engine/blob/b228865915584fabd271957d5ab76b0aa33fc296/genai-engine/dockerfile#L85

PR #427, opened 2025-09-05
UP-3006: ML Engine Dev Doc Enhancements

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-3006


ml-engine/scripts/install_db_dependencies.sh
```diff
@@ -0,0 +1,65 @@
+#!/bin/bash
+
+# Database Dependencies Installation Script for ML Engine (macOS Only)
+# This script installs system dependencies required for:
+# - psycopg (PostgreSQL)
+# - pymysql (MySQL)
+# - pyodbc (ODBC)
+# - MSSQL connections
+
+set -e  # Exit on any error
```
alexshanabrook, line 10: @iabouhashish I moved this script into the `scripts` directory but also removed the information it had about Oracle because I don't think it will be read & moved that documentation to the README. Would appreciate if you take a look at the changes & make sure they work for you too

alexshanabrook: Didn't realize that changing the directory & the script at the same time would make the diff hard to read, sorry, but mostly I just removed everything oracle-specific from this

PR #440, opened 2025-09-08
chore(deps): update arthur-common to v2.1.61 and add openinference-se…

iabouhashish:
chore(deps): update arthur-common to v2.1.61 and add openinference-semantic-conventions v0.1.21

- Updated arthur-common dependency version in pyproject.toml and poetry.lock.
- Added new dependency openinference-semantic-conventions v0.1.21 in poetry.lock.
- Refactored imports in multiple files to use new request and response schemas.


alexshanabrook:
> or actually, what I'm curious about here @alexshanabrook @jdbarthur is why the arthur-common dep merged into the arthur-engine w/o passing some sort of linting/tests that would have caught the dependencies were bad

Can you explain what you mean by why the arthur-common dep merged?

This work was in progress to be handled by Martin after he released breaking changes to arthur-common: https://github.com/arthur-ai/arthur-engine/pull/363 as part of the API consolidation effort. As far as I can tell, arthur-common wasn't auto-merged & ibrah raised this PR himself.

https://github.com/arthur-ai/arthur-engine/pull/363/files#r2325117479 : I already have this comment in that PR to change how we pin the version of arthur-common. Since we're not semantically versioning the releases, we need to specify specific versions, not any version above some tested version. Not sure if that helps with what you're referring to?

PR #458, opened 2025-09-12
add timeout for custom aggregation calculations

notthattal:
[Jira Ticket](https://arthurai.atlassian.net/browse/UP-2841)

- Add a timeout for the calculations of both default and custom aggregations
- Create two new environment variables 'ML_ENGINE_AGGREGATION_TIMEOUT' and 'ML_ENGINE_DEFAULT_THREAD_POOL_MAX_WORKERS' to specify how long the timeout should be and how many workers should be used (the latter of which has a default value set in code which is only overriden when the env var is > 0)
- Created a new unit test to verify the timeout error is raised and the message matches what is expected


ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
```diff
@@ -280,12 +291,28 @@ def _calculate_metrics(
...
+                        init_args,
+                        aggregate_args,
+                    )
+                    metrics_to_add = future.result(
+                        timeout=ML_ENGINE_AGGREGATION_TIMEOUT,
+                    )
+
                 self._add_dataset_dimensions_to_metrics(metrics_to_add, aggregate_args)
                 metrics.extend(metrics_to_add)
+            except TimeoutError as exc:
```
alexshanabrook, line 309: Is there a reason we can't handle this by the `except Exception` below? It includes the `exc_info` of the original exception so it should still have the specifics that this was a `TimeoutError`. The idea with this try-except is that if one aggregation for a model fails, we still want to calculate all of the other aggregations before raising the error at the end (in a function that calls this one). That way if one aggregation is misconfigured, the other ones still produce metrics

ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
```diff
@@ -45,6 +49,13 @@
 from tools.connector_constructor import ConnectorConstructor
 from tools.validators import validate_schedule
 
+ML_ENGINE_DEFAULT_THREAD_POOL_MAX_WORKERS = (
```
alexshanabrook, line 52: Curious, why was it important to add this support? It seems like since we're only calling one function instead of multiple in the `submit` function, we'll only ever have 1 used anyway?

ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
notthattal: done [here](https://github.com/arthur-ai/arthur-engine/pull/458/commits/e47bd6aa275d75be55e310ccf5a4b9d38342d275)

notthattal:
> Member

Created partner MR [here](https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1338)

alexshanabrook:
lgtm!

PR #468, opened 2025-09-17
add agg id and version as metric dimensions

notthattal:
## Description

* Added aggregation id to all metric dimensions
* Adds aggregation version if it exists to metric dimensions (will only affect custom aggregations)
* Added a unit test to test the function \_add_dimensions_to_metrics correctly returns the right dimensions (i.e.

  dataset_name, dataset_id, aggregation_id and aggregation_version only if it exists in the AggregationSpec).

## Type of change

- [ ] Bug fix (fixes an issue)
- [x] New feature (adds functionality)
- [x] Add tests (tests functionality)

## Jira Ticket

> https://arthurai.atlassian.net/browse/UP-2843

### DIFF
--- a/ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
+++ b/ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
@@ -296,6 +296,9 @@ class MetricsCalculationExecutor:
                     )
 
                 self._add_dataset_dimensions_to_metrics(metrics_to_add, aggregate_args)
+                self._add_dimensions_to_metrics(
+                    metrics_to_add, aggregate_args, agg_spec
+                )
                 metrics.extend(metrics_to_add)
             except Exception as exc:
                 # continue with future metrics calculations in case they're successful and log an error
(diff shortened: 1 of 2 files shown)


ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
```diff
@@ -296,6 +296,9 @@ def _calculate_metrics(
                     )
 
                 self._add_dataset_dimensions_to_metrics(metrics_to_add, aggregate_args)
```
alexshanabrook, line 298: It seems like we should delete this line? Looks like you handled the dataset dimensions in the `_add_dimensions_to_metrics` call

ml-engine/src/ml_engine/job_executors/metrics_calculation_executor.py
```diff
@@ -334,12 +338,20 @@ def _add_dataset_dimensions_to_metrics(
             # no dataset arg found, metric isn't tied to a dataset
```
alexshanabrook, line 338: while this doesn't happen today, I think even if no dataset arg is found we would still want to include the aggregation ID and version in the dimensions. Right now this function will leave it out since this check happens first

alexshanabrook (commented):
otherwise lgtm!

PR #473, opened 2025-09-19
[NO-REF]: remove entrypoint override from docker-compose

alexshanabrook:
Slack thread: https://arthur-ai.slack.com/archives/C06F737P85R/p1758292285940169


alexshanabrook:
I want to just remove the entrypoints from here—there's no reason we would have to override them when they're internally defined & it just means we duplicate logic in multiple places

PR #509, opened 2025-09-26
Engine UI Foundation

mcgrawia:


genai-engine/ui/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "dev": "vite --mode dev",
+    "build": "tsc -b && vite build",
+    "lint": "eslint .",
+    "preview": "vite preview",
+    "generate-api": "npx swagger-typescript-api@13.0.16 -p ../staging.openapi.json -o src/lib/api-client --name api-client.ts --axios --clean-output --extract-request-params --extract-request-body --extract-response-body --extract-response-error --extract-enums --union-enums --add-readonly --sort-routes --sort-types",
+    "generate-api:clean": "rm -rf src/lib/api-client && npm run generate-api"
+  },
+  "dependencies": {
+    "@arizeai/openinference-semantic-conventions": "^2.1.1",
+    "@next/font": "^14.2.15",
```
xV3rt, line 16: We probably don't want next/font dependency in a React Vite app

genai-engine/ui/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "react": "^19.1.1",
+    "react-dom": "^19.1.1",
+    "react-router-dom": "^7.9.2",
+    "swagger-typescript-api": "^13.2.13",
+    "tailwindcss": "^4.1.13",
+    "weaviate-client": "^3.8.1"
+  },
+  "devDependencies": {
+    "@eslint/js": "^9.36.0",
+    "@tailwindcss/vite": "^4.1.13",
```
xV3rt, line 30: If I'm not mistaken, that dependency should make @tailwindcss/postcss AND autoprefixer unnecessary (or certainly both of them shouldn't be needed)

Either way, we discussed Tailwind CSS with Martin and we might want to leave it in the repo but we also want to use MUI components for the ease of future migration to the old repository.

genai-engine/ui/package.json
```diff
@@ -0,0 +1,44 @@
...
+    "lint": "eslint .",
+    "preview": "vite preview",
+    "generate-api": "npx swagger-typescript-api@13.0.16 -p ../staging.openapi.json -o src/lib/api-client --name api-client.ts --axios --clean-output --extract-request-params --extract-request-body --extract-response-body --extract-response-error --extract-enums --union-enums --add-readonly --sort-routes --sort-types",
+    "generate-api:clean": "rm -rf src/lib/api-client && npm run generate-api"
+  },
+  "dependencies": {
+    "@arizeai/openinference-semantic-conventions": "^2.1.1",
+    "@next/font": "^14.2.15",
+    "@tailwindcss/postcss": "^4.1.13",
+    "@types/react-router-dom": "^5.3.3",
```
xV3rt, line 18: react-router-dom seems to be written in typescript and doesn't require a separate an additional type dependency anymore in v7

xV3rt (commented):
I think it's a good baseline. There is only 1 major change that needs to happen (not necessarily before the merge) which is utilising MUI components for the ease of future merge into the other repository.

This doesn't necessarily mean tailwind has to go, we could use both - something to be discussed. Removing tailwind and focusing on MUI might overall be better in terms of consistency. Tailwind does probably have an advantage of LLM's working better with it though, not sure how relevant that is

PR #525, opened 2025-10-01
add support for nested segmentation columns

notthattal:
## Description
- Replaces column_scalar_dtype_from_dataset_schema to use a column id rather than a column name in order to get nested columns
- Creates a column name to ID map 
- Creates a new test case to verify nested columns pass validation in process_agg_args and aggregate() still works appropriately

## Jira Ticket
https://arthurai.atlassian.net/browse/UP-3075

## Partner MRs
1. https://github.com/arthur-ai/arthur-common/pull/52
2. https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1385


ml-engine/src/ml_engine/metric_calculators/metric_calculator.py
```diff
@@ -164,6 +165,7 @@ def _validate_segmentation_args(
                     )
                 for column_name in arg_val:
                     self._validate_segmentation_single_column(
+                        col_names_to_id.get(column_name, ""),
```
alexshanabrook, line 168: Why would a column not have an ID? I'm curious hat happens if we pass `col_id` as an empty string

ml-engine/src/ml_engine/metric_calculators/metric_calculator.py
```diff
@@ -243,28 +246,33 @@ def _all_dataset_columns(self, datasets: List[Dataset]) -> Dict[str, str]:
...
+                    nested_fields = self._field_names_from_object_type_column(
+                        column.definition.actual_instance,
+                        escape_identifier(column.source_name),
                     )
+                    all_dataset_columns.update(nested_fields)
+                    for field_id, field_name in nested_fields.items():
+                        col_names_to_id[field_name] = field_id
 
-        return all_dataset_columns
+        return all_dataset_columns, col_names_to_id
```
alexshanabrook, line 275: Can we explain what we're returning in the docstring? I'm worried all these maps are going to get confusing—at some point this will probably all have to be refactored to make more sense, but in the meantime we should make sure we're writing as clear comments as we can. My experience of working in this section of the codebase is that it's pretty confusing

ml-engine/src/ml_engine/metric_calculators/metric_calculator.py
notthattal: The only way a column wouldn't have an ID I think would be if we passed it in incorrectly or it didn't actually exist. Having that failsafe just ensures that column_scalar_dtype_from_dataset_schema() returns None and that this ValueError gets raised:

```python
raise ValueError(
    "Could not fetch scalar column data type for evaluation of segmentation column "
    "requirements. Either the column does not exist or it is an object or list type "
    "or a nested column, which are not supported for segmentation columns.",
)
```

alexshanabrook: Okay great thanks for explaining!

Do you mind updating that `ValueError` to remove the comment saying we don't support nested columns? Seems like that's changed in this PR

notthattal: Updated [here](https://github.com/arthur-ai/arthur-engine/pull/525/commits/e8642135dd217a9886f0a76567e64c7b33bd55ef)

alexshanabrook (approved):
approving although may want to bump the arthur-common version in this PR before merging

notthattal:
> approving although may want to bump the arthur-common version in this PR before merging

Bumped the arthur-common version in the ml engine

alexshanabrook (approved):

alexshanabrook (approved):

PR #543, opened 2025-10-06
UP-3099: Add secret storage table

Kacpersoftberry:
1. Add secret storage table


genai-engine/src/db_models/secret_storage_models.py
```diff
@@ -0,0 +1,22 @@
+from datetime import datetime
+
+from sqlalchemy import TIMESTAMP, ForeignKey, String
+from sqlalchemy.orm import Mapped, mapped_column
+
+from db_models.base import Base
+from db_models.custom_types import EncryptedJSON
+
+
+class DatabaseSecretStorage(Base):
```
mcgrawia, line 10: I don't think I see this table in your migration above?

genai-engine/src/db_models/custom_types.py
```diff
@@ -29,3 +35,36 @@ def process_result_value(self, value, engine):
...
         else:
             return []
+
+
+class EncryptedJSON(types.TypeDecorator):
+    impl = types.Text
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        passphrase = os.getenv("SECRET_PASSPHRASE")
```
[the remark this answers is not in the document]
mcgrawia: also, can we make the name of this env var a bit more specific? Thinking like `ARTHUR_ENGINE_SECRET_STORE_KEY` or something that conveys its use.

genai-engine/src/db_models/custom_types.py
```diff
@@ -29,3 +35,36 @@ def process_result_value(self, value, engine):
...
+        if not passphrase:
+            raise ValueError("SECRET_PASSPHRASE environment variable not set")
+        salt = b"some_static_or_dynamic_salt"  # Should be store/define securely
+        kdf = PBKDF2HMAC(
+            algorithm=hashes.SHA256(),
+            length=32,
+            salt=salt,
+            iterations=100000,
+            backend=default_backend(),
+        )
```
mcgrawia, line 55: is there a reason we need to do the password hashing? if we're already expecting someone to set the secret externally, it seems reasonable they can just set a base64 encoded random string?

genai-engine/src/db_models/custom_types.py
```diff
@@ -29,3 +35,36 @@ def process_result_value(self, value, engine):
             return roles
         else:
             return []
+
+
+class EncryptedJSON(types.TypeDecorator):
```
mcgrawia, line 40: Interesting! So if I understand this right, this custom type allows devs to pass in a `dict` value and it will be auto-encrypted/decrypted on save/load from the DB?

genai-engine/alembic/versions/2025_10_06_1525-d5305fe6b6e4_add_secret_storage.py
```diff
@@ -0,0 +1,202 @@
...
+        nullable=False,
+        existing_server_default=sa.text("CURRENT_TIMESTAMP"),
+    )
+    op.drop_index(
+        op.f("idx_traces_covering"),
+        table_name="trace_metadata",
+        postgresql_include=["trace_id", "end_time", "span_count"],
+    )
+    op.drop_index(op.f("idx_traces_task_start"), table_name="trace_metadata")
+    op.drop_index(op.f("idx_traces_task_time_range"), table_name="trace_metadata")
```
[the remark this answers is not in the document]
Kacpersoftberry: We were using alembic to autogenerate scripts. I see we are missing part on how to autogenerate. I will add those in this PR.

genai-engine/src/db_models/custom_types.py
Kacpersoftberry: We can pass the string itself, it's still encoded inside the Fernet init method.

genai-engine/README.md
```diff
@@ -112,13 +116,13 @@ A Postgres database is required to run the GenAI Engine. The easiest way to get
...
+#### Set up variables for Alembic
 
-The Alembic database migration tool needs to be run the first time and every time a new database schema change is added.
-Maks sure the Poetry install is complete and you have a running Postgres instance first.
-
-`cd` to `/genai-engine` and run the commands below:
+Make sure the Poetry install is complete and you have a running Postgres instance first. After that setup the variables
+(example contains default valuse)
 
+ExampleL
```
mcgrawia, line 125: Example?

genai-engine/README.md
```diff
@@ -127,10 +131,28 @@ export POSTGRES_PORT=5435
...
+If you made some changes to DB models you should create migration script. You could use alembic to generate such script. If you create
+a new file that contains DB changes import this file to [DB Modelse init file](src/db_models/__init__.py).
+
+After that run following command:
+```bash
+poetry run alembic revision --autogenerate "<commit message>"
+```
+
+**Keep the message short, avoid special characters.**
+
```
mcgrawia, line 155: thank you for adding this, FYI @videetparekh

genai-engine/src/db_models/custom_types.py
mcgrawia: Ok, let's do that then. I'd rather not have hard coded salts and manual hashing in the code if we can avoid it

genai-engine/src/db_models/custom_types.py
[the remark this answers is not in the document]
mcgrawia: @Kacpersoftberry thanks for updating all the deployments to use this, but I think the fact there were so many places we needed to update is a big flag this is a breaking change. We have customers who have installed the engine not using these exact deployment scripts either, so requiring this ENV var to be set going forward is going to be a major upgrade headache.. Additionally, it's going to be rather insecure for anyone who uses the default install without reading the documentation to change this default setting, or else they will use the same key as every other customer's env. Plus, that default secret is exposed publicly in our open source code..

As a result of all these signals, I think we need to implement the default mechanism I described above. I think it could be pretty simply to add, just a single new `keys` database table, and changing the Fernet encryption key to use a `Multifernet` where the first key comes from the `ARTHUR_ENGINE_SECRET_STORE_KEY` environment variable, if set, and the second key comes from the database one, which should be generated on boot if one does not already exist?

Open to alternatives too, but simplifying the install and upgrade process is crucial to get users on this new product, so I don't think we can add new, required configuration. It needs to be secure by default without manual actions from devs similar to starting keycloak.

Thoughts?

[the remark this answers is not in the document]
mcgrawia: looks like langfuse expects users to set an encryption key in docker compose: https://github.com/langfuse/langfuse/blob/main/docker-compose.yml#L25

I don't see an equivalent on Phoenix, so not sure if it just expects all provider API secrets to be set via ENV? https://arize.com/docs/phoenix/self-hosting/deployment-options/docker#docker

Either way, I think for now if Nori is agreed adding a new required env var isn't a big issue for deployments, we can go with the current approach. I think the big risk is that all customers will share the same default key, just set in the deployment scripts instead of the DB, so the majority of users who just spin up the platform without reading docs will all use the same default secret, meaning they can all decrypt each other's secrets if they are leaked.

genai-engine/alembic/versions/2025_10_07_1542-ad13a540c9e3_add_secret_storage.py
```diff
@@ -0,0 +1,218 @@
+"""Add secret storage
+
+Revision ID: ad13a540c9e3
+Revises: 4617f961d9f0
```
mcgrawia, line 4: also just calling out so we don't forget, this will need to be rebased as there have been other MRs merged with newer migrations

genai-engine/alembic/versions/2025_10_07_1542-ad13a540c9e3_add_secret_storage.py
```diff
@@ -0,0 +1,218 @@
...
+    )
+    op.drop_index(
+        op.f("idx_metric_results_span_id_metric_type"),
+        table_name="metric_results",
+    )
+    op.alter_column(
+        "spans",
+        "raw_data",
+        existing_type=postgresql.JSONB(astext_type=sa.Text()),
+        type_=postgresql.JSON(astext_type=sa.Text()),
```
videetparekh, line 62: This column should be retained as JSONB (binary formatted) for better querying and faster access. Instead of this change, maybe we change the definition in the ORM Model to JSONB.

genai-engine/alembic/versions/2025_10_07_1542-ad13a540c9e3_add_secret_storage.py
```diff
@@ -0,0 +1,218 @@
...
+        existing_type=postgresql.TIMESTAMP(),
+        nullable=False,
+        existing_server_default=sa.text("CURRENT_TIMESTAMP"),
+    )
+    op.drop_index(
+        op.f("idx_spans_llm_task_time"),
+        table_name="spans",
+        postgresql_where="((span_kind)::text = 'LLM'::text)",
+    )
+    op.drop_index(
```
videetparekh, line 84: Similar to above

mcgrawia (commented):
also, please update the start script for local devs running outside docker with the new env var: https://github.com/arthur-ai/arthur-engine/blob/dev/genai-engine/scripts/start_local_server.sh

deployment/docker-compose/arthur-engine/docker-compose.yml
```diff
@@ -32,10 +32,11 @@ services:
       # The address you want to access the engine from (e.g. hostname, DNS)
       - GENAI_ENGINE_INGRESS_URI=http://localhost:3030
       - GENAI_ENGINE_ADMIN_KEY=changeme_genai_engine_admin_key
+      - GENAI_ENGINE_SECRET_STORE_KEY=changeme_secret_store_key
```
mcgrawia: @Kacpersoftberry looks like from the docker compose variable precedence, this is going to override the generated value in the `env_file` from the script, so every docker compose deployment is going to use the secret `changem_secret_store_key` https://docs.docker.com/compose/how-tos/environment-variables/envvars-precedence/

genai-engine/src/db_models/secret_storage_models.py
```diff
@@ -0,0 +1,23 @@
...
+from db_models.custom_types import EncryptedJSON
+from schemas.enums import SecretType
+
+
+class DatabaseSecretStorage(Base):
+    __tablename__ = "secret_storage"
+    id: Mapped[str] = mapped_column(String, primary_key=True)
+    name: Mapped[str] = mapped_column(String)
+    value: Mapped[dict] = mapped_column(EncryptedJSON)
+    owner_id: Mapped[str] = mapped_column(
```
mcgrawia: @Kacpersoftberry the way this foreign key is set up today, if the corresponding api key is deleted, this link will cause the delete to fail with a foreign key constraint error. Going to remove this for now since it's not needed

genai-engine/src/db_models/secret_storage_models.py
```diff
@@ -0,0 +1,18 @@
...
+from db_models.base import Base
+from db_models.custom_types import EncryptedJSON
+from schemas.enums import SecretType
+
+
+class DatabaseSecretStorage(Base):
+    __tablename__ = "secret_storage"
+    id: Mapped[str] = mapped_column(String, primary_key=True)
+    name: Mapped[str] = mapped_column(String)
+    value: Mapped[dict] = mapped_column(EncryptedJSON)
```
mcgrawia: although sleek, I don't think this approach is going to work well when we need to implement secret rotation. From this chat thread, it sounds like sqlalchemy uses the unencrypted value to determine if it should update the column in the DB. When we do a secret rotation, the unencrypted value will remain the same, but the encrypted value will change, so we need to handle the encryption outside of sqlalchemy so it can detect re-encryption with a new key: https://chatgpt.com/share/68f2ae8a-d9f0-8005-ac6b-21b53a02f322. Going to work on refactoring this into a repo

PR #546, opened 2025-10-06
Fix UI pipeline during arthur-engine release

alexshanabrook:
See failing pipeline in this PR: https://github.com/arthur-ai/arthur-engine/pull/545

This is happening because if the lockfile hasn't been generated, the UI fails in the build stage.


mcgrawia (approved):
@xV3rt I think this is correct but let me know if not and we can make a follow up MR

PR #598, opened 2025-10-15
(feat) Added Model name to validate_prompt and validate_response.

iabouhashish:
Add Model Name Support to Inference Validation

## Summary
This PR adds support for tracking and validating model names in the GenAI Engine's inference validation system. Users can now specify which model was used for generating prompts and responses, enabling better tracking and analysis of model-specific performance.

### Core Features
- **Model Name Field**: Added `model_name` field to inference models and validation requests
- **Database Schema**: New Alembic migration adds `model_name` column to inference tables
- **API Support**: Updated validation endpoints to accept and return model names
- **Backward Compatibility**: Model name is optional, existing functionality remains unchanged

### Database Changes
- Added `model_name` column to `DatabaseInference` model
- Created Alembic migration: `2025_10_14_0905-1e4ff2a22cbb_adding_model_name.py`
- Updated repository methods to handle model name parameter

### API Updates
- **Prompt Validation**: `POST /api/v2/validate_prompt` now accepts `model_name` parameter
- **Response Validation**: `POST /api/v2/validate_response` now accepts `model_name` parameter
- **ValidationResult**: Response schema includes `model_name` field

### Testing
- **Enhanced Test Coverage**: Added comprehensive test suite for model name functionality
- **Integration Tests**: 5 new test functions with `_with_model_name` suffix
- **Test Client Updates**: Updated `GenaiEngineTestClientBase` to support model name parameter
- **Validation Assertions**: Tests verify model name is properly stored and returned

## 📊 Impact
- **Files Changed**: 9 files
- **Lines Added**: 392 lines
- **Lines Removed**: 12 lines
- **Test Coverage**: 100% coverage for new model name functionality

### DIFF
--- a/genai-engine/alembic/versions/2025_10_14_0905-1e4ff2a22cbb_adding_model_name.py
+++ b/genai-engine/alembic/versions/2025_10_14_0905-1e4ff2a22cbb_adding_model_name.py
@@ -1,7 +1,7 @@
 """Adding model_name
 
 Revision ID: 1e4ff2a22cbb
-Revises: ecc45f321532
+Revises: 5b5d630418ba
 Create Date: 2025-10-14 09:05:42.220289
 
 """
@@ -12,7 +12,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "1e4ff2a22cbb"
-down_revision = "ecc45f321532"
+down_revision = "5b5d630418ba"
 branch_labels = None
 depends_on = None
 
--- a/genai-engine/alembic/versions/2025_10_22_1646-5b5d630418ba_add_versioning_to_agentic_prompts.py
+++ b/genai-engine/alembic/versions/2025_10_22_1646-5b5d630418ba_add_versioning_to_agentic_prompts.py
@@ -1,7 +1,7 @@
 """add versioning to agentic prompts
 
 Revision ID: 5b5d630418ba
-Revises: 16d81eaac5c2
+Revises: ecc45f321532
 Create Date: 2025-10-22 16:46:40.462930
 
 """
@@ -12,7 +12,7 @@ from alembic import op
 
 # revision identifiers, used by Alembic.
 revision = "5b5d630418ba"
-down_revision = "16d81eaac5c2"
+down_revision = "ecc45f321532"
 branch_labels = None
 depends_on = None
 


genai-engine/src/db_models/inference_models.py
```diff
@@ -29,6 +29,7 @@ class DatabaseInference(Base):
         lazy="joined",
     )
     task: Mapped["DatabaseTask"] = relationship(lazy="joined")
+    model_name: Mapped[str] = mapped_column(String, nullable=True)
```
alexshanabrook: Curious, why did you did you decide to put this here instead of in `DatabaseInferenceResponse` or `DatabaseInferencePrompt`? I don't know enough about the GenAI resources to be sure which would be expected, so I'm not saying it needs to be moved, just wanted a little more context

genai-engine/src/validation/prompt.py
```diff
@@ -27,10 +28,12 @@ def validate_prompt(
...
         conversation_id=body.conversation_id,
         user_id=body.user_id,
+        model_name=body.model_name,
     )
 
     return ValidationResult(
         inference_id=inference_prompt.inference_id,
         rule_results=inference_prompt._to_response_model().prompt_rule_results,
         user_id=body.user_id,
+        model_name=body.model_name,
```
alexshanabrook: also curious how this is relevant to a `ValidationResult`? It seems like it's more being handled as new metadata for a `PromptValidationRequest`, in which case it's not really a result of the validation so much as a config on the initial inference?

genai-engine/tests/integration/test_rule_execution.py
```diff
@@ -313,3 +314,322 @@ def test_rule_result_skipped(client: GenaiEngineTestClientBase):
...
+
+
+@pytest.mark.parametrize(
+    ("prompt_str", "expected_rule_results"),
+    [
+        ["Hello World", [RuleResultEnum.PASS]],
+        [" ".join(["Hello World" for _ in range(5000)]), [RuleResultEnum.SKIPPED]],
+    ],
+)
+@pytest.mark.integration_tests
```
alexshanabrook: could you help me understand how this case & all of the ones below it are edge cases for the handling of `model_name`? To me it seems like they're testing non-model name code paths (as in, cases that have nothing to do with whether or not `model_name` is being set correctly & will have the same behavior every time). If that's the case, could we remove them? I want to be careful of adding unnecessary tests that we'll need to maintain moving forward if they're not testing specific cases we're interested in

iabouhashish: It's just to make sure that adding model_name hasn't resulted in any changes to the tests above, which don't contain model_name. I can remove most of them if we think this is overkill.

genai-engine/src/schemas/internal_schemas.py
```diff
@@ -1166,6 +1170,7 @@ class Inference(BaseModel):
     inference_response: Optional[InferenceResponse] = None
     inference_feedback: List[InferenceFeedback]
     user_id: Optional[str] = None
+    model_name: Optional[str] = None
```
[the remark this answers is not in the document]
iabouhashish: Same as above. I don't foresee model_name being used any differently currently than user_id, so following the same pattern.

iabouhashish:
> > @videetparekh @alexshanabrook I have now removed it from validate_prompt, only added to validate_response.
> 
> Sorry, how did you do that? Don't you have to remove it from arthur-common first here? https://github.com/arthur-ai/arthur-common/blob/bf1950ab2e847fbff5525fc69cacd0725442cc15/src/arthur_common/models/request_schemas.py#L334

Here: https://github.com/arthur-ai/arthur-common/pull/56

PR #599, opened 2025-10-15
add streaming functionality to agentic prompts

notthattal:
## Description
- Add the ability to stream responses from litellm for both unsaved and saved agentic prompts
- Adds support in the permission_checker decorator to work with async functions
- Adds a unit test to test streaming for the unsaved and saved prompt routes

## Jira Ticket
https://arthurai.atlassian.net/browse/UP-3181

### DIFF
--- a/genai-engine/tests/unit/repositories/test_agentic_prompts_repository.py
+++ b/genai-engine/tests/unit/repositories/test_agentic_prompts_repository.py
@@ -14,11 +14,14 @@ from schemas.agentic_prompt_schemas import (
     AgenticPromptMessage,
     AgenticPrompts,
     CompletionRequest,
+    LLMResponseFormat,
+    LLMResponseSchema,
     PromptCompletionRequest,
     ToolChoice,
     ToolChoiceFunction,
     VariableTemplateValue,
 )
+from schemas.common_schemas import JsonSchema
 from schemas.response_schemas import AgenticPromptRunResponse
 
 
@@ -78,6 +81,23 @@ def sample_db_prompt(sample_prompt_data):
 def expected_db_prompt_messages(sample_db_prompt):
     """Create expected DatabaseAgenticPrompt messages"""
     return to_agentic_prompt_messages(sample_db_prompt.messages)
+
+
+def mock_completion(*args, **kwargs):
+    response_format = kwargs.get("response_format")
+    json_schema = response_format.get("json_schema")
+    schema = json_schema.get("schema")
+
+    if json_schema.get("strict") and "additionalProperties" not in schema:
+        raise ValueError(
+            "Invalid schema for response_format 'joke_struct': "
+            "In context=(), 'additionalProperties' is required to be supplied and to be false.",
+        )
+
+    mock_response = MagicMock()
+    mock_response.choices = [MagicMock()]
+    mock_response.choices[0].message = {"content": "ok", "tool_calls": None}
+    return mock_response
 
 
 @pytest.mark.unit_tests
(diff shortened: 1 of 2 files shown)


genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -55,6 +56,33 @@ def get_validated_agentic_task(
     return task
 
 
+async def execute_prompt_completion(
+    agentic_prompt_service: AgenticPromptRepository,
+    prompt: AgenticPrompt,
+    run_config: AgenticPromptRunConfig,
+):
```
mcgrawia: this should have a return type

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -129,20 +157,36 @@ def get_all_agentic_prompt_versions(
...
+        unsaved_prompt: AgenticPromptUnsavedRunConfig
+        current_user: User
+
+    Returns:
+        AgenticPromptRunResponse or StreamingResponse
+    """
     try:
         agentic_prompt_service = AgenticPromptRepository(None)
-        return agentic_prompt_service.run_unsaved_prompt(run_config)
+        prompt, run_config = unsaved_prompt._to_prompt_and_config()
```
mcgrawia: generally private methods should not be called outside the class. If this method is meant to be used externally, it shouldn't start with `_`

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -129,20 +157,36 @@ def get_all_agentic_prompt_versions(
 
 @agentic_prompt_routes.post(
     "/completions",
-    summary="Run an agentic prompt",
-    description="Run an agentic prompt",
+    summary="Run/Stream an unsaved agentic prompt",
+    description="Runs or streams an unsaved agentic prompt",
     response_model=AgenticPromptRunResponse,
```
mcgrawia: how does openAPI reflect that this endpoint can return both a static and streaming response? I'm wondering how our generated client is going to handle this

genai-engine/src/schemas/agentic_prompt_schemas.py
```diff
@@ -398,6 +405,37 @@ def run_chat_completion(
...
+        )
+
+        cost = completion_cost(complete_response)
+        msg = complete_response.choices[0].message
+
+        yield AgenticPromptRunResponse(
+            content=msg.get("content"),
+            tool_calls=msg.get("tool_calls"),
+            cost=f"{cost:.6f}",
+        )
```
mcgrawia: what exactly is the wire format of this stream? shouldn't we just use what's returned from liteLLM?

notthattal: sorry what do you mean by wire format? We do return what liteLLM responds but then after it's done streaming we just reconstruct the full message so we can report the cost of the call so we don't need to make a separate call to get the cost with the same data afterwards.

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -20,20 +19,20 @@ def __init__(self, db_session: Session):
...
-        task_id: str,
-        prompt_name: str,
+        prompt: AgenticPrompt,
         run_config: AgenticPromptRunConfig = AgenticPromptRunConfig(),
-    ) -> AgenticPromptRunResponse:
-        prompt = self.get_prompt(task_id, prompt_name)
-        return prompt.run_chat_completion(run_config)
+    ):
+        async for chunk in prompt.stream_chat_completion(run_config):
+            yield chunk
```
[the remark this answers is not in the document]
notthattal: FastAPI won't actually stream the response if we return just the generator directly

genai-engine/src/schemas/agentic_prompt_schemas.py
```diff
@@ -398,6 +405,37 @@ def run_chat_completion(
...
+        response = await acompletion(model=model, **completion_params)
+
+        collected_chunks = []
+        async for chunk in response:
+            collected_chunks.append(chunk)
+
+            if chunk.choices and chunk.choices[0].delta:
+                delta = chunk.choices[0].delta.get("content", "")
+                if delta:
+                    yield delta
```
mcgrawia: why are we stripping out the content from the chunks? it looks like there's a bunch of other info in the chunks that will be helpful for the UI to know: https://docs.litellm.ai/docs/#streaming-response-format-openai-format

for example, if it makes a tool call?

genai-engine/src/repositories/agentic_prompts_repository.py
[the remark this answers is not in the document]
mcgrawia: I guess is this function even really needed? why doesn't `generate_prompt_stream` just call `prompt.stream_chat_completion(run_config)` directly?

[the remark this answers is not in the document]
notthattal: testing it out it seems when returning the generator directly fastapi tries to serialize it as a normal response and starts throwing this err:

`TypeError: 'async_generator' object is not iterable`

genai-engine/src/schemas/agentic_prompt_schemas.py
notthattal: I can have it stream the whole thing

genai-engine/src/schemas/agentic_prompt_schemas.py
notthattal: this is an example of what each chunk looks like when returned from the acompletion call to litellm:

```json
{
  "id": "chatcmpl-CR5tesCJcyRUIjyELmt5epy5cA61U",
  "created": 1760573559,
  "model": "gpt-4o",
  "object": "chat.completion.chunk",
  "system_fingerprint": "fp_eb3c3cb84d",
  "choices": [
    {
      "finish_reason": None,
      "index": 0,
      "delta": {
        "provider_specific_fields": None,
        "refusal": None,
        "content": " capital",
        "role": None,
        "function_call": None,
        "tool_calls": None,
        "audio": None
      },
      "logprobs": None
    }
  ],
  "provider_specific_fields": None,
  "citations": None,
  "service_tier": "default",
  "obfuscation": "0yca5w9R"
}
```

genai-engine/src/repositories/agentic_prompts_repository.py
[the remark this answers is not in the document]
notthattal: I just modified the function that is called to return the generator to test it. Also to answer above that function was just made to delegate whether to call the streaming function or the non-streaming version run_chat_completion

genai-engine/src/routers/v1/agentic_prompt_routes.py
notthattal: Added in [this commit](https://github.com/arthur-ai/arthur-engine/pull/599/commits/3f29e748c4369a8046e879c73a712ede3d76d7eb)

mcgrawia (approved):

PR #601, opened 2025-10-15
Add search dataset versions endpoint

alexshanabrook:


genai-engine/src/schemas/internal_schemas.py
```diff
@@ -1885,9 +1887,27 @@ def _from_database_model(
...
+
+    def to_response_model(self) -> DatasetVersionMetadataResponse:
+        return DatasetVersionMetadataResponse(
+            version_number=self.version_number,
+            created_at=_serialize_datetime(self.created_at),
+        )
+
+
+class DatasetVersion(DatasetVersionMetadata):
     dataset_id: uuid.UUID
```
mcgrawia: curious, should `dataset_id` be in the version metadata?

mcgrawia (approved):
lgtm otherwise!

genai-engine/src/schemas/internal_schemas.py
alexshanabrook: okay makes sense, I can add it then!

mcgrawia (approved):

PR #608, opened 2025-10-16
Fix Pagination Arthur-Common Release

alexshanabrook:


genai-engine/src/schemas/common_schemas.py
```diff
@@ -3,22 +3,12 @@
...
 
 from arthur_common.models.enums import (
-    PaginationSortMethod,
     UserPermissionAction,
     UserPermissionResource,
 )
 from pydantic import BaseModel, Field
 
 
-class PaginationParameters(BaseModel):
```
alexshanabrook: this should have been removed during the API consolidation—I made sure all imports already reference the object in `arthur-common`

PR #614, opened 2025-10-17
UP-3202: add column names to dataset versions

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-3202


genai-engine/alembic/versions/2025_10_17_1349-7c330aebf8e7_add_dataset_columns.py
```diff
@@ -0,0 +1,40 @@
...
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "dataset_versions",
+        sa.Column("column_names", sa.JSON(), nullable=True),
+    )
+    op.execute(
+        """
+        UPDATE dataset_versions
+        SET column_names = '[]'
```
alexshanabrook: since the previous changes weren't deployed anywhere or used by anyone, I figured it would be fine to just set this to an empty list—let me know if you disagree & I can add better support

genai-engine/src/server.py
```diff
@@ -299,7 +299,7 @@ def get_base_app(
         kwargs["lifespan"] = lifespan
     app = FastAPI(
         title="Arthur GenAI Engine",
-        version=version,
+        version="2.1.94",
```
mcgrawia: what is this change for?

PR #624, opened 2025-10-20
UP-3146 prompt versioning v1

martin-arthur-ai:
Key changes:

- Backend returns timestamp to the frontend
- Optimize front end to prevent duplicate API queries
- Updated the Tools functions to match the backend structure

### DIFF
--- a/genai-engine/src/schemas/agentic_prompt_schemas.py
+++ b/genai-engine/src/schemas/agentic_prompt_schemas.py
@@ -206,7 +206,7 @@ class ToolChoice(BaseModel):
 class AgenticPromptBaseConfig(BaseModel):
     created_at: Optional[datetime] = Field(
         default=None,
-        description="Timestamp when the prompt was created. Used for versioning.",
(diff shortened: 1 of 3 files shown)


genai-engine/src/schemas/agentic_prompt_schemas.py
```diff
@@ -289,6 +289,10 @@ class Config:
 
 class AgenticPrompt(AgenticPromptBaseConfig):
     name: str = Field(description="Name of the agentic prompt")
+    created_at: Optional[datetime] = Field(
+        default=None,
+        description="Timestamp when the prompt was created. Used for versioning.",
+    )
 
     def _get_completion_params(
```
notthattal: Since you're adding created_at to the class you will most likely need to add it to the exclude in this function in order to properly send it to liteLLM without getting errors. You can verify it's working by sending a request to the /completions route for a saved prompt. Although since you're not serializing the created_at right now I think it will work no matter what?

PR #626, opened 2025-10-20
Dataset Enhancements

alexshanabrook:
Changes discussed in stand up:

1. last updated should represent when a version was created, not just the dataset metadata changes
2. sort list datasets endpoint by last modified date by default
3. dataset rows should have created_at date to sort by
4. add latest version to the dataset response object 
5. add hard limits to dataset size (250 rows)

### DIFF
--- a/genai-engine/tests/unit/routes/datasets/test_dataset_versions_crud.py
+++ b/genai-engine/tests/unit/routes/datasets/test_dataset_versions_crud.py
@@ -286,17 +286,6 @@ def test_dataset_versions_basic_functionality(
     # The 'profession' column should be removed since Alice Brown was the only row with that column
     assert set(version_3.column_names) == {"name", "age"}
 
-    # Test 5: Verify deleting dataset with versions doesn't result in an error
-    status_code = client.delete_dataset(dataset_id)
-    assert status_code == 204
-
-    # Test 6: Verify getting dataset version for deleted dataset returns error code
-    status_code, _ = client.get_dataset_version(
-        dataset_id=dataset_id,
-        version_number=1,
-    )
-    assert status_code == 404
-
     # test exceeding max value of allowed rows
     new_rows = [
         NewDatasetVersionRowRequest(
@@ -307,7 +296,7 @@ def test_dataset_versions_basic_functionality(
                 ),
             ],
         )
-        for i in range(248)
+        for i in range(249)
     ]
 
     status_code, _ = client.create_dataset_version(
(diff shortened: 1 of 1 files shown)


genai-engine/alembic/versions/2025_10_20_1302-7fb45748e807_add_dataset_row_created_at_and_latest_.py
```diff
@@ -0,0 +1,59 @@
...
+        SET latest_version_number = (
+            SELECT MAX(version_number)
+            FROM dataset_versions
+            WHERE dataset_versions.dataset_id = datasets.id
+        )
+        WHERE EXISTS (
+            SELECT 1 FROM dataset_versions
+            WHERE dataset_versions.dataset_id = datasets.id
+        )
+        """,
```
mcgrawia: two quick questions on this:
1. why do you need the WHERE EXISTS if the subquery above already has a join on dataset_id?
2. should we make `latest_version_number` non-null after this? Or is there possibly a state when the dataset is first created that there are no versions?

genai-engine/src/schemas/internal_schemas.py
```diff
@@ -2046,23 +2055,32 @@ def _from_request_model(
...
                 ],
+                created_at=curr_time,
             )
             for new_row in new_version.rows_to_add
         ]
         all_rows = unchanged_rows + new_rows + updated_rows
 
+        if len(all_rows) > MAX_DATASET_ROWS:
+            raise HTTPException(
+                status_code=404,
```
mcgrawia: would a 400 bad request make more sense than a 404 here?

genai-engine/src/schemas/internal_schemas.py
```diff
@@ -2046,23 +2055,32 @@ def _from_request_model(
                             DatasetVersionRowColumnItem._from_request_model(row_item)
                             for row_item in updated_row.data
                         ],
+                        created_at=existing_row_id_to_row[updated_row.id].created_at,
```
mcgrawia: I'm wondering for updated rows, if we should set the timestamp to the edit time? from a user perspective I'm trying to decide if I would expect rows I edited to also appear at the top, or stay in their old position with the unchanged rows? @xV3rt not sure if you have thoughts on this?

On second thought maybe leaving them in the original positions would make sense? Otherwise a user would have to go find where it ended up, vs they probably know where it was in the current position from having edited it?

alexshanabrook: Yeah—I was thinking it might make more sense if they just stay where they were when the user edited them, but let me know if you disagree @xV3rt , I can see both sides

genai-engine/src/schemas/internal_schemas.py
alexshanabrook: oops yes thanks!

genai-engine/src/schemas/internal_schemas.py
xV3rt: Definitely leave at the current positions. Main purpose is to keep new rows at the bottom for good UX. If we want to allow further order shenanigans, we should probably support drag and drop reordering in the future

mcgrawia (approved):

PR #628, opened 2025-10-20
UP-3100 add page for model provider management

mcgrawia:
# New UX for model provider management

Adds a page to view all model providers and their configuration status:
Screenshot 2025-10-22 at 4 31 32 PM

Adds a modal to allow setting the API key for a provider:
Screenshot 2025-10-22 at 4 31 42 PM

Adds a delete confirmation before deleting the credentials for a provider
Screenshot 2025-10-22 at 4 31 47 PM

# New UX for creating tasks on the home page

This wasn't part of the ticket, but it annoyed me so made UX to create the first task on a new DB install :)
Screenshot 2025-10-22 at 4 33 34 PM

Also made a button to add new tasks when some already exist:
Screenshot 2025-10-22 at 4 33 40 PM

### DIFF
--- a/genai-engine/ui/src/App.tsx
+++ b/genai-engine/ui/src/App.tsx
@@ -63,16 +63,16 @@ function App() {
                 }
               />
 
-                <Route
-                  path="/tasks/:id/model-providers"
-                  element={
-                    <ProtectedRoute>
-                      <TaskLayout>
-                        <ModelProviders />
-                      </TaskLayout>
-                    </ProtectedRoute>
-                  }
-                />
+              <Route
+                path="/tasks/:id/model-providers"
+                element={
+                  <ProtectedRoute>
+                    <TaskLayout>
+                      <ModelProviders />
+                    </TaskLayout>
+                  </ProtectedRoute>
+                }
+              />
 
               <Route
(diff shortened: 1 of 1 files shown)


notthattal (dismissed):
looks good to me

mcgrawia:
@martin-arthur-ai thanks for flagging. Exactly the kind of comment I was expecting. I had cursor help with this, and don't really know MUI, so will gladly switch it to use that instead. Are there any other components I should pull from MUI?

TheLukaszNs-Blazity:
@mcgrawia I'd say we should pull most of the UI components (`Input`s, `Button`s, etc.) from MUI to make the app consistent and, as Martin said, easy to share between SaaS and OS apps. 
It may be worth adding some rule for the Cursor agent to use components from MUI directly, since right now it seems to only do what it remembers best - copy shadcn/ui components

martin-arthur-ai:
@TheLukaszNs-Blazity @notthattal I converted a few remaining elements to MUI. This should be good to go.

**Before**
Screenshot 2025-10-23 at 14 31 35

**After**
Screenshot 2025-10-23 at 14 40 34

PR #629, opened 2025-10-20
add versioning for agentic prompts

notthattal:
## Description
- Adds an integer version to each agentic prompt
- Modifies the route to get a prompt to get a specific version (either 'latest', by timestamp to the nearest second or by version number) and adds a flag to the route for whether to include deleted versions or not
- Modifies the route to get all prompts by adding the flag to the route for whether to include deleted versions or not
- Modifies the route to get all prompt versions by adding the flag to the route for whether to include deleted versions or not
- Adds a prompt /{task_id}/agentic_prompts/names that gets all unique prompt names for a given task
- Adds verifications in run_chat_completion and stream_chat_completion to check if a prompt has been deleted before attempting to run it
- Modifies the save_prompt function to increment the version for a given prompt when trying to save a prompt with the same name for a particular task
- Modifies the delete endpoint to permanently delete all versions of a given prompt
- Modifies the delete_versions endpoint to soft-delete a singular version of an agentic prompt. Once soft-deleted all parameters are cleared except the necessary ones (i.e. task_id, name, created_at, deleted_at, and version number)
- Adds unit tests to verify functionality to both test_agentic_prompt_routes and test_agentic_prompt_repository

## Jira Tickets:
1.  https://arthurai.atlassian.net/browse/UP-3188
2. https://arthurai.atlassian.net/browse/UP-3185

### DIFF
--- a/genai-engine/src/routers/v1/agentic_prompt_routes.py
+++ b/genai-engine/src/routers/v1/agentic_prompt_routes.py
@@ -73,7 +73,7 @@ def get_validated_agentic_task(
 def prompts_get_all_filter_parameters(
     prompt_names: Optional[list[str]] = Query(
         None,
-        description="Prompt names to filter on using partial matching. If provided, prompts matching any of these name patterns will be returned. Supports SQL LIKE pattern matching with % wildcards.",
+        description="Prompt names to filter on using partial matching. If provided, prompts matching any of these name patterns will be returned.",
     ),
     model_provider: Optional[str] = Query(
         None,
@@ -81,7 +81,7 @@ def prompts_get_all_filter_parameters(
     ),
     model_name: Optional[str] = Query(
         None,
-        description="Filter by model name (e.g., 'gpt-4', 'claude-3-5-sonnet'). Supports SQL LIKE pattern matching with % wildcards.",
+        description="Filter by model name (e.g., 'gpt-4', 'claude-3-5-sonnet').",
     ),
     created_after: Optional[str] = Query(
         None,
@@ -111,7 +111,7 @@ def prompts_get_versions_filter_parameters(
     ),
     model_name: Optional[str] = Query(
         None,
-        description="Filter by model name (e.g., 'gpt-4', 'claude-3-5-sonnet') Supports SQL LIKE pattern matching with % wildcards.",
+        description="Filter by model name (e.g., 'gpt-4', 'claude-3-5-sonnet').",
     ),
     created_after: Optional[str] = Query(
         None,


genai-engine/alembic/versions/2025_10_20_1422-96e667d4e2af_add_versioning_to_agentic_prompts.py
```diff
@@ -0,0 +1,65 @@
...
+    op.create_primary_key(
+        "agentic_prompts_pkey",
+        "agentic_prompts",
+        ["task_id", "name", "version"],
+    )
+    op.create_unique_constraint(
+        "uq_task_prompt_version",
+        "agentic_prompts",
+        ["task_id", "name", "version"],
+    )
```
mcgrawia: did you have amebic generate this migration, or did you write it manually? If manually, we should make sure we're auto-generating the migrations first so they stay up to date with the database objects: https://github.com/arthur-ai/arthur-engine/blob/UP-3100-add-screen-for-model-provider-management/genai-engine/README.md#generate-a-new-alembic-migration

that said, I don't think you need an explicit unique constraint here, the primary key constraint enforced uniqueness

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -17,39 +20,184 @@ def __init__(self, db_session: Session):
...
-            .filter(
-                DatabaseAgenticPrompt.task_id == task_id,
-                DatabaseAgenticPrompt.name == prompt_name,
-            )
-            .first()
+    def get_prompt(
+        self,
+        task_id: str,
+        prompt_name: str,
+        prompt_version: str = "latest",
```
mcgrawia: based on the comment below, it doesn't seem like this is only a string type? I think you need a union of `str | int | datetime`?

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -58,22 +206,122 @@ def save_prompt(self, task_id: str, prompt: AgenticPrompt | Dict[str, Any]) -> N
...
+    def soft_delete_prompt(
+        self,
+        task_id: str,
+        prompt_name: str,
+        prompt_version: str,
+    ) -> bool:
+        """
+        Soft delete a prompt by task_id, name and version. This will delete all the data associated with the prompt
+        except for task_id, name, created_at, deleted_at and version. If all versions of a prompt are soft-deleted,
+        this will permanently delete all versions from the database.
```
mcgrawia: to help clarity, I would rename this function and description to deleting a "prompt version" instead of "prompt". I would expect deleting a prompt to delete all versions

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -58,22 +206,122 @@ def save_prompt(self, task_id: str, prompt: AgenticPrompt | Dict[str, Any]) -> N
...
             raise ValueError(
-                f"Prompt '{prompt.name}' already exists for task '{task_id}'",
+                f"Failed to save prompt '{prompt.name}' for task '{task_id}' — possible duplicate constraint.",
+            )
+
+    def soft_delete_prompt(
+        self,
+        task_id: str,
+        prompt_name: str,
+        prompt_version: str,
```
mcgrawia: same comment as above, we should define an explicit type for PromptVersion that is used everywhere this is passed, including in request objects and responses

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -58,22 +206,122 @@ def save_prompt(self, task_id: str, prompt: AgenticPrompt | Dict[str, Any]) -> N
...
+                )
+            except ValueError:
+                raise ValueError(
+                    f"Invalid prompt_version format '{prompt_version}'. Must be 'latest', "
+                    f"a version number, or an ISO datetime string.",
+                )
+
+        if not db_prompt:
+            raise ValueError(
+                f"No matching version of prompt '{prompt_name}' found for task '{task_id}'",
```
mcgrawia: this section seems like a lot of repeated code for adding version filters to a prompt query. Is it possible to modularize this so it can be reused wherever we look up a prompt by version?

genai-engine/src/routers/v1/agentic_prompt_routes.py
mcgrawia: I think this type also needs to be updated. Additionally, we should have a Path annotation parameter for this value with a description that includes the details about what is acceptable for a prompt version similar to what you put in the repository functions. An example:

```python
from fastapi import FastAPI, Path

app = FastAPI()

@app.get("/items/{item_id}")
async def read_item(
    item_id: int = Path(
        ...,  # This indicates the parameter is required
        description="The ID of the item to retrieve. This must be an integer.",
        title="Item ID",
        ge=1 # Example of a numeric validation: greater than or equal to 1
    )
):
    """
    Retrieve a single item by its ID.
    """
    return {"item_id": item_id}
```

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -119,10 +131,40 @@ def get_all_agentic_prompts(
...
+            include_deleted=include_deleted,
+        )
+    except ValueError as e:
+        raise HTTPException(status_code=400, detail=str(e))
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+
+
+@agentic_prompt_routes.get(
+    "/{task_id}/agentic_prompts/names",
```
mcgrawia: I think this endpoint is getting at my comment above, but it shouldn't just be names. It should be names and metadata like in the screenshot above

genai-engine/src/routers/v1/agentic_prompt_routes.py
mcgrawia: just calling out here as well we should update the type for `prompt_version` to reflect it's more than a str

genai-engine/src/schemas/agentic_prompt_schemas.py
```diff
@@ -331,6 +343,9 @@ def run_chat_completion(
         llm_client: LLMClient,
         completion_request: PromptCompletionRequest = PromptCompletionRequest(),
     ) -> AgenticPromptRunResponse:
+        if self.has_been_deleted():
+            raise ValueError("This prompt has been deleted")
```
mcgrawia: small suggestion on the comment:

```
f"Cannot run completion for this prompt because it was deleted on: {self.deleted_at}"
```

genai-engine/src/routers/v1/agentic_prompt_routes.py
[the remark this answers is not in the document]
mcgrawia: an example in swagger: 
image

genai-engine/src/schemas/agentic_prompt_schemas.py
mcgrawia: I realize you didn't make this change, but can we move this field to the bottom of the definition, otherwise it's the first field shown in the schema, which doesn't make a ton of sense for a response-only field

Image

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -58,22 +206,122 @@ def save_prompt(self, task_id: str, prompt: AgenticPrompt | Dict[str, Any]) -> N
...
+            prompt_name: str - the name of the prompt
+            prompt_version: str - the version of the prompt
+
+        * Note - Supports:
+            - prompt_version='latest' -> marks latest version as deleted
+            - prompt_version=<string number> (e.g. '1', '2', etc.) -> marks that version as deleted
+            - prompt_version=<datetime> (YYYY-MM-DDTHH:MM:SS, checks to the second) -> marks version created at that time
+
+        Returns:
+            - True if all prompt versions for this prompthave been hard-deleted, False if just this one version was soft-deleted
```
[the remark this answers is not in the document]
notthattal: okay will remove

genai-engine/src/repositories/agentic_prompts_repository.py
notthattal: I think so

genai-engine/src/routers/v1/agentic_prompt_routes.py
mcgrawia: this request should return the prompt version object that was created. Right now it just returns a success message but the FE will want to use the data from the newly created prompt like version number, etc. Otherwise it's going to have to make two calls. One to create the version, and one to fetch it, but it won't know the version number that just got created to use

Image

genai-engine/src/routers/v1/agentic_prompt_routes.py
notthattal: would we even need this endpoint if it's also going to return the metadata? Seems like at that point it would just do the same thing as get_all_prompts. I mainly included this for UP-3185 "Create an endpoint to only fetch Prompt names"

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -322,10 +377,9 @@ async def run_saved_agentic_prompt(
     Returns:
         AgenticPromptRunResponse or StreamingResponse
     """
-    # TODO: Implement with versioning
     try:
         agentic_prompt_service = AgenticPromptRepository(db_session)
-        prompt = agentic_prompt_service.get_prompt(task.id, prompt_name)
+        prompt = agentic_prompt_service.get_prompt(task.id, prompt_name, prompt_version)
```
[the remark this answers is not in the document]
notthattal: it would return an error in retrieving the prompt. Having it in the completion itself was more of a fail-safe than anything

genai-engine/src/schemas/enums.py
```diff
@@ -167,6 +167,7 @@ class ModelProvider(str, Enum):
     ANTHROPIC = "anthropic"
     OPENAI = "openai"
     GEMINI = "gemini"
+    NONE = ""
```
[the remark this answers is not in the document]
notthattal: when we clear all of the metadata for a soft-deleted prompt, when we retrieve that prompt the AgenticPrompt class still has a model_provider field that's required and expects it to be a valid ModelProvider enum

genai-engine/src/routers/v1/agentic_prompt_routes.py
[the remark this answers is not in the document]
mcgrawia: also, double check what you're actually returning from the endpoint matches the doc. Right now the deletes return: 
```json
{
    "message": "All versions of generate-chart deleted successfully"
}
```
but the docs don't expect a response (and when you switch over to 204 response with no content, it should look like the above^)

these are the current docs as an example:

Image

genai-engine/alembic/versions/2025_10_20_1422-96e667d4e2af_add_versioning_to_agentic_prompts.py
notthattal: Also if yes to ^ also drop it here?

```
# Ensure name is unique within each task
   __table_args__ = (
      UniqueConstraint("task_id", "name", "version", name="uq_task_prompt_version"),
)
```

genai-engine/src/repositories/agentic_prompts_repository.py
notthattal: Just double checked and all types "latest", "integer" and "timestamp" all come in as string types

genai-engine/src/routers/v1/agentic_prompt_routes.py
mcgrawia: one thing I noticed about these completion endpoints, they seem to succeed even if I don't provide variable mappings for the expected template values.. I was kind of expecting these to complain if I didn't provide the expected templates.. what do you think? On one hand it may prevent outages when agents are running if they accidentally miss providing a template value, but on the other hand, it might result in subtle bugs where the agents keep executing without necessary information in the prompts

genai-engine/src/repositories/agentic_prompts_repository.py
mcgrawia: based on thread above, these types are strings end-to-end so no change needed

genai-engine/src/schemas/enums.py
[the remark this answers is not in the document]
mcgrawia: ok, in that case I think it may be fine to leave the provider configured for soft deleted versions. This enum likely will be used in the UI and other places that a users selects model providers to use, so I'd rather not introduce a user-facing None option that will be shown in a bunch of provider selection menus. The providers are well-known, so it's not possible for the user to enter sensitive data in a provider field, so it should be ok to leave set when deleting.

genai-engine/alembic/versions/2025_10_20_1422-96e667d4e2af_add_versioning_to_agentic_prompts.py
```diff
@@ -0,0 +1,65 @@
...
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "96e667d4e2af"
+down_revision = "897c39909e73"
```
mcgrawia: also just a reminder before you merge to double check this is still the latest revision. There's a couple MRs open right now all building on this one

https://github.com/arthur-ai/arthur-engine/pull/631

genai-engine/src/repositories/agentic_prompts_repository.py
notthattal: Updated in [this commit](https://github.com/arthur-ai/arthur-engine/pull/629/commits/c7a6f0a8e660fdd51486303a7004fb3c0ee5a0d6)

genai-engine/src/repositories/agentic_prompts_repository.py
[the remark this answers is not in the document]
notthattal: Removed in [this commit](https://github.com/arthur-ai/arthur-engine/pull/629/commits/c7a6f0a8e660fdd51486303a7004fb3c0ee5a0d6)

genai-engine/src/routers/v1/agentic_prompt_routes.py
notthattal: Added in [this commit](https://github.com/arthur-ai/arthur-engine/pull/629/commits/c7a6f0a8e660fdd51486303a7004fb3c0ee5a0d6)

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -91,15 +90,28 @@ def get_agentic_prompt(
     db_session: Session = Depends(get_db_session),
     current_user: User | None = Depends(multi_validator.validate_api_multi_auth),
     task: Task = Depends(get_validated_agentic_task),
+    include_deleted: bool = Query(
+        False,
+        description="returns all unique prompt names including soft-deleted prompts",
+    ),
```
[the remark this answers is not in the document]
notthattal: Latest is always non-deleted now with [this commit](https://github.com/arthur-ai/arthur-engine/pull/629/commits/c7a6f0a8e660fdd51486303a7004fb3c0ee5a0d6)

genai-engine/src/schemas/enums.py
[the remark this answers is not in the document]
notthattal: Removed in [this commit](https://github.com/arthur-ai/arthur-engine/pull/629/commits/c7a6f0a8e660fdd51486303a7004fb3c0ee5a0d6) and now we're not clearing the model_provider

genai-engine/alembic/versions/2025_10_20_1422-96e667d4e2af_add_versioning_to_agentic_prompts.py
notthattal: will do

genai-engine/src/routers/v1/agentic_prompt_routes.py
notthattal: Created follow-up MR ticket: https://arthurai.atlassian.net/browse/UP-3220

notthattal: Returning the saved agentic prompt after the "update prompt routes and response models" commit

genai-engine/alembic/versions/2025_10_20_1422-96e667d4e2af_add_versioning_to_agentic_prompts.py
notthattal: Updated this since it wouldn't have been at the head. The new file was uploaded in the "add pagination, sorting and filtering to list endpoints" commit

mcgrawia (dismissed):
I didn't review the code changes that closely as this MR has gotten pretty big, but tested out the endpoints locally and they're looking great and working well. Nice work 👏

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -61,6 +67,55 @@ def get_validated_agentic_task(
     return task
 
 
+def agentic_prompt_filter_parameters(
```
mcgrawia: one other small comment.. it seems like the same set of filters is used for both the `/task/id/prompts` and the `task/id/prompts/name/versions` APIs, but I don't think all filters are relevant to both endpoints? 

For example here's the filters for the get prompts for task below. It seems like most are relevant except the `exclude_deleted`, `min_version`, and `max_version` filters? Those seem to apply to the `task/id/prompts/name/versions` endpoint. 

Similarly, for the `task/id/prompts/name/versions` endpoint, the `prompt_names` filter doesn't seem relevant because all versions of a prompt will have the same name. It might be worth making two separate filter structs for these cases?

Image

One other point of naming, I might suggest changing the `start_time` and `end_time` to be something like `created_after` and `created_before`? It's not exactly clear what start and end are in this case.

notthattal: Updated in the "split filters into separate classes" commit.  I created a base class for the prompt filter requests so we could enforce including the filtering function for future filters we may want to create. I noticed there are a lot of duplicate variables still between them, but decided to keep them separate in case there's a different prompt filter request we create in the future that wouldn't necessarily need those variables. Let me know your thoughts

PR #635, opened 2025-10-21
UP-3199: Add linters and static type check as part of the workflow

Kacpersoftberry:
TODO:
fix mypy issues

### DIFF
--- a/.github/workflows/arthur-engine-workflow.yml
+++ b/.github/workflows/arthur-engine-workflow.yml
@@ -49,7 +49,7 @@ jobs:
         id: autoflake
         continue-on-error: true
         run: |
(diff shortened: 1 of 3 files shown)


ml-engine/src/ml_engine/job_executors/task_management_job_executors.py
```diff
@@ -42,6 +41,8 @@
 from tools.connector_constructor import ConnectorConstructor
 from tools.converters import common_to_client_put_dataset_schema
 
+import genai_client.exceptions
+
```
mcgrawia: same here, the tools packages should be recognized as local with the genai_client

mcgrawia (dismissed):

PR #641, opened 2025-10-22
feat: add tracing view with filtering

TheLukaszNs-Blazity:


genai-engine/ui/.vite/deps/_metadata.json
```diff
@@ -0,0 +1,8 @@
+{
+  "hash": "89bb71bb",
```
martin-arthur-ai: Are these files necessary for build, or can they be safely added to gitignore?

genai-engine/ui/src/components/traces/components/SpanTree/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+          alignItems="center"
+          spacing={1}
+          data-has-children={hasChildren ? "" : undefined}
+          className="group-data-selected:rounded-t rounded-b group-data-selected:data-open:data-has-children:rounded-b-none transition-all duration-75"
+          sx={{
+            pl: `${4 + level * 16}px`,
+            py: 0.5,
+            backgroundColor: isSelected ? "primary.main" : "transparent",
+            color: isSelected ? "primary.contrastText" : "text.primary",
+          }}
```
xV3rt: For `4 + level * 16`, could we also use constants? Will be easier to adjust indentation if ever needed. This maybe should even be a global constant, as I assume we will want to indent in more components in the future and we would prefer it to be consistent

genai-engine/ui/src/components/traces/components/filtering/filters-row.tsx
xV3rt: This file is rather large and has multiple components inside. Maybe it would be worth to split it up into separate files? I understand that those are not reusable, but think for readability it would help if we split the code into separate files

genai-engine/ui/src/components/traces/components/CommonDrawer.tsx
```diff
@@ -0,0 +1,122 @@
...
+          >
+            {history.slice(0, history.length - 1).map((entry) => (
+              <Button
+                key={entry.id}
+                variant="text"
+                onClick={() => handleBreadcrumbNavigation(entry)}
+              >
+                {entry.for} ({entry.id})
+              </Button>
+            ))}
```
xV3rt: Here since we combine different entities, if Im not mistaken `entry.id` might not be unique. In that case, it would probably be safer to create the key as a combination of entry.id and entry.for

genai-engine/ui/package.json
```diff
@@ -56,5 +70,8 @@
     "typescript": "~5.9.0",
     "typescript-eslint": "^8.44.0",
     "vite": "^7.1.7"
+  },
+  "optionalDependencies": {
+    "@rollup/rollup-linux-x64-gnu": "4.52.2"
   }
```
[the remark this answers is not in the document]
TheLukaszNs-Blazity: I believe the docker image couldn't build without this before

genai-engine/ui/src/components/traces/components/filtering/filters-row.tsx
TheLukaszNs-Blazity: Yeah I don't know, the `createFilterRow` function provides nice type handling for all the instances. We can move some reusable components to their own files later

genai-engine/ui/package.json
```diff
@@ -25,17 +26,30 @@
     "@mui/icons-material": "^7.3.4",
     "@mui/material": "^7.3.3",
     "@tanstack/react-form": "^1.23.7",
+    "@tanstack/react-pacer": "^0.16.4",
```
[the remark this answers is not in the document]
fryz: +1 to martin - we want to avoid future issues w/ migrating things into the shared component library and should stick with packages that we're already using to avoid any future issues when we go to bring the functionality into our saas application. 

please do as martin is asking and use existing libraries where possible

PR #642, opened 2025-10-22
Rag Provider Connection Support

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-3197


genai-engine/alembic/versions/2025_10_22_1209-5a02e38003af_create_rag_configs.py
```diff
@@ -0,0 +1,70 @@
...
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "5a02e38003af"
+down_revision = "7fb45748e807"
```
mcgrawia: just flagging I just approved Videet's MR that adds a new migration so that may merge first

genai-engine/src/repositories/rag_providers_repository.py
```diff
@@ -0,0 +1,168 @@
...
+            raise HTTPException(
+                status_code=404,
+                detail="RAG provider configuration not found",
+            )
+        return db_config
+
+    def get_rag_provider_configuration(
+        self,
+        config_id: UUID,
+    ) -> Optional[RagProviderConfiguration]:
```
mcgrawia: does this ever return None? it seems like it would raise an Exception or return a config object?

alexshanabrook: no thanks!

mcgrawia (approved):

PR #653, opened 2025-10-23
UP-3147 Load prompt from span

martin-arthur-ai:
[UP-3147](https://arthurai.atlassian.net/browse/UP-3147?atlOrigin=eyJpIjoiZDA4OGQwZTcxZmI2NDM1NmJhYzkxZjUyNDNkN2JjMTgiLCJwIjoiaiJ9)

Ability to pull a span ID from the URL and load the data into a prompt. Also added some CSS changes and stubbed out functionality that needs to be fixed in the versions branch. The backend endpoint signatures changed.

### DIFF
--- a/genai-engine/ui/src/components/prompts/PromptsPlayground.tsx
+++ b/genai-engine/ui/src/components/prompts/PromptsPlayground.tsx
@@ -10,11 +10,14 @@ import { useSearchParams } from "react-router-dom";
 import PromptComponent from "./PromptComponent";
 import { PromptProvider } from "./PromptContext";
 import { promptsReducer, initialState } from "./reducer";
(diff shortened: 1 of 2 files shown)


genai-engine/ui/src/components/prompts/sampleSpans.ts
notthattal: What is this file used for? I couldn't find any of the samples used anywhere outside of this file

notthattal: What is this file used for? I couldn't find any of the samples used anywhere outside of this file

notthattal: If you're still planning on using it in the future I think it's fine

PR #671, opened 2025-10-28
UP-3199: Fixup mypy for GenAI

Kacpersoftberry:


genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -38,27 +42,25 @@ def _get_latest_db_prompt(self, base_query: Query) -> DatabaseAgenticPrompt:
...
+            target_epoch = target_dt.timestamp()
+            created_at_epoch = DatabaseAgenticPrompt.created_at.timestamp()
             return (
                 base_query.filter(
-                    sa.func.abs(
-                        sa.func.extract("epoch", DatabaseAgenticPrompt.created_at)
-                        - sa.func.extract("epoch", target_dt),
-                    )
-                    < 1,
+                    sa.func.abs(created_at_epoch - target_epoch) < 1,
```
mcgrawia: curious, why did the original code need to change?

genai-engine/src/repositories/inference_repository.py
```diff
@@ -279,12 +279,14 @@ def query_inferences(
...
+            inference_stmt = inference_stmt.order_by(desc(DatabaseInference.created_at))
         elif sort == PaginationSortMethod.ASCENDING:
-            stmt = stmt.order_by(asc(DatabaseInference.created_at))
-        results = stmt.all()
+            inference_stmt = inference_stmt.order_by(asc(DatabaseInference.created_at))
+        results: list[DatabaseInference] = inference_stmt.all()
 
         inferences = [Inference._from_database_model(di) for di in results]
+        if not inferences:
+            return [], 0
```
mcgrawia: why is this return needed? IIUC it seems like inferences will always be some sort of list, if it's an empty list, I'd expect count to also be zero, so this feels like the same values would be returned in the line below

genai-engine/src/custom_types/custom_types.py
```diff
@@ -0,0 +1,8 @@
+from typing import Any, Callable, ParamSpec, TypeVar
+
+from sqlalchemy.orm import Query
+
+QueryT = TypeVar("QueryT", bound=Query)
+FunctionT = TypeVar("FunctionT", bound=Callable[..., Any])
+P = ParamSpec("P")
+T = TypeVar("T")
```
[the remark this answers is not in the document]
Kacpersoftberry: I needed to create a custom types so mypy would not raise errors. In case:
- T: this one is created for decorator, so basically it tells that whatever comes in goes out as a type (we don't modify the return type of the function)
- P: same situation as with T, but this time refers to parameters
Those two types are created only for decorator that we are using if every endpoint is either secured or public.

- FunctionT: this is a decorator for function where we operate to check if api_key has permission for specific endpoint. This one is the helper there
- QueryT: this type I've added so to simplify returning of the Query type when returning from function that aggregates the filters

genai-engine/src/repositories/agentic_prompts_repository.py
mcgrawia: @Kacpersoftberry I think maybe the point of confusion is these functions do not actually perform the subtraction. They just generate the SQL that performs the subtraction when the query is executed. One of the values in the subtraction is a column in the DB, so it can't be performed in python

genai-engine/tests/unit/repositories/test_agentic_prompts_repository.py
```diff
@@ -1097,3 +1097,203 @@ async def test_stream_deleted_prompt_spawns_error(
         "event: error\ndata: Cannot stream chat completion for this prompt because it was deleted on"
         in events[0]
     )
+
+
+@pytest.mark.unit_tests
+def test_get_db_prompt_by_datetime_success(agentic_prompt_repo, mock_db_session):
```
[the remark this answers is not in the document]
mcgrawia: or if there's cases missing, we should extend those tests rather than write new ones

genai-engine/src/repositories/agentic_prompts_repository.py
Kacpersoftberry: Problematic line was the previous one, the one that extract from a model. Modified that. Also added TC that covers this one.

mcgrawia (dismissed):

genai-engine/src/repositories/agentic_prompts_repository.py
```diff
@@ -38,25 +42,31 @@ def _get_latest_db_prompt(self, base_query: Query) -> DatabaseAgenticPrompt:
...
         prompt_version: str,
-    ) -> DatabaseAgenticPrompt:
+    ) -> Optional[DatabaseAgenticPrompt]:
         try:
+            from datetime import timezone
+
             target_dt = datetime.fromisoformat(prompt_version)
+            # Treat the datetime as UTC
+            if target_dt.tzinfo is None:
+                target_dt = target_dt.replace(tzinfo=timezone.utc)
```
Kacpersoftberry: @notthattal I've experience strange issue while running tests on my local machine. The diff between those was 3600 seconds (1 hour, looks like timezone issue). I've added this replace to always use UTC. Can you confirm if this will work?

mcgrawia: this comes from the user right? I think it's safe to assume if a user doesn't specify a timezone in their request that we treat it as UTC

genai-engine/src/routers/v1/agentic_prompt_routes.py
```diff
@@ -94,6 +94,11 @@ def get_agentic_prompt(
     except ValueError as e:
         if "not found" in str(e).lower():
             raise HTTPException(status_code=404, detail=str(e))
+        elif "attempting to retrieve a deleted prompt" in str(e).lower():
+            raise HTTPException(
+                status_code=400,
+                detail="Attempting to retrieve a deleted prompt",
+            )
```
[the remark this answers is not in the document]
Kacpersoftberry: removed

genai-engine/tests/unit/routes/conftest.py
```diff
@@ -22,3 +26,37 @@ def changed_user_client(
...
+        model_provider="openai",
+        temperature=0.7,
+        max_tokens=100,
+    )
+    _, agentic_prompt = client.create_agentic_prompt(
+        task_id=create_agentic_task.id,
+        prompt_name=prompt_name,
+        prompt_data=prompt_data,
+    )
+    yield agentic_prompt
```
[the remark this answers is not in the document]
Kacpersoftberry: Fixed that. Those methods are now properly used in tests.

genai-engine/tests/clients/base_test_client.py
```diff
@@ -3143,6 +3144,41 @@ def delete_rag_search_setting_version(
...
+        resp = self.base_client.get(
+            f"/api/v1/tasks/{task_id}/prompts/{prompt_name}/versions/{version}",
+            headers=self.authorized_user_api_key_headers,
+        )
+        log_response(resp)
+        return (
+            resp.status_code,
+            AgenticPrompt.model_validate(resp.json()),
+        )
+
```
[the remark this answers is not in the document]
mcgrawia: I'm still seeing these?

PR #681, opened 2025-10-29
UP-3235: Use postgres version 15

Kacpersoftberry:


ntatsumi (changes requested):
Can we consider aligning the DB version to what's on CFT? They should be upgraded together and we also need to think about how the upgrade is performed with CFT. Users could run into issues as we're testing on the Postgres version in CFT (i.e. 15).

PR #683, opened 2025-10-29
fix messages exporting incorrectly

notthattal:
## Description
- replaces the current way we are setting content to be exported as spans to adhere to open-inference formatting
- Fixes the case where a singular text content is incorrectly passed in as an array but still set to content (i.e. openai formatted):
```json
[{
    "type": "text",
    "text": "test"
}]
```
so we still export the content like this:
`"llm.input_messages.1.message.content": "test"`
instead of how we were exporting it before:
`"llm.input_messages.1.message.content": "[{\"type\":\"text\",\"text\":\"test\"}]"`


mcgrawia (approved):
thanks for fixing!

PR #691, opened 2025-10-30
UP-3207: More RAG search funtionality

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-3207

This just adds the keyword search, but I want to make smaller PRs now that the scaffolding exists.

I originally was planning on using separate response objects for the similarity vs keyword vs hybrid search results in case they ever diverge. Thinking about it more though, they're all querying the exact same vector store & objects, so I can't see why they would ever return different object structures—today they're shared in weaviate. I consolidated them in this PR since the old similarity-specific types were never picked up by the FE, so I don't think the breaking change matters.


genai-engine/src/clients/rag_providers/weaviate_client.py
```diff
@@ -80,31 +116,45 @@ def vector_similarity_text_search(
...
+        collection = self.client.collections.use(weaviate_settings.collection_name)
+
+        # construct parameters to bm25 query function
+        weaviate_settings_dict = weaviate_settings.model_dump(
+            exclude={
+                "collection_name",
+                "rag_provider",
+                "minimum_match_or_operator",
+                "and_operator",
+            },
```
mcgrawia: this `exclude` makes me a bit nervous if we ever add settings to this object in the future, we have to remember to come back to these query functions and black list the new values. Instead, could we create functions on the object that convert the settings to a dict for each search method? I think that allows for two things:
1. it's a white list instead of a black list, so new values would need to be added to be passed to the search methods
2. the functions are defined right by the members of the settings, so it's easy for devs to see when they update the settings object

mcgrawia (commented):
actually, one more thing I'm noticing. In the similarity search, the `query` is a list of strings, while in the keyword search, the query is just a string? Is this intentional?

```json
{
  "settings": {
    "collection_name": "string",
    "limit": 0,
    "include_vector": false,
    "offset": 0,
    "auto_limit": 0,
    "return_metadata": [
      "creation_time"
    ],
    "return_properties": [
      "string"
    ],
    "rag_provider": "weaviate",
    "query": [
      "string"
    ],
    "certainty": 1,
    "distance": 0,
    "target_vector": "string"
  }
}
```

vs
```json
{
  "settings": {
    "collection_name": "string",
    "limit": 0,
    "include_vector": false,
    "offset": 0,
    "auto_limit": 0,
    "return_metadata": [
      "creation_time"
    ],
    "return_properties": [
      "string"
    ],
    "rag_provider": "weaviate",
    "query": "string",
    "minimum_match_or_operator": 0,
    "and_operator": true
  }
}
```

Edit: Nevermind, looks like this is by design: https://docs.weaviate.io/weaviate/api/graphql/search-operators#concept-parsing

mcgrawia (dismissed):

mcgrawia (approved):

PR #700, opened 2025-10-31
Add token count and cost

videetparekh:


genai-engine/alembic/versions/2025_10_31_1141-332148919437_add_token_count_cost_columns_to_spans.py
```diff
@@ -0,0 +1,246 @@
...
+        text(
+            """
+        UPDATE spans
+        SET
+            prompt_token_count = (raw_data->'attributes'->'llm'->'token_count'->>'prompt')::integer,
+            completion_token_count = (raw_data->'attributes'->'llm'->'token_count'->>'completion')::integer,
+            total_token_count = COALESCE(
+                (raw_data->'attributes'->'llm'->'token_count'->>'total')::integer,
+                (raw_data->'attributes'->'llm'->'token_count'->>'prompt')::integer +
+                (raw_data->'attributes'->'llm'->'token_count'->>'completion')::integer
```
mcgrawia: not sure how likely this is but what happens if one of the prompt or completion token counts is null while the other is not? I think we'd still want to report the total as the sum of all non-null counts?

genai-engine/alembic/versions/2025_10_31_1141-332148919437_add_token_count_cost_columns_to_spans.py
```diff
@@ -0,0 +1,246 @@
...
+        "trace_metadata",
+        ["total_token_count"],
+        unique=False,
+    )
+    op.create_index(
+        "idx_traces_total_token_cost",
+        "trace_metadata",
+        ["total_token_cost"],
+        unique=False,
+    )
```
mcgrawia: I think we should add these columns also on the session metadata and trace users tables too right? those views will want cost summaries in the UI

mcgrawia: ah nvm, forgot these were just aggregations and not persistent tables 👍

genai-engine/src/utils/trace.py
```diff
@@ -304,3 +305,103 @@ def validate_span_version(raw_data: Dict[str, Any]) -> bool:
...
+        return result
+
+    try:
+        attributes = span_raw_data.get("attributes", {})
+
+        # Extract token counts
+        token_count = get_nested_value(attributes, "llm.token_count") or {}
+        prompt_tokens = token_count.get("prompt")
+        completion_tokens = token_count.get("completion")
+        total_tokens = token_count.get("total")
```
mcgrawia: is it possible to use the semantic conventions instead of strings here? Wondering if we can instead call get nested value 3x and use the constants:
```python
prompt_tokens = get_nested_value(attributes, SemanticConventions.LLM_TOKEN_COUNT_PROMPT)
```

mcgrawia (commented):
overall looks good. I think we need to add some tests though. I'm particularly curious what would happen when receiving spans with only a subset of the token count attributes present. I want to make sure if the agent instrumentation is partially working (only reporting prompt or completion counts), we still compute total counts and cost ok so we can show users as much as we've received. I don't think it would make sense to skip or error calculating totals and cost if we're missing attributes. Best practice would be to assume we'll have messy data, so we should make sure we can handle missing attributes 👍

genai-engine/src/schemas/internal_schemas.py
```diff
@@ -1631,11 +1631,14 @@ def input_content(self) -> Optional[str]:
         Uses get_nested_value for safe nested dictionary access.
         Converts dicts/lists to JSON strings to ensure string return type.
         """
+        value = trace_utils.get_nested_value(
+            self.raw_data,
+            "attributes.input.value",
```
mcgrawia: +1 on using the SemConv constants here

genai-engine/src/utils/trace.py
```diff
@@ -112,8 +112,12 @@ def extract_span_features(span_dict):
         attributes = span_dict.get("attributes", {})
 
         # Access already-normalized nested messages
-        input_messages = get_nested_value(attributes, "llm.input_messages") or []
-        output_messages = get_nested_value(attributes, "llm.output_messages") or []
+        input_messages = get_nested_value(attributes, "llm.input_messages", default=[])
+        output_messages = get_nested_value(
+            attributes,
+            "llm.output_messages",
```
mcgrawia: +1 on using the SemConv constants here and above

genai-engine/alembic/versions/2025_11_03_1039-aa259a314eeb_add_token_cost_and_input_output_content.py
```diff
@@ -0,0 +1,312 @@
...
+                    WHERE id = :span_id
+                """,
+                ),
+                {
+                    "prompt_cost": prompt_cost,
+                    "completion_cost": completion_cost,
+                    "total_cost": total_cost,
+                    "span_id": span_id,
+                },
+            )
```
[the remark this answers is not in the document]
fryz: Im thinking in a different direction - would prefer to present to the user in a way that made it obvious we were estimating. 

So we could either: 
* Present a range (lower bound input, upper bound output) 
* Go either direction but with some kind of tooltip indicating it's an estimate (and why it's an estimate) 

Ultimately I don't think it really matters if it's over or under (I could see an argument for either), but I do think it's important to call out that we had to estimate.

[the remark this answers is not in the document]
mcgrawia: copying suggestion from our slack convo:

discussed a small change to the API format that allows the BE to communicate to the FE when a total cost is estimated by using a polymorphic cost type:

```json
"counts": {
  "input": 50,
  "output": 50,
  "total": 50
},
"cost": {
  "type": "actual",
  "input": 0.0001,
  "output": 0.123,
  "total": 0.1231
}
// or in the case of missing input/output counts
"counts": {
  "input": null,
  "output": null,
  "total": 50
},
"cost": {
  "type": "estimated",
  "total": 0.1231
}
```

genai-engine/alembic/versions/2025_11_03_1039-aa259a314eeb_add_token_cost_and_input_output_content.py
```diff
@@ -0,0 +1,285 @@
...
+            raw_data->'attributes'->'input'->'value' as input_value,
+            raw_data->'attributes'->'output'->'value' as output_value
+        FROM spans
+        WHERE parent_span_id IS NULL
+        ORDER BY trace_id, start_time ASC
+    """,
+        ),
+    )
+
+    rows = result.fetchall()
```
mcgrawia: I think this also needs to be batched otherwise it's going to pull every trace from the DB into memory which is going to be huge

mcgrawia (approved):
awesome lgtm! let's make sure @TheLukaszNs-Blazity knows on the FE we need to update the cost/token count visuals to use your new fields 👍 

Image

PR #701, opened 2025-10-31
Add input/output_contents and remove span message properties

videetparekh:


genai-engine/src/services/trace/trace_ingestion_service.py
```diff
@@ -432,86 +431,80 @@ def _batch_upsert_trace_metadata(self, spans: list[DatabaseSpan]):
...
+            # Prefer new non-null values for input/output content
+            # This allows updates if a new batch provides a better root span
+            input_content=func.coalesce(
+                stmt.excluded.input_content,
+                DatabaseTraceMetadata.input_content,
+            ),
+            output_content=func.coalesce(
+                stmt.excluded.output_content,
+                DatabaseTraceMetadata.output_content,
+            ),
```
mcgrawia: curious, does this query basically take any input/output content from a span that is non-null? I think we'd want these fields to be only the input / output content from the root span. If there are multiple, it could pick the first non-null one. 

I don't think it would make sense for this to be a nested span's input / output because then the trace metadata might show some internal tool call as the trace-level summary, instead of the top level input / output which should be the users original message, and the agent's response

mcgrawia: ah whoops, GitHub linked me to only a single commit instead of the full MR. Seeing the code above that extracts input/output only if the span is a root span 👍

PR #717, opened 2025-11-03
UP-3198: Settings crud

alexshanabrook:
Jira ticket: https://arthurai.atlassian.net/browse/UP-3198

this is just the initial 3 endpoints, more to follow


genai-engine/alembic/versions/2025_11_03_1527-ebbf3cd9f512_create_rag_configs.py
```diff
@@ -0,0 +1,69 @@
...
+        unique=False,
+    )
+    op.create_table(
+        "rag_setting_configuration_versions",
+        sa.Column("setting_configuration_id", sa.UUID(), nullable=False),
+        sa.Column("version_number", sa.Integer(), nullable=False),
+        sa.Column("settings", postgresql.JSON(astext_type=sa.Text()), nullable=False),
+        sa.Column("tags", postgresql.JSON(astext_type=sa.Text()), nullable=True),
+        sa.Column("created_at", sa.TIMESTAMP(), nullable=False),
+        sa.Column("updated_at", sa.TIMESTAMP(), nullable=False),
```
mcgrawia: one thing we decided was important UX for prompt versions was that they should be soft deleted so the version history is maintained. As a result, @notthattal added a `deleted_at` field that represents a version was soft deleted. When a prompt version is soft deleted, we cleared all the user-configurable fields. I think we could do something similar here and clear the settings/tags for the deleted version?

Just want to make sure this, prompts, and evals all have the same versioning / lifecycle UX so we don't have different styles of versioning for different resources.

Long shot, but I wonder if it's possible to make an abstract object that makes the versioning on-rails for DB storage? Then all 3 resources could use that instead?

genai-engine/src/schemas/request_schemas.py
```diff
@@ -265,15 +278,12 @@ class RagVectorSimilarityTextSearchSettingRequest(BaseModel):
     )
 
 
-class WeaviateKeywordSearchSettingsRequest(WeaviateSearchCommonSettings):
+class WeaviateKeywordSearchSettingsConfigurationRequest(
+    WeaviateSearchCommonSettingsRequest,
+):
     rag_provider: Literal[RagProviderEnum.WEAVIATE] = RagProviderEnum.WEAVIATE
+    search_kind: Literal[RagSearchKind.KEYWORD_SEARCH] = RagSearchKind.KEYWORD_SEARCH
```
mcgrawia: Slightly confused on this.. we said for RAG providers we could have multiple providers of the same instance right? How do we specify which instance of a provider we want to use in these search settings? I see there's this provider field, but I don't see how to reference a specific weaviate provider?

From the request `POST /api/v1/tasks/{task_id}/rag_provider_settings`:
```json
{
  "settings": {
    "collection_name": "string",
    "limit": 0,
    "include_vector": false,
    "offset": 0,
    "auto_limit": 0,
    "return_metadata": [
      "creation_time"
    ],
    "return_properties": [
      "string"
    ],
    "rag_provider": "weaviate",
    "search_kind": "hybrid_search",
    "alpha": 0.7,
    "query_properties": [
      "string"
    ],
    "fusion_type": "FUSION_TYPE_RANKED",
    "max_vector_distance": 0,
    "minimum_match_or_operator": 0,
    "and_operator": true,
    "target_vector": "string"
  },
  "name": "string",
  "description": "string",
  "tags": [
    "string"
  ]
}
```

alexshanabrook: Hmm—my idea was that the settings configuration & the provider authentication configuration would be decoupled. That way someone could save a settings configuration that works for them but not need to specify the ID of the rag provider config they're using in case they want to use the same configuration against another provider. But thinking more about this, I guess they're not decoupled anyway since we're including `collection_name` in the config, which is specific to the vector store. So I can change the `rag_provider` field to be the ID of a provider config if that makes more sense

alexshanabrook: yeah, I think it makes sense to include the provider ID in that case—it also seems like what the users will be pursuing is "the settings config that works best" for their datasets, which might be different depending on the vector DB, collection, etc.—even if users have the same collection in multiple providers, they may use different vectorizers or other configs on the collections themselves, in which case the "setting that works best" may be different for the different providers. If there's no meaningful difference between the collection in multiple providers, why have them in both? 

plus, there's an easy path to making this an `optional` field if we get the feedback that it's inconvenient that these are tied to specific providers. I think the only way this might get annoying would be if someone wants to delete a provider config & keep the settings—eg. they rotated their API key—but in that case we can just encourage them to replace the API key in the existing provider config instead of creating a new one so they can still use their old settings

alexshanabrook: @mcgrawia Actually, in this case do you think it makes sense to scope these settings configuration endpoints under a Rag provider ID?

mcgrawia: I think it could work either way. In terms of resource scope, I have thought about these less as a sub-resource of a provider, and more as a stand-alone resource that references a provider, but I think they're equivalent.

One thing that may be easier to do if we don't nest them under a provider is being able to list all search config settings resources regardless of provider? Similar to prompts, I think users will want to see all their saved search configs in one list, where each one might reference a different provider? Thinking that endpoint could let you filter based on provider type as well as a specific provider instance if they wanted to narrow it down? 

Thoughts?

alexshanabrook: okay yeah makes sense, thanks! agreed if we want to support that view—also makes the path to making the provider optional in the future easier in case we need to

PR #737, opened 2025-11-06
UP-3281: Increase start period. Properly handle graceful termination

Kacpersoftberry:
- increase StartPeriod for engines (increase to 5 minutes)
- introduce health check grace period (set to 5 minutes)
- improve graceful handling of termination


deployment/cloudformation/ml-engine/arthur-ml-engine-ecs-task-definition.yml
```diff
@@ -121,7 +121,7 @@ Resources:
             Interval: 30
             Timeout: 5
             Retries: 3
-            StartPeriod: 60
+            StartPeriod: 300
```
mcgrawia: the ML engine shouldn't need this long of a start up period. 60 is fine

PR #742, opened 2025-11-06
add get requests for llm evals

notthattal:
## Description
- Adds the ability to get an llm eval version, list all versions of an llm eval, or list all llm evals for a specific task id
- Fixes a bug in agentic prompts that raised an error in list versions when no versions were found (it now just returns an empty object)
- Fixes some comments in the agentic prompts that weren't correct
- Adds unit tests for all get requests for both the router and repository for llm evals
- Modifies GetAllMetadata response object from being specific to AgenticPrompts to be more generic since both prompts and llm evals have the exact same metadata parameters so it can be reused.
- Modifies filter requests for agentic prompts to also be more generic since llm evals has the exact same filters 
- Extracts dependencies that were in the agentic prompt router file to the dependencies file so it can be reused by llm evals

## Jira Tickets
- https://arthurai.atlassian.net/browse/UP-3274

## TODO
- Implement the ability to run an llm eval
- Extract template evals for FE to use

### DIFF
--- a/genai-engine/src/schemas/request_schemas.py
+++ b/genai-engine/src/schemas/request_schemas.py
@@ -1,6 +1,6 @@
 from abc import ABC, abstractmethod
 from datetime import datetime
-from typing import Any, List, Literal, Optional, Union
+from typing import Any, List, Literal, Optional, Type, Union
 from uuid import UUID
 
 from fastapi import HTTPException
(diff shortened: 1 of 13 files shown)


genai-engine/src/schemas/request_schemas.py
```diff
@@ -633,35 +635,34 @@ def apply_filters_to_query(
             Query - the query with filters applied
         """
         # Filter by prompt names using LIKE for partial matching
-        if self.prompt_names:
+        if self.llm_asset_names:
             name_conditions = [
-                DatabaseAgenticPrompt.name.like(f"%{name}%")
-                for name in self.prompt_names
+                db_model.name.like(f"%{name}%") for name in self.llm_asset_names
```
mcgrawia: I didn't catch this in the original MR, but I think it's a bad practice to have the HTTP request objects reference/import/use the database objects. Only the repository layer should import the db objects or we risk introducing circular imports. I'm ok with merging this for now but we should make a tech debt ticket to move this code to the repo/internal schemas objects instead.

PR #751, opened 2025-11-07
[no-ref] fix migrations, remove commits

mcgrawia:
need to remove commits from migrations, it was causing alembic to run the migrations without marking the version as updated


mcgrawia:
closing for #752

PR #763, opened 2025-11-10
Fix GenAI client generation in ML engine build

alexshanabrook:
Follow up to build failures generated by this PR: https://github.com/arthur-ai/arthur-engine/pull/738/files#diff-6eb1562a25bfdc7454e1b9dfad230c1e42e403d09c3f51618ec0fdbfd292786fL345-R347

### DIFF
--- a/ml-engine/scripts/openapitools.json
+++ b/ml-engine/scripts/openapitools.json
@@ -8,8 +8,7 @@
   "inputSpec": "genai-engine/staging.openapi.json",
   "outputDir": "ml-engine/src/genai_client",
   "validateSpec": false,
-  "configOptions": {
-    "packageName": "genai_client",
-    "packageVersion": "{{info.version}}"
+  "additionalProperties": {
+    "packageName": "genai_client"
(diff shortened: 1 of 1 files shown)


ml-engine/scripts/openapitools.json
```diff
@@ -8,8 +8,7 @@
   "inputSpec": "genai-engine/staging.openapi.json",
   "outputDir": "ml-engine/src/genai_client",
   "validateSpec": false,
-  "configOptions": {
-    "packageName": "genai_client",
-    "packageVersion": "{{info.version}}"
+  "additionalProperties": {
```
alexshanabrook: the incorrect config name meant the package was being generated in the CICD with the package name `openapi_client`, which is why the ML engine couldn't find the expected genai_client package.

The packageVersion was not deserializing from the templating—the genai_client is being installed as version `1.0.0` in every ML engine build for now. That behavior existed before the breaking MR, so not sure if we need to fix it today

PR #772, opened 2025-11-13
Example of passing userId and sessionId in mastra traceOptions metadata

mcgrawia:
1. upgrade mastra and other components
2. create a custom MastraAgent subclass that passes in `userId` and `sessionId` to the `agent.stream` method. Verified they show up in the engine:

Screenshot 2025-11-12 at 11 27 00 PM
Screenshot 2025-11-12 at 11 27 03 PM

Traces also look much prettier now:

image

### DIFF
--- a/genai-engine/ui/yarn.lock
+++ b/genai-engine/ui/yarn.lock
@@ -1641,6 +1641,13 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@tanstack/devtools-event-client@npm:^0.3.5":
+  version: 0.3.5
+  resolution: "@tanstack/devtools-event-client@npm:0.3.5"
+  checksum: 10c0/10e4feec2cd346cd47adf94889f3f12d8662fb291d0d17fbf2a2f9c643f083a70d081b0fb104c2f54c0bd89dcfb1e31942dc959b74592fce9a7b9d55ed233220
+  languageName: node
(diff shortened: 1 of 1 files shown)


genai-engine/examples/agents/analytics-agent/src/app/lib/ag-ui/mastra.ts
```diff
@@ -0,0 +1,429 @@
...
+        const response = await this.agent.stream(convertedMessages, {
+          threadId,
+          resourceId,
+          runId,
+          clientTools,
+          runtimeContext,
+          tracingOptions: {
+            metadata: {
+              userId: "123",
+              sessionId: "456",
```
mcgrawia: @videetparekh here's where the userId/sessionId is passed to the `agent.stream` call. I think if you want to take this MR and hook up this so it's passed in from this class's constructor, then wire it up to pass them into the constructor from your telemetry headers in the request, this should work

genai-engine/examples/agents/analytics-agent/src/app/api/copilotkit/route.ts
mcgrawia: this is where you'd need to update to pass `userId` / `sessionId` in from the incoming request

videetparekh (approved):

PR #775, opened 2025-11-14
Prompt Playground Improvements

fryz:
This PR makes a handful of improvements to the prompt playground: 

1. Moves the variable pane to a dropdown from the top and relocates the add prompt + run all prompt buttons
2. Fixes dynamically resizing the individual prompt containers when adding new prompts depending on screen size (3 prompts for small screens, 5 prompts for large screens before scrolling)
3. Ensures that the format response doesn't get overwritten when running prompts
4. Shows when a prompt version has been modified (eg: dirty state) when making changes
5. Handles API errors when running prompts gracefully w/ a toast

Note: code written was vibe coded, but tested while developing. 

It may be better to review this PR commit by commit instead of as one entire unit as the changes listed above line up with the commits


genai-engine/ui/src/components/prompts-playground/prompts/PromptComponent.tsx
```diff
@@ -75,30 +81,64 @@ const Prompt = ({ prompt }: PromptComponentProps) => {
...
-            <Button variant="outlined" size="small" onClick={handleOpenResponseSchemaDialog}>
-              Format Response
-            </Button>
           </div>
           <div className="flex justify-end items-center gap-1 flex-shrink-0">
             <ManagementButtons prompt={prompt} setSavePromptOpen={setSavePromptOpen} />
           </div>
         </div>
-        <div className="min-w-[300px] flex-shrink-0">
+        <div className="flex-shrink-0" style={{ minWidth: 0 }}>
```
TheLukaszNs-Blazity: Style instead of className

xV3rt (approved):
Nitpicks, regarding the UI on specific screens, don't think it should be a blocker, but it is definitely worth flagging and creating a ticket for

PR #814, opened 2025-11-17
[UP-3238] [UP-3258] [UP-3293] [UP-3295] [UP-3330] [UP-3174] feat: add tracing tweaks - next batch

TheLukaszNs-Blazity:
[UP-3238]
[UP-3258]
[UP-3293]
[UP-3295]
[UP-3330]
[UP-3174]

### DIFF
--- a/genai-engine/ui/src/components/traces/components/DrawerPagination/index.tsx
+++ b/genai-engine/ui/src/components/traces/components/DrawerPagination/index.tsx
@@ -17,9 +17,12 @@ export const DrawerPagination = () => {
       </span>
     );
 
-  if (!context.ids.length) return <span className="text-sm text-gray-400">No context available...</span>;
+  if (!context.ids.length) return <span className="text-sm text-gray-400">No context available.</span>;
 
   const idx = context.ids.indexOf(current.target.id.toString());
+
+  if (idx === -1) return <span className="text-sm text-gray-400">Current item not found in context.</span>;
+
   const hasNext = idx < context.ids.length - 1;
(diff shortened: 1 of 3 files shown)


genai-engine/ui/src/components/traces/components/llm/ToolsTab.tsx
```diff
@@ -0,0 +1,76 @@
+import { Collapsible } from "@base-ui-components/react/collapsible";
```
martin-arthur-ai: I'm not sure why we're mixing libraries here, instead of using MUI

genai-engine/ui/src/components/traces/components/DrawerPagination/index.tsx
```diff
@@ -0,0 +1,64 @@
...
+      <span className="text-sm text-gray-400">
+        Currently navigating on <code className="font-mono bg-gray-700 text-white px-1 py-0.5 rounded-md">{context.type}</code> level.
+      </span>
+    );
+
+  if (!context.ids.length) return <span className="text-sm text-gray-400">No context available...</span>;
+
+  const idx = context.ids.indexOf(current.target.id.toString());
+  const hasNext = idx < context.ids.length - 1;
+  const hasPrevious = idx > 0;
```
xV3rt: I think we might want to explicitly handle the case when the index is not found in `context.ids`. For example:

```
if (idx === -1) return <span className="text-sm text-gray-400">Current item not found in context.</span>
```

PR #815, opened 2025-11-17
UP-3319: Fix mypy issues for GenAI Engine pt 2

Kacpersoftberry:


genai-engine/src/utils/claim_parser.py
```diff
@@ -22,14 +23,14 @@ def _deduplicate(self, seq: list[str]) -> list[str]:
         """
 
         seen = set()
-        return [x for x in seq if not (x in seen or seen.add(x))]
+        return [x for x in seq if not (x in seen or seen.add(x))]  # type: ignore[func-returns-value]
```
alexshanabrook: is there a reason you ignored this warning? I wonder if we could just rewrite it to be clearer with what it's doing—this may be coming from the fact that `set.add` doesn't have a return type but its return type is being evaluated here?

genai-engine/src/utils/utils.py
```diff
@@ -271,6 +271,24 @@ async def common_pagination_parameters(
     return PaginationParameters(sort=sort, page_size=page_size, page=page)
 
 
+@overload
```
alexshanabrook: @Kacpersoftberry sorry, I was thinking something more like this:

In python, you can make typing generics as follows:
`T = TypeVar("T", bound=Union[list[str], str])`

This creates a generic type variable that can either be a list of strings or strings.

If you then use that as the type hint for `pad_text`, without these overload changes, pydantic knows that you output the same sub-element of the type var that you pass in:

`text: T` and `-> T` will tell mypy that the type of the parameter, whether `list[str]` or `str`, will be the same as the type of the return value. That way I think this mypy check should pass without being overriden.

does that make sense? here are some docs if you're curious https://typing.python.org/en/latest/reference/generics.html

It's useful to do instead of overriding so you don't have to add an overload function for every individual type you're allowing the parameter to be bound to

genai-engine/src/utils/claim_parser.py
Kacpersoftberry: In the stackoverflow (the link in docstring) author of this method explains why it's faster then using the function in a way that we expect :) that's why I went with ignoring type here

genai-engine/src/utils/utils.py
alexshanabrook: sorry, why do you consider an overload stricter? If you set the generic type variable to be `bound` to a set of types, pydantic will still enforce that any parameters/return values fall in that set of types. It's the same as a parameter with a `Union` type except you can enforce that the return value is the same sub-type within that union as the parameter.

genai-engine/src/utils/claim_parser.py
alexshanabrook: okay, fine by me then. thanks for explaining!

PR #816, opened 2025-11-17
Fix message rendering. Add tool call prettification

videetparekh:
This fixes a bug in the message rendering where assistants making tool calls may (Langchain) or may not (Mastra) have message contents. It also adds some prettification to the tool call message by adding a `Formatted | Raw` toggle to see raw tool call JSON

### Formatted tool call message
image

### Raw tool call message
image

### Assistant message containing both content and tool calls ((with better object handling for the formatted view)
image

### Tool Result Message containing Formatted | Raw options (with better object handling for the formatted view)
image


martin-arthur-ai (approved):
LGTM

PR #837, opened 2025-11-19
Adding changes to make it easier to sync datasets from external sources

mcgrawia:
1. Adding an API that lets users run trace transforms on an existing dataset/trace 
2. Updates the create dataset version API to allow specifying a filter for rows to delete, in case the user wants to bulk delete some rows in a new version
3. fixes the add to dataset drawer so if the dataset has more columns than the transform, the extra columns are still displayed so the user can fill them in
Screenshot 2025-11-19 at 4 57 57 PM
Screenshot 2025-11-19 at 4 58 20 PM
Screenshot 2025-11-19 at 5 27 21 PM

### DIFF
--- a/genai-engine/ui/src/components/datasets/transforms/TransformDetailsModal.tsx
+++ b/genai-engine/ui/src/components/datasets/transforms/TransformDetailsModal.tsx
@@ -1,3 +1,4 @@
+import { useState } from "react";
 import {
   Dialog,
   DialogTitle,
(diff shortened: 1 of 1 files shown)


genai-engine/src/utils/transform_executor.py
```diff
@@ -0,0 +1,121 @@
...
+        # Find matching spans by span_name
+        matching_spans = [span for span in flat_spans if span.span_name == span_name]
+
+        if not matching_spans:
+            # No matching span found, use fallback
+            columns.append(
+                NewDatasetVersionRowColumnItemRequest(
+                    column_name=column_name,
+                    column_value=stringify_value(
+                        fallback if fallback is not None else ""
```
videetparekh: Is an empty value represented by an "" or None in the dataset?

videetparekh (approved):

PR #842, opened 2025-11-20
Eval editing

mcgrawia:
1. adding Edit button to eval details page
2. fixing a bug where only OOTB evals are shown in the creation modal
3. fixing a bug where creation a new eval doesn't open it's version page


notthattal (approved):
Looks good to me not sure if you also want martin's approval since this is a frontend change

PR #844, opened 2025-11-20
Feature/prompt studio

videetparekh:
# Summary

This PR implements experiment execution in the prompts playground when loaded with a config, enabling users to run experiments directly from the playground interface with full results tracking and visualization.

Users can now run unsaved prompts. 

Users can open prompt experiments in the playground using the Open In Notebook button. 

### Unsaved Prompts in Experiments and Experiment Views
image

### Open In Notebook
image

### Run unsaved and saved prompt
image

### Playground Config View
image


mcgrawia:
implemented in #859

PR #851, opened 2025-11-21
add default model provider to prompts and evals

notthattal:
## Description
- Add default model provider option to ModelProvider enum
- Create a global default model provider client that will be used if default model provider is selected for either prompts or evals
- Made the default llm client just override any model providers/model names that don't match if default is selected
- Added Azure as one of the available options for ModelProvider since that's what the default llm client would need its provider set to if GENAI_ENGINE_OPENAI_PROVIDER=Azure in a user's .env
- Add unit tests 

## Jira Ticket
- https://arthurai.atlassian.net/browse/UP-3381


mcgrawia (commented):
@notthattal can you explain bit about how this works? How do users set up the default model provider? From the ticket, it looks like the purpose of this is to eventually enable evals to run without and config required? If that's the goal, how do users configure the default provider? It seems like we need something in the UI? Or does this work OOTB if we just add the `DEFAULT` to the supported provider enum?

PR #879, opened 2025-11-26
move datasets under tasks

notthattal:
## Description
- Adds a task_id column explicitly to datasets that is now a foreign key to the tasks table. This was already included in the metadata of a dataset but now we get the added benefits of making it a foreign key.
- The migration first will assign a task id to a dataset if it exists in its metadata. If not it will check datasets related to expirements or notebooks which are already associated to tasks to assign the task id. As a fallback assign the dataset to the newest created task
- Modifies the route to create a dataset and to search datasets to require a task id
- Updates tests to use the new routes

## Jira Ticket
- https://arthurai.atlassian.net/browse/UP-3442

## Next Steps
- Make transforms global
- Create online evals

### DIFF
--- a/genai-engine/tests/unit/routes/datasets/test_datasets_crud.py
+++ b/genai-engine/tests/unit/routes/datasets/test_datasets_crud.py
@@ -24,6 +24,7 @@ def test_user_story_datasets_crud(client: GenaiEngineTestClientBase) -> None:
     )
     assert status_code == 200
     assert created_dataset.name == dataset_name
+    assert created_dataset.task_id == agentic_task.id
     assert created_dataset.description == dataset_description
     assert created_dataset.metadata == dataset_metadata
     assert created_dataset.id is not None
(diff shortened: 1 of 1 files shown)


genai-engine/alembic/versions/2025_11_26_1127-6273338ffe42_add_task_id_to_datasets.py
```diff
@@ -0,0 +1,85 @@
...
+# revision identifiers, used by Alembic.
+revision = "6273338ffe42"
+down_revision = "ee4b2e171875"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Step 1: Add task_id column as nullable first
+    op.add_column("datasets", sa.Column("task_id", sa.String(), nullable=True))
```
mcgrawia: should this be a UUID column?

PR #889, opened 2025-11-28
UP-3345: Cleanup when application is teardown

Kacpersoftberry:
When application was closing we were getting warning that tempfiles were not properly cleanup. I've added teardown mechanism in the lifespan to delete/close properly everything.


genai-engine/src/server.py
```diff
@@ -219,6 +219,83 @@ async def lifespan(app: FastAPI):
...
+                torch_instantiator._TEMP_DIR = None
+                logger.debug("PyTorch temporary directory cleaned up successfully")
+        except Exception as e:
+            logger.debug(f"Could not cleanup PyTorch temp directory: {e}")
+
+        for i in range(3):
+            collected = gc.collect()
+            logger.debug(f"Garbage collection pass {i+1}: freed {collected} objects")
+
+        time.sleep(0.2)
```
alexshanabrook: why was this needed?

Kacpersoftberry: This wait is here, cause emptying cache in PyTorch is done async, and to close all weak references.

genai-engine/src/server.py
```diff
@@ -219,6 +219,83 @@ async def lifespan(app: FastAPI):
 
     yield
 
+    logger.debug("Cleaning up models...")
+    try:
+        import gc
```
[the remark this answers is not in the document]
Kacpersoftberry: moved to top of the file

genai-engine/src/server.py
```diff
@@ -219,6 +220,46 @@ async def lifespan(app: FastAPI):
 
     yield
 
+    # We need to explicitly clean up models to prevent ResourceWarning when the server is stopped
+    logger.debug("Cleaning up models...")
+    try:
+        from utils import model_load
```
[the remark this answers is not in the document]
alexshanabrook: @Kacpersoftberry

genai-engine/src/server.py
alexshanabrook: @Kacpersoftberry

genai-engine/src/server.py
```diff
@@ -231,6 +223,43 @@ async def lifespan(app: FastAPI):
 
     yield
 
+    # We need to explicitly clean up models to prevent ResourceWarning when the server is stopped
+    logger.debug("Cleaning up models...")
```
alexshanabrook: @Kacpersoftberry curious, did you check that this clean up is still needed after the pytorch upgrade? Based on the warning message in the ticket, I'm not really clear on what this clean up would apply to. My best guess is this fix was for the second warning message:

```
/usr/lib/python3.12/tempfile.py:940: ResourceWarning: Implicitly cleaning up <TemporaryDirectory '/tmp/tmp8iigijpt'>
  _warnings.warn(warn_message, ResourceWarning)
/home/vagrant/.cache/pypoetry/virtualenvs/arthur-genai-engine-_YwsPmY_-py3.12/lib/python3.12/site-packages/litellm/llms/custom_httpx/async_client_cleanup.py:66: DeprecationWarning: There is no current event loop
  loop = asyncio.get_event_loop()
  ```
  
  but it seems to me like:
  1) the temporary directory warning should be suppressed by the pytorch upgrade
  2) the event loop warning wouldn't have anything to do with this clean up
  
  Additionally, all of the warnings included in the ticket were at application boot time. This logic occurs when the server is stopped. I'm curious what problem it's trying to solve? Could you include the issue you were seeing somewhere so that we know why we've added this?

PR #894, opened 2025-12-01
make transforms global

notthattal:
## Description
- Makes what was dataset transforms now global "trace transforms"
- There is now a trace transforms table which holds all transform information and the dataset transforms table now only holds a unique id, the dataset id, the transform id associated with that dataset and when it was added
- Crud operations for transforms within the dataset router now only do the following:
        - Get/list transforms associated with a specific dataset (by dataset id)
        - Add a transform to a dataset
        - Remove a transform from a dataset
 - Crud operations for global transforms now handle
        - Getting a transform by task id and transform id
        - list all transforms for a specific task (with filtering and pagination)
        - Create a new transform with variables (what was columns before)
        - Update a transform with a new definition/name/description
        - Delete a transform
- The execute transforms function was now split into two functions, one which handles the basic execution and one specific for datasets which formats it the same way it was expected to be formatted before these changes

## Jira Ticket
- https://arthurai.atlassian.net/browse/UP-3443

## Next Steps
- Use the new global transforms to make online evals work

### DIFF
--- a/genai-engine/src/utils/trace.py
+++ b/genai-engine/src/utils/trace.py
@@ -318,15 +318,13 @@ def get_nested_value(
         elif isinstance(current, list):
             # Try to parse key as integer index
             if not key.isdigit():
-                raise ValueError(f"Key '{key}' is not a valid integer index")
+                return default
 
             index = int(key)
             if 0 <= index < len(current):
                 current = current[index]
             else:
-                raise ValueError(
-                    f"Index {index} not in range for list of length {len(current)}",
-                )
+                return default
         else:
             return default
 
--- a/genai-engine/tests/unit/routes/transforms/test_transform_execution.py
+++ b/genai-engine/tests/unit/routes/transforms/test_transform_execution.py
@@ -699,7 +699,7 @@ def test_execute_transform_array_index_extraction(
         assert variables["output_id_1"] == "2"
         assert variables["output_name_1"] == "Jane"
 
-        # Test the failing case where the index is out of bounds
+        # Test the failing case where the index is out of bounds returns the default value
         transform_definition = {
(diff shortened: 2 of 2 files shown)


genai-engine/src/db_models/transform_models.py
```diff
@@ -0,0 +1,41 @@
...
+    from db_models.task_models import DatabaseTask
+
+
+class DatabaseTraceTransform(Base):
+    __tablename__ = "trace_transforms"
+    id: Mapped[uuid.UUID] = mapped_column(
+        UUID,
+        primary_key=True,
+        default=uuid.uuid4,
+        index=True,
```
mcgrawia: I think if you have `primary_key=True` you don't need to set `index=True`?

genai-engine/src/db_models/task_models.py
```diff
@@ -38,6 +39,10 @@ class DatabaseTask(Base, IsArchivable):
         back_populates="task",
         lazy="select",
     )
+    trace_transforms: Mapped[list["DatabaseTraceTransform"]] = relationship(
+        back_populates="task",
+        lazy="select",
+    )
```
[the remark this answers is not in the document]
notthattal: removed in [this commit](https://github.com/arthur-ai/arthur-engine/pull/894/commits/3c1eba754460608836d7fc64eae5549c926451d9)

genai-engine/src/db_models/transform_models.py
notthattal: removed in [this commit](https://github.com/arthur-ai/arthur-engine/pull/894/commits/3c1eba754460608836d7fc64eae5549c926451d9)

PR #898, opened 2025-12-03
Accept traces w/o task ids + API for identifying these

fryz:
This PR implements support for accepting traces without a task_id and adds a new API endpoint to group and return root-spans for unregistered traces. 

This enables users to send traces before registering applications and helps governance users identify unknown use cases.

The API endpoint works by: 
* Identifying traces without task_ids 
  * Identified by root spans (spans w/o parent ids) that don't have a task id
* Grouping the spans by the "name" attribute 
* Returning the group name + number of traces in that group


genai-engine/alembic/versions/2025_12_03_1444-make_trace_metadata_task_id_nullable.py
```diff
@@ -0,0 +1,46 @@
...
+Create Date: 2025-12-03 14:44:00.000000
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "a1b2c3d4e5f67890"
+down_revision = "8a2f6f0802a3"
```
fryz: same here

genai-engine/src/services/trace/span_query_service.py
```diff
@@ -1103,3 +1103,48 @@ def _build_token_cost_aggregations(
...
+        # Base query for grouping
+        base_query = (
+            select(
+                DatabaseSpan.span_name,
+                func.count().label("count"),
+            )
+            .where(DatabaseSpan.parent_span_id.is_(None))
+            .where(DatabaseSpan.task_id.is_(None))
+            .group_by(DatabaseSpan.span_name)
+        )
```
mcgrawia: from the looks of this, I think this is going to do a full span table scan. Should we introduce some time bounds, like on the GET /traces/spans endpoints, so this query bounds the results within the last day/week/month of time? Upsolve has 40k spans and increasing, this query might get pretty slow if done over the full span table as it grows unbounded into the future

PR #968, opened 2025-12-15
Fix filtering on annotations for trace metadata

notthattal:
## Description
- Backend: Changes the way we filter trace metadata to retrieve all annotations for the traces that have been filtered out based on all filters (including annotation filters) and removes the function which re-filtered them after-the-fact 
- Frontend: Fixes an issue where filtering on continuous eval name wouldn't work because the request to the backend was formatted incorrectly

## Jira Ticket
- https://arthurai.atlassian.net/browse/UP-3528

### DIFF
--- a/genai-engine/src/schemas/internal_schemas.py
+++ b/genai-engine/src/schemas/internal_schemas.py
@@ -744,13 +744,10 @@ class TraceMetadata(TokenCountCostSchema):
 
     @staticmethod
     def _from_database_model(x: DatabaseTraceMetadata):
-        # Add annotations if they're loaded
-        annotations = None
-        if hasattr(x, "annotations") and x.annotations:
-            annotations = [
-                AgenticAnnotation.from_db_model(annotation)
-                for annotation in x.annotations
-            ]
+        # Add formatted annotations
+        annotations = [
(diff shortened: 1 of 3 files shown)


genai-engine/src/schemas/internal_schemas.py
```diff
@@ -744,6 +744,14 @@ class TraceMetadata(TokenCountCostSchema):
 
     @staticmethod
     def _from_database_model(x: DatabaseTraceMetadata):
+        # Add annotations if they're loaded
```
mcgrawia: is there ever a time we don't return the annotations on the metadata? I would expect this to always be returned

mcgrawia (dismissed):

genai-engine/src/schemas/internal_schemas.py
mcgrawia: also, is it true that if they're not loaded this attribute won't exist? I don't know if sqlalchemy will still create the attribute, but it will be null or something when not loaded

notthattal: I think it still exists that was just a safety check but I can update it to simplify. and yeah I can update it to always return annotations

mcgrawia (commented):
lgtm!

PR #975, opened 2025-12-16
Traces onboarding wizard empty state

hashnadz:
This MR replaces the empty state for traces view with a guided wizard to help new users get connected with each new task they create. Once set up, they will never see this view again unless they create a new task.


genai-engine/ui/src/components/TracesView.tsx
```diff
@@ -17,6 +18,58 @@ import { FilterStoreProvider } from "./traces/stores/filter.store";
...
+      if (stored) {
+        try {
+          const parsed = JSON.parse(stored);
+          // Only show tabs if explicitly dismissed
+          setWelcomeDismissed(parsed.dismissed === true);
+        } catch {
+          // ignore
+        }
+      }
+    };
```
xV3rt: Storing `welcomeDismissed` in a state while having it potentially changed somewhere else in the app creates many issues here. For such usecase we want to store and handle that value in a Zustand store, which will make it trivial to access it throughout the app and potentially update from different places. We can also persist it in local storage if needed using Zustand persist

genai-engine/ui/src/components/traces/components/tables/SpanLevel.tsx
```diff
@@ -122,12 +132,20 @@ export const SpanLevel = () => {
...
           </Typography>
         </TracesEmptyState>
+      ) : welcomeDismissed ? (
+        <TracesEmptyState title="No spans yet">
+          <Typography variant="body1" color="text.secondary">
+            Spans will appear here once your application starts sending data
+          </Typography>
+        </TracesEmptyState>
+      ) : (
+        <TracesWelcomePage />
```
xV3rt: This part of UI is reused multiple times in other files. We should probably extract it as a separate component with configurable props and import it in other places to avoid code duplication and make the future maintenance easier.

PR #1062, opened 2026-01-07
add ability to disable continuous evals

notthattal:
## Description
- Add the ability to enable/disable CEs so they don't run over incoming traces without having to delete them
- Add the ability to change enabled parameter in the creation and update routes for CEs
- Add the ability to filter on if a CE is enabled in both listing CEs and also listing CE run results
- Prevent rerunning evaluations if the CE has been disabled
- Add unit tests to validate CRUD and rerun functionality as well as one to ensure only enabled CEs are enqueued 

## Jira Ticket
- https://arthurai.atlassian.net/browse/UP-3534


alexshanabrook:
qq—by default do list endpoints exclude continuous evals that have been disabled? if so lgtm!

PR #1079, opened 2026-01-12
feat: add agentic notebooks

TheLukaszNs-Blazity:
https://github.com/user-attachments/assets/3059a361-62ea-4aaa-9978-b829557234bf


genai-engine/ui/src/components/agent-experiments/new/context/validation-context.tsx
```diff
@@ -0,0 +1,3 @@
+type ValidatorPair = readonly [];
+
+const VALIDATORS = {};
```
xV3rt: a leftover file?

PR #1111, opened 2026-01-16
Add synthetic data generation workflow in datasets

fryz:


notthattal:
i could only seem to get this to happen once so I'm not sure if it's an actual bug or just an issue with gpt for that one generation but one run when I had it generate content in the UI it cutoff all rows and added ellipses like "To access the analytics dashboard of the Arthur Platform, log in to your account and navigate to th..." not just in the view but when actually adding the generated data to the dataset

PR #1113, opened 2026-01-16
Agent request body update

alexshanabrook:
In response to slack thread: https://arthur-ai.slack.com/archives/C095P231EEB/p1768404947113699

Also updated session-id to only be sent to agent through headers, not the headers or request body—missed that in the initial PR

how this looks in the UI:
Screenshot 2026-01-16 at 2 04 44 PM

Screenshot 2026-01-16 at 2 08 43 PM


ian-moonnox (approved):
BE LGTM as well. thanks @alexshanabrook for making this adjustment 🙏

PR #1146, opened 2026-01-26
Poc/agent discovery

videetparekh:


genai-engine/src/routers/v1/trace_api_routes.py
```diff
@@ -96,6 +97,76 @@ def receive_traces(
         db_session.close()
 
 
+@trace_api_routes.post(
+    "/traces/gcp",
```
videetparekh: Would this endpoint ever be used anymore? Since we're running a polling job, we don't really need this endpoint right?

PR #1188, opened 2026-02-02
Fix transforms list limited to 10 items

madeleinelane:
## Summary
- Fixed transforms table only showing 10 transforms instead of all available transforms
- Fixed transforms dropdown in "add traces to dataset" flow also limited to 10 items
- Root cause: The API defaults to `page_size=10` and the frontend hooks weren't specifying a larger value

## Changes
Updated three `useTransforms` hooks to request `page_size: 1000`:
- `genai-engine/ui/src/components/transforms/hooks/useTransforms.ts`
- `genai-engine/ui/src/components/traces/components/add-to-dataset/hooks/useTransforms.ts`
- `genai-engine/ui/src/hooks/transforms/useTransforms.tsx`

## Test plan
- [x] Verified transforms table shows all transforms (tested with 15+ transforms)
- [x] Verified transforms dropdown in add-to-dataset shows all transforms
- [x] TypeScript type-check passes
- [x] ESLint passes


xV3rt (approved):

TheLukaszNs-Blazity:
I'm worried about UI performance drop after this PR. We will probably need to find another way to deal with this later

PR #1202, opened 2026-02-04
Add pagination metadata to transforms list endpoint

madeleinelane:
## Summary
- Add `total_count`, `page`, `page_size`, and `total_pages` to the transforms list endpoint response
- Enables proper pagination UI without over-fetching data
- Addresses performance concerns raised about fetching 1000 items when there might only be a few

## Changes
- Update `TraceTransformRepository.list_transforms()` to return total count before pagination
- Create new `TraceTransformListResponse` schema with pagination metadata fields
- Update route handler to calculate `total_pages` and return full pagination info
- Update tests to verify pagination metadata is correctly returned
- Regenerate OpenAPI spec and TypeScript client

## Test plan
- [x] Unit tests pass locally (11 tests in `test_transform_crud.py`)
- [x] Verified API returns pagination metadata via curl
- [x] Frontend compiles with new TypeScript types
- [ ] Manual testing in UI


notthattal:
I think this PR is kind of overkill based on the ticket. I think they just wanted to add the count to the ListTraceTransformsResponse class not all the pagination information. That class exists in arthur-common so we would need a partner MR in that repo. We shouldn't add this new class that's in this PR.

fryz (approved):
Approved. Clean addition of pagination metadata to the transforms list endpoint. The pattern of counting before applying pagination is correct, and the response schema follows the established conventions in the codebase. All CI passes. LGTM.

PR #1203, opened 2026-02-04
Add agent metadata

videetparekh:
## GenAI Engine (Data Plane)
- Add task metadata storage to support agent registration metadata. 
- Added `task_metadata` JSONB column to `DatabaseTask` model for flexible metadata storage. 
- Updated Task internal schema to convert between `AgentMetadata` request/response models and database dict storage. 
- Enables storing GCP agent source information in task records.

## ML Engine (Job Executor)
- Pass agent metadata through task creation job flow. 
- Updated `CreateModelTaskJobSpec` to include `agent_metadata` field. 
- Modified `ShieldConnector.create_task()` to accept and serialize agent metadata for the generated client. 
- Updated `TaskCreator` to forward metadata from job spec to connector. 
- Enables metadata flow from control plane to data plane.


fryz (dismissed):
Need to fix the lint + tests, but overall looks good. 

Same nit as in other PR - might consider renaming and/or otherwise commenting that "metadata" here refers to the provenance of the agent (eg: where does it come from)

PR #1204, opened 2026-02-04
New UIX and metric KPIS for the All Tasks tile cards

hashnadz:
- New metrics (traces, tokens and success)
- New tool tips and user context messaging on hover 
- New last activity data point 
- New "View traces" text displays on hover 
- In card success message when Task ID is copied 
- Task names will stay in first line and elipsis if they exceed available space
- When success is above 1% but below 50%, section of tile will turn red and display warning message on hover 

Screenshot 2026-02-04 at 6 42 48 PM

### DIFF
--- a/genai-engine/ui/src/components/TaskCard.tsx
+++ b/genai-engine/ui/src/components/TaskCard.tsx
@@ -217,7 +217,17 @@ export const TaskCard: React.FC<TaskCardProps> = ({ task }) => {
           </Stack>
 
           {/* Footer */}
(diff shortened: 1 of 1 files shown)


TheLukaszNs-Blazity (approved):

TheLukaszNs-Blazity:
@hashnadz for your question, I think these are related to change in the position or ids of some elements that meticulous expected. For the traces, I think these disappeared because we are only showing past month time frame by default and they are simply older

PR #1254, opened 2026-02-12
Up 3673 currency select

martin-arthur-ai:
Uses backend service to convert currency if the config is set at application startup. Changes currency symbol if the backend is not using USD.

formatCurrency is using a local version again. This is because the shared-components version does not have a localization value. To be updated in a later branch/[ticket](https://arthurai.atlassian.net/browse/UP-4032).

For currencies that are not in the Frankfurter API, or environments that are air-gapped, ENV variables can be used to set the static currency and exchange rate:
```
# Use static provider so you control the rate
CURRENCY_PROVIDER=static

# The currency code (e.g. Saudi Riyal)
CURRENCY_DEFAULT_CURRENCY=SAR

# USD → SAR rate (e.g. 3.75; adjust to your desired rate)
CURRENCY_EXCHANGE_RATE=3.75
```

**USD**
Screenshot 2026-02-26 at 13 34 07

**AUD**
Screenshot 2026-02-26 at 13 41 23


genai-engine/src/routers/v2/system_management_routes.py
```diff
@@ -65,6 +69,25 @@ def get_token_usage(
...
+@public_endpoint
+def get_display_settings(
+    db_session: Session = Depends(get_db_session),
+    application_config: ApplicationConfiguration = Depends(get_application_config),
+) -> DisplaySettingsResponse:
+    try:
+        return DisplaySettingsResponse(
+            default_currency=get_display_currency(application_config),
+        )
+    finally:
```
notthattal: this finally is unnecessary we auto-close sessions in that get_db_session function

PR #1255, opened 2026-02-12
fixed navbar

martin-arthur-ai:
[Original comment by Ashley](https://arthur-ai.slack.com/archives/C095P231EEB/p1770755006366429)

The navigation bar was being duplicated 4 times across the loading/error/standard/notFound states. When a user would click an item, the app would swap components. To the user it looked like the navbar was scrolling back up because state was reset.

I fixed it by using only one instance of the navbar and using an OUTLET to display components.


genai-engine/ui/src/components/SidebarNavigation.tsx
```diff
@@ -36,60 +36,60 @@ interface NavigationItem {
...
-      id: "settings",
-      label: "Settings",
-      items: [
-        { id: "task-details", label: "Task Details", icon: <SettingsOutlined /> },
-        { id: "model-providers", label: "Model Providers", icon: <AppsOutlined /> },
-        { id: "api-keys", label: "API Keys", icon: <KeyOutlined /> },
-      ],
-    },
-  ];
+const navigationSections: NavigationSection[] = [
```
martin-arthur-ai: Moved out of SidebarNavigation because we don't need to redefine it on every re-render.

PR #1295, opened 2026-02-19
Global Loop for Task Polling

videetparekh:


genai-engine/src/routers/v1/agent_polling_routes.py
```diff
@@ -19,33 +17,34 @@
 
 
 @agent_polling_routes.post(
-    "/tasks/{task_id}/agent-polling/retry/{agent_polling_data_id}",
-    description="Retry a failed agent polling job for a given agent polling data id.",
+    "/tasks/{task_id}/agent-polling/execute",
+    description="Manually trigger a polling job for a task. "
+    "Does not require any particular state — admins can use this "
+    "to force an immediate poll outside the normal loop cadence.",
```
mcgrawia: Hmm I had a different requirement for this endpoint.. I thought the point of it was to manually trigger the full scan job, not just for already discovered tasks? For example, if a user wants to initiate a new discovery run without having to wait for the next run, how would they do that?

The main reason is I'm thinking we may eventually need to trigger the full scan job whenever the ML-engine discover agents jobs run. 

I can see a world where admins will bring a new agent online, then immediately go to the Arthur platform and click "Scan for new agents" and expect it to show up. That will almost certainly be the first thing they do when testing out this feature for the first time

mcgrawia: yeah probably true, especially since we rely on manually fetching traces for "registered" agents as well, which will result in delayed traces showing up in the UI + metrics since they aren't sent to the engine in real time

ntatsumi:
@claude run a full code review on this pull request again. Don't stop yourself from running the review even though you'll see that you already ran a review and left comments.

Below link has the instructions on how to run the code review. Make sure you follow it.
https://github.com/anthropics/claude-code/blob/main/plugins/code-review/commands/code-review.md

notthattal (commented):
I think calling the LLMModel class "Model" in the arthur-common partner MRs to this one may be the source of failures in scope. it's been pretty confusing to debug:

https://gitlab.com/ArthurAI/arthur-scope/-/jobs/13229399914

PR #1335, opened 2026-02-24
Fix GCP span kind not stored in span_kind column

videetparekh:
## Summary
- After `convert_gcp_labels_to_openinference` runs, it explodes all dotted attribute keys into nested dicts (e.g. `"openinference.span.kind"` → `{"openinference": {"span": {"kind": "..."}}}`)

- The subsequent flat `attributes.get("openinference.span.kind", "LLM")` and `attributes.get("session.id")` calls always miss, falling back to `"LLM"` and `None` respectively

- Fixed by replacing both with `trace_utils.get_nested_value(...)` which navigates the nested structure via dot-path traversal (same utility used elsewhere in the ingestion pipeline)

### DIFF
--- a/deployment/ecs/model-upload/poetry.lock
+++ b/deployment/ecs/model-upload/poetry.lock
@@ -62,18 +62,18 @@ uvloop = ["uvloop (>=0.15.2)"]
 
 [[package]]
 name = "boto3"
-version = "1.42.54"
+version = "1.42.55"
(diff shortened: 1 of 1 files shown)


mcgrawia (dismissed):
lgtm but should we have some sort of test that covers this to make sure it doesn't break in the future?

mcgrawia (approved):

PR #1340, opened 2026-02-24
UP-3149-trace-retention

martin-arthur-ai:
[UP-3149](https://arthurai.atlassian.net/browse/UP-3149?atlOrigin=eyJpIjoiZjFjZDgzMTkxYTk1NDQyNzg3ZmIyZGE1ZTM2NWEyYWIiLCJwIjoiaiJ9)

Screenshot 2026-04-06 at 09 48 44
Screenshot 2026-04-07 at 13 29 53


genai-engine/src/repositories/trace_retention_repository.py
```diff
@@ -0,0 +1,128 @@
...
+    DatabaseAgenticAnnotation,
+    DatabaseMetricResult,
+    DatabaseResourceMetadata,
+    DatabaseSpan,
+    DatabaseTraceMetadata,
+)
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_BATCH_SIZE = 500
```
fryz: would call this something more specific - `DEFAULT_RETENTION_BATCH_SIZE`, etc.

genai-engine/src/repositories/trace_retention_repository.py
```diff
@@ -0,0 +1,128 @@
...
+
+def get_expired_trace_ids(
+    db_session: Session,
+    cutoff: datetime,
+    batch_size: int = DEFAULT_BATCH_SIZE,
+) -> list[str]:
+    """Return up to batch_size trace_ids from trace_metadata with end_time < cutoff."""
+    stmt = (
+        select(DatabaseTraceMetadata.trace_id)
+        .where(DatabaseTraceMetadata.end_time < cutoff)
```
fryz, line 37: just want to check some of the edge-cases here 

what happens when: 
* database trace doesn't have an end time 
* span(s) have an end time that exceeds the trace end time (is this possible?) 
* end_time is null/none/etc. 
* end_time is greater than time.now()

genai-engine/src/repositories/trace_retention_repository.py
```diff
@@ -0,0 +1,128 @@
...
+DEFAULT_BATCH_SIZE = 500
+
+
+def get_expired_trace_ids(
+    db_session: Session,
+    cutoff: datetime,
+    batch_size: int = DEFAULT_BATCH_SIZE,
+) -> list[str]:
+    """Return up to batch_size trace_ids from trace_metadata with end_time < cutoff."""
+    stmt = (
```
fryz, line 35: we probably want this ordered in some time-sorted order

genai-engine/src/repositories/trace_retention_repository.py
```diff
@@ -0,0 +1,128 @@
...
+        )
+        db_session.execute(
+            delete(DatabaseResourceMetadata).where(
+                DatabaseResourceMetadata.id.in_(batch_resource_ids),
+                ~DatabaseResourceMetadata.id.in_(remaining_span_resource_ids),
+                ~DatabaseResourceMetadata.id.in_(remaining_trace_resource_ids),
+            )
+        )
+
+    db_session.commit()
```
fryz: i think the way this is written will likely cause some issues when the engine is deployed in multi-node setup 

we'll have multiple nodes racing to the delete the same spans, and because its touching multiple resources in a single commit, the first committer is going to win and other nodes will have problems.

genai-engine/src/services/trace_retention_service.py
```diff
@@ -0,0 +1,104 @@
...
+                delete_trace_batch(db_session, trace_ids)
+                db_session.commit()
+                total_deleted += len(trace_ids)
+
+            if total_deleted > 0:
+                logger.info(
+                    "Trace retention: deleted %d traces older than %s",
+                    total_deleted,
+                    cutoff.isoformat(),
+                )
```
fryz: would consider adding: 

1. A circuit-breaker here - something that trips if there are issues and requires manual intervention to restore (eg: restart the service)
2. Some delay/lag between deleting batches so we don't put a huge amount of load on the DB on startup

genai-engine/src/repositories/trace_retention_repository.py
martin-arthur-ai: Changed to default trace retention batch size

genai-engine/src/repositories/trace_retention_repository.py
martin-arthur-ai: I got a long-winded summary from opus, but here is the summary: All four edge cases are safe. end_time is NOT NULL at the database level, trace_metadata.end_time is always >= max(span.end_time) by the ingestion upsert logic, and future-dated traces are naturally excluded from the < cutoff filter.

genai-engine/src/repositories/trace_retention_repository.py
martin-arthur-ai: Added SELECT ... FOR UPDATE SKIP LOCKED to get_expired_trace_ids. On PostgreSQL, each node now row-locks the trace_metadata rows it selects; concurrent nodes skip those locked rows and pick different batches. The locks are held for the full batch lifecycle (until commit() after delete_trace_batch completes), so there's no interleaving window.

Also added a standalone index on trace_metadata.end_time — the retention query filters and orders by end_time across all tasks, and the existing composite indexes (all task_id-leading) wouldn't help. Without it the FOR UPDATE would escalate a sequential scan into a row-locking sequential scan.

mcgrawia:
@martin-arthur-ai minor UX thing, rather than adding a new menu section for Application Settings, can we just rename "User Settings" to "Settings", and add this there as another section? I don't think creating a separate section is worth it when the "User Settings" modal already has similar settings

image

genai-engine/src/repositories/trace_retention_repository.py
```diff
@@ -0,0 +1,129 @@
...
+
+    # 4. Trace metadata
+    db_session.execute(
+        delete(DatabaseTraceMetadata).where(
+            DatabaseTraceMetadata.trace_id.in_(trace_ids)
+        )
+    )
+
+    # 5. Orphan resource_metadata: only those referenced by this batch and no longer referenced elsewhere
+    if batch_resource_ids:
```
mcgrawia: @videetparekh question for you.. do we need to delete the resources for these traces? Or should we leave them? I'm not sure what would happen if a trace arrives again for the same resource, and it creates a new resource because the old one was deleted? Would that cause the trace to map to a new task?

genai-engine/src/routers/v2/system_management_routes.py
```diff
@@ -88,7 +88,6 @@ def get_display_settings(
     "/configuration",
     description="Get application configuration settings.",
     response_model=ApplicationConfigurationResponse,
-    include_in_schema=False,
 )
```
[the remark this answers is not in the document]
mcgrawia: @martin-arthur-ai this seems like a valid flag

genai-engine/src/services/trace_retention_service.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+class TraceRetentionJob(BaseQueueJob):
+    """Job representing one run of the trace retention cleanup."""
+
+    def __init__(self, delay_seconds: int = 0) -> None:
+        super().__init__(delay_seconds=delay_seconds)
+
+
+class TraceRetentionService(BaseQueueService[TraceRetentionJob]):
```
mcgrawia, line 73: I'm not very familiar with the BaseQueueService. Does it do some sort of leader election/detection? How does this job work if there's multiple replicas?

genai-engine/src/services/trace_retention_service.py
```diff
@@ -0,0 +1,104 @@
...
+        """Run retention once, then wait 24 hours and enqueue again."""
+        logger.info("Background thread started for %s", self.service_name)
+        if not self.shutdown_event.is_set():
+            self.enqueue(TraceRetentionJob(delay_seconds=0))
+
+        while not self.shutdown_event.is_set():
+            if self.shutdown_event.wait(timeout=ONE_DAY_SECONDS):
+                break
+            if not self.shutdown_event.is_set():
+                self.enqueue(TraceRetentionJob(delay_seconds=0))
```
mcgrawia: I'm don't quite follow this function.. Does it essentially run the retention clean up when the server boots, when it waits on a signal for 24 hours and queues it again? Why not just set the `delay_seconds` to 1 day and let the queue take care of it?

mcgrawia (dismissed):

genai-engine/src/routers/v2/system_management_routes.py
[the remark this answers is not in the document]
martin-arthur-ai: the endpoints were hidden because they weren't yet exposed to end-users through a polished UI. Now that trace retention configuration is surfaced in the Settings modal, these endpoints are part of the public API surface and should be in the schema.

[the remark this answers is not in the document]
mcgrawia: @martin-arthur-ai that's fine, but the second part of the issue above says that the FE isn't using the API client for this endpoint now that it's included in the schema. If we include it, we should use the generated client

genai-engine/src/repositories/trace_retention_repository.py
martin-arthur-ai: The  `.order_by(DatabaseTraceMetadata.end_time.asc())` is below the next comment

genai-engine/src/services/trace_retention_service.py
martin-arthur-ai: **Circuit breaker:** TraceRetentionService tracks consecutive failures and trips a fail-stop latch after 3 consecutive errors (CIRCUIT_BREAKER_THRESHOLD). Once tripped, the background loop exits permanently with a CRITICAL log message stating manual intervention (process restart) is required to resume. A runaway deletion cap (MAX_TRACES_PER_RUN = 100,000) also trips the latch as a safety net.

**Inter-batch delay:** There's a 1-second pause (INTER_BATCH_DELAY_SECONDS) between each batch of 500 deletions, implemented via shutdown_event.wait() so it's interruptible on graceful shutdown. This throttles DB load during cleanup.

genai-engine/src/routers/v2/system_management_routes.py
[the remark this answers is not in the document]
martin-arthur-ai: It's now using the generated schema

genai-engine/src/services/trace_retention_service.py
martin-arthur-ai: I changed the function to match existing patterns. BaseQueueService already exists in a number of services

martin-arthur-ai:
@claude review

genai-engine/src/repositories/trace_retention_repository.py
videetparekh: Resources are logical groupings of traces, they shouldn't be deleted.

genai-engine/src/services/trace_retention_service.py
mcgrawia: I'm not quite following your answer and what you mean by "match existing patterns". I agree the BaseQueueService is already used elsewhere. What I'm asking though is how does your implementation handle the leader election issue if there are multiple replicas of the BE running? Or is the plan that this job will run on every replica?

martin-arthur-ai: The implementation uses pg_try_advisory_lock for leader election (the _background_loop method). Only the replica that acquires the lock runs retention; others stand by and retry each interval. This follows the same pattern as GlobalAgentPollingService. If the leader goes down, the lock auto-releases when its DB session closes.

genai-engine/src/repositories/trace_retention_repository.py
martin-arthur-ai: Removed the orphan resource deletion block and associated resource ID collection from delete_trace_batch in trace_retention_repository.py

genai-engine/src/services/trace_retention_service.py
mcgrawia: ok thanks for confirming

genai-engine/src/schemas/response_schemas.py
```diff
@@ -45,6 +45,8 @@ class ApplicationConfigurationResponse(BaseModel):
         None
     )
     max_llm_rules_per_task_count: int
+    trace_retention_days: int
+    allowed_trace_retention_days: tuple[int, ...] = ()
```
mcgrawia, line 50: @martin-arthur-ai what is the `...` here for?

mcgrawia (dismissed):

fryz (approved):

PR #1354, opened 2026-02-26
UP-3917 - Add ability to unarchive tasks

videetparekh:
https://arthurai.atlassian.net/browse/UP-3917
1. Adding include_archived to task search request
2. Adding is_archived to the task response
3. Adding an unarchive endpoint (unarchives rule and metrics)
4. Add logs to raise a warning if data is being sent to an archived task

### DIFF
--- a/genai-engine/ui/src/components/AllTasks.tsx
+++ b/genai-engine/ui/src/components/AllTasks.tsx
@@ -365,11 +365,11 @@ export const AllTasks: React.FC = () => {
                       </Stack>
                     </Box>
 
-                    <div className="grid gap-4 sm:grid-cols-2 lg:grid-cols-3">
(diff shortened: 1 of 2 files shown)


genai-engine/src/repositories/tasks_repository.py
```diff
@@ -307,6 +310,31 @@ def archive_task(self, task_id: str) -> None:
...
+
+        for link in db_task.rule_links:
+            if link.rule.scope == RuleScope.TASK:
+                self.rule_repository.unarchive_rule(link.rule_id)
+
+        for metric_link in db_task.metric_links:
+            self.metric_repository.unarchive_metric(metric_link.metric_id)
+
+        db_task.archived = False
+        self.db_session.commit()
```
mcgrawia: curious, does this work ok calling commit multiple times? I think the unarchive_rule, unarchive_metric, and unarchive_task all end with a db_session.commit(). 

Ideally, there should only be one commit per API request so the operation either succeeds or fails atomically. Right now if there's a failure part way through, the state will be inconsistent

PR #1380, opened 2026-03-03
Up 3602 integration

martin-arthur-ai:
This PR integrates the components from the unify-frontend back into the genai-engine. To install locally, run `export GITLAB_UNIFY_FRONTEND_TOKEN=<token>` where the token is in the shared 1password

### DIFF
--- a/genai-engine/ui/src/components/traces/components/llm/ToolsTab.tsx
+++ b/genai-engine/ui/src/components/traces/components/llm/ToolsTab.tsx
@@ -17,8 +17,8 @@ type Props = {
 };
 
 export const ToolsTab = ({ span }: Props) => {
-  // TODO: Fix type error
-  const tools = getTools(span as unknown as Parameters<typeof getTools>[0]).map(getToolDefinition);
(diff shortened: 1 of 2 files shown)


genai-engine/ui/src/components/traces/components/CommonDrawer.tsx
```diff
@@ -52,47 +50,22 @@ export const CommonDrawer = () => {
...
-      </Drawer>
-    </>
+    <SharedCommonDrawer
+      open={open}
+      target={target}
+      id={id}
+      onClose={handleClose}
+      renderContent={({ target, id: contentId }: { target: DrawerTarget; id: string }) => {
+        const Content = CONTENT_MAP[target];
+        return <Content id={contentId} />;
```
xV3rt: Let's handle the case when `CONTENT_MAP[target]` is `undefined`

```
if (!Content) {
    console.error(`Unknown drawer target: ${target}`);
    return null;
}
```

xV3rt (dismissed):

xV3rt (approved):

PR #1402, opened 2026-03-07
fix: annotation explanation shows [object Object] instead of text (ae-lyd)

fryz:
## Summary
- Fixes copy/paste of annotation explanation showing `[object Object]` instead of text
- Properly converts annotation array to string before clipboard write
- Source issue: ae-lyd / Jira: UP-4000

🤖 Merged by Refinery (arthur_engine/refinery) via Gas Town merge queue


fryz:
Retargeting to dev branch.

PR #1414, opened 2026-03-08
Fix annotation explanation displaying [object Object] in traces

fryz:
## Summary
Fixed a display bug where annotation explanations in the traces interface showed "[object Object]" instead of readable text content.

## Problem
The traces interface was rendering annotation explanations as "[object Object]" instead of displaying the actual text content, making annotation data unreadable for users reviewing trace information.

## Solution
Properly serialized annotation description values and reverted the annotation description cell to use a plain div element with correct value handling in the AnnotationCell table component.

## User Impact
Users can now see readable annotation explanations in the traces interface instead of the cryptic "[object Object]" placeholder text. This restores the ability to understand annotation context when reviewing traces.

## Changed Areas
• **Annotation Cell Table** — Fixed value serialization and rendering for annotation descriptions
• **Trace Content Cell** — Updated annotation description display logic

## Type
Bug Fix

## Breaking Changes
None


TheLukaszNs-Blazity (dismissed):

jdbarthur (approved):
lgtm

PR #1416, opened 2026-03-08
fix: View Results navigates to specific result record, not eval list (ae-wvi)

fryz:
Fixes UP-4018: clicking "View Results" from the Annotations for trace section was navigating to the continuous evals list instead of the specific result record.

**Root cause:** `Link` component inside a `Menu.Portal` with a `Dialog` container — the Dialog's `stopPropagation` handler and Portal container were interfering with navigation, causing it to land on the management tab instead.

**Fix:** Replace `Link` with `useNavigate`/`onClick` for programmatic navigation that bypasses event propagation issues.

Same fix applied to "Rerun Annotation" menu item (same pattern).

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>


genai-engine/ui/src/components/traces/components/AnnotationCell/table.tsx
```diff
@@ -169,8 +182,7 @@ const createColumns = ({ taskId, container }: { taskId: string; container: React
                   render={
                     <ListItemButton
                       disabled={annotation.run_status !== "error"}
-                      component={Link}
-                      to={`/tasks/${taskId}/continuous-evals?id=${annotation.id}&tab=results&action=rerun`}
+                      onClick={() => onNavigate(`/tasks/${taskId}/continuous-evals?id=${annotation.id}&tab=results&action=rerun`)}
```
TheLukaszNs-Blazity: Same as above:
```suggestion
                      onClick={() => onNavigate(`/tasks/${taskId}/evaluate?id=${annotation.id}&section=ce-results&action=rerun`)}
```

PR #1427, opened 2026-03-09
fix(ui): sort and filter tasks by actual trace activity

hashnadz:
- Add useAllTasksActivity hook to bulk-fetch last 30 days of trace activity across all task IDs and return a taskId → lastActiveMs map
- Fix 'Recently updated' sort to use most recent trace timestamp instead of task.updated_at (which equals created_at on the backend)
- Fix 'Active in last X days' filter to use trace activity date instead of task creation date


fryz (dismissed):
Approved. Good fix - using actual trace activity timestamps for sort/filter rather than the task updated_at field gives more accurate recency.

One minor observation: the useAllTasksActivity hook caps at page_size=1000 traces across all tasks. With many tasks or high-volume deployments, recent activity for some tasks could be pushed out of this window. Worth noting as a follow-up if it becomes an issue in practice. For the current use case (sorting the task list) this is an acceptable trade-off.

Meticulous visual test failures appear to be baseline diffs from the changed sort order rather than regressions - expected for this change. All other CI passes. LGTM.

hashnadz:
@fryz Can you review/approve this again please. It got stopped by merge conflicts -- now addressed.

PR #1431, opened 2026-03-10
UP-4020: Do not allow transform deletion when dependant entities exist

xV3rt:


genai-engine/src/schemas/request_schemas.py
```diff
@@ -160,6 +160,32 @@ class TraceTransformUpdateRequest(BaseModel):
...
+        description="Agentic notebooks that reference this transform.",
+    )
+
+    @property
+    def has_dependents(self) -> bool:
+        return bool(
+            self.continuous_evals or self.agentic_experiments or self.agentic_notebooks
+        )
+
+
```
mcgrawia: I believe the `request_schemas.py` file is meant for HTTP request objects. IIUC this isn't http-user facing right? Would it make sense to keep this internal somewhere and not have it in with the public facing request objects?

PR #1447, opened 2026-03-12
UP-3948 Fix encoding special characters in entity names

xV3rt:


fryz (approved):
**Approved.** Clean fix for double-encoding path parameters containing special characters.

The approach is sound:
- Backend:  utility using  strips the inner encoding layer after Uvicorn's ASGI layer strips the outer one. Safe for non-double-encoded values.
- Frontend:  applies  twice — correct for values placed directly in API call path segments (not React Router URLs, which correctly use single ).
- FastAPI  pattern keeps the decoding logic DRY across all affected endpoints.

All CI checks pass. LGTM.

fryz (dismissed):
Approved. Clean fix for double-encoding path parameters with special characters. Backend decode_path_param uses unquote to strip inner encoding layer after Uvicorn strips the outer; frontend encodePathParam double-encodes for API calls while React Router links use single encodeURIComponent correctly. DRY pattern via FastAPI Depends. All CI passes. LGTM.

videetparekh (approved):

PR #1452, opened 2026-03-13
Add advanced filtering and sorting for traces and spans

xV3rt:
## Summary
This change adds comprehensive filtering and sorting capabilities to the trace viewer, allowing users to filter traces by span count and token count, and sort both traces and spans by multiple criteria. Users can now quickly find relevant traces in large datasets and organize them according to their analysis needs.

## Problem
The existing trace viewer lacked advanced filtering and sorting options, making it difficult for users to efficiently navigate and analyze large volumes of trace data. Users could not filter traces based on structural properties like span count or resource usage like token count, nor could they sort traces and spans by different attributes to prioritize their investigation workflow.

## Solution
Implemented a comprehensive filtering and sorting system that extends the existing trace query infrastructure. Added new filter types for span count and token count with range-based filtering, and built a flexible sorting system that works across both trace-level and span-level views. The UI was enhanced with updated filter modals and sorting controls integrated into the existing trace tables.

## User Impact
Users can now filter traces by the number of spans they contain and by token count ranges, enabling them to focus on traces that match specific complexity or cost criteria. They can sort both traces and spans by various attributes to organize their analysis workflow more effectively. This significantly improves the ability to navigate and analyze large trace datasets, reducing time spent searching for relevant traces and enabling more targeted debugging and performance analysis.

## Changed Areas
• **Trace Query Service** - Enhanced with new filtering logic for span count and token count, plus sorting capabilities
• **Legacy Span Routes** - Extended API endpoints with advanced filtering and sorting parameters
• **Span Repository** - Updated to support new query parameters and sorting options
• **Trace API Routes** - Modified to handle additional filtering and sorting requests
• **Trace Filtering UI** - Rebuilt filter modal with new span count and token count filter types
• **Trace Tables** - Added sorting controls and updated to use new filtering capabilities
• **API Client** - Generated new endpoints supporting advanced filtering and sorting parameters

## Type
Feature

## Breaking Changes
None


notthattal:
There's this other ticket that was assigned to me: https://arthurai.atlassian.net/browse/UP-3949 that is almost a duplicate. The only thing that's different is it adds a span count filter for traces too. Would it be possible to just add the span count filter to this PR if it's not too much work? Then we could close out all 3

xV3rt:
Will do, @notthattal !

notthattal (dismissed):
looks good to me

TheLukaszNs-Blazity (dismissed):

TheLukaszNs-Blazity (approved):

PR #1455, opened 2026-03-14
fix: Correct back navigation from Prompt Notebook and Experiment Details

fryz:
## Summary
Fixes incorrect redirect URLs when users navigate back from Prompt Notebook and Experiment Details views, ensuring they land on the expected pages within the task workflow.

## Problem
Users navigating back from Prompt Notebook were incorrectly redirected to the notebooks tab instead of the main prompts view, and users returning from Experiment Details weren't being directed to the prompt experiments tab as expected.

## Solution
Updated navigation logic in both components to redirect to the correct URLs: Prompt Notebook now returns to `/tasks/id/prompts` and Experiment Details returns to `/tasks/id/prompts?tab=prompt-experiments`.

## User Impact
Users now experience consistent and intuitive back navigation when working with prompts and experiments. When leaving the Prompt Notebook, they return to the main prompts view, and when leaving Experiment Details, they return directly to the prompt experiments tab where they can continue their workflow seamlessly.

## Changed Areas
• **Prompt Experiments module** — Updated ExperimentDetailView component back navigation target
• **Prompts Playground module** — Fixed PlaygroundHeader component redirect URL for notebook back navigation

## Type
Bug Fix

## Breaking Changes
None

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,3 +45,8 @@ ml-engine/src/genai_client/**
 models/
 arthur-engine.code-workspace
 
+
+# Gas Town (added by gt)
+.runtime/
+.claude/commands/
+.logs/
--- a/.gitignore
+++ b/.gitignore
@@ -45,8 +45,3 @@ ml-engine/src/genai_client/**
 models/
 arthur-engine.code-workspace
 
-
-# Gas Town (added by gt)
-.runtime/
-.claude/commands/
-.logs/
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -45,3 +45,8 @@ ml-engine/src/genai_client/**
 models/
 arthur-engine.code-workspace
 
+
+# Gas Town (added by gt)
+.runtime/
+.claude/commands/
+.logs/
```
fryz: these need to be managed in a global gitignore on your machine

TheLukaszNs-Blazity (approved):

PR #1456, opened 2026-03-14
feat(UP-4039): Auto-select single available prompt version in experiment creation

fryz:
Fixes UP-4039: UX improvement — auto-select single available prompt version when creating an experiment.

**Changes:**
- When creating a prompt experiment, if only one version of a prompt is available, it is now auto-selected instead of requiring manual selection.

**AC:** If there is only one version, make it auto-selected so the user doesn't have to select it manually.

Polecat: furiosa | Source: ae-40h | MR: ae-wisp-9l1ku7


.gitignore
```diff
@@ -45,3 +45,8 @@ ml-engine/src/genai_client/**
 models/
 arthur-engine.code-workspace
 
+
+# Gas Town (added by gt)
+.runtime/
+.claude/commands/
+.logs/
```
fryz: these should not ever be checked in - we need to manage these gitignores in a global gitignore on your machine

martin-arthur-ai (approved):

xV3rt (approved):

PR #1470, opened 2026-03-18
Add arm64 CPU images for Mac users

ntatsumi:
Add arm64 image for Mac users: The docker compose deployment using images on Docker Hub currently does not work for Mac users because the x86 emulation is not working well for our very large image.

ARM64 images will be available for the CPU version of the engine only.


jdbarthur (approved):
looks good!

PR #1495, opened 2026-03-23
fix(tracer): Claude Code tracer reliability fixes

fryz:
## Summary

A series of fixes to the Claude Code tracer (`integrations/claude-code/claude_code_tracer.py`) addressing reliability issues discovered through real-world usage:

- **Token counting** (`80452f72`): Group multi-entry API responses by `message.id` so input/output/cache tokens are summed correctly per LLM call instead of double-counted
- **Subagent context** (`d27657c9`): Propagate `traceparent` header into Agent tool spawns so subagent spans link back to the parent trace
- **Transcript path caching** (`d6c69b87`): Cache resolved transcript path in session state to prevent mid-session redirect when worktree paths change
- **Shadow transcript filtering** (`0ad76261`): Skip `pr-link`-only JSONL files created by `gh pr create` in worktrees — these were causing LLM span extraction to silently produce 0 spans
- **Turn boundary** (`88e94451`): Stop LLM span extraction at the next user message boundary to prevent bleeding spans from future turns into the current one
- **Context continuation** (`46b874ae`): Detect when Claude Code compresses context mid-session (no `UserPromptSubmit` fires) and correctly start a new trace; also always update `last_llm_output` so the CHAIN span output shows the final text response rather than stale tool JSON

## Test plan
- [ ] All existing unit tests pass: `cd integrations/claude-code && python -m pytest test_tracer.py -v`
- [ ] 8 new regression tests added covering the fixed scenarios
- [ ] Manually verified in Arthur Engine UI: correct per-turn traces, Skill spans captured, final assistant response as CHAIN output


integrations/claude-code/claude_code_tracer.py
```diff
@@ -1045,11 +1276,29 @@ def handle_pre_tool(data: dict, config: dict) -> None:
...
+    # by more than one human message since that trace started.
+    if state.get("current_trace"):
+        ct = state["current_trace"]
+        _tp = _get_cached_transcript_path(data, state, session_id)
+        if _tp:
+            _cur_human = _count_human_messages(_tp)
+            _start = ct.get("human_count_at_start", 0)
+            if _cur_human > _start + 1:
+                # A new turn started without UserPromptSubmit. Complete the old
+                # trace and fall through to the fallback that creates a new one.
```
ntatsumi, line 1291: This context continuation detection won't trigger for traces that don't trigger tool calls.

Adding the same _cur_human > _start + 1 detection at the top of handle_stop is likely a good place to cover the gap.

ntatsumi (approved):

PR #1502, opened 2026-03-24
UP-3635: Transform versioning

iabouhashish:
## Summary
- Add `transform_versions` data model and Alembic migration
- Implement version history endpoints: list, get snapshot, restore (non-destructive)
- Regenerate OpenAPI spec
- Frontend Edit History panel listing versions with author/timestamp
- Frontend read-only snapshot viewer
- Frontend restore action with confirmation dialog

## Issues
- ae-pqg: Data model + migration
- ae-gn5: API endpoints + OpenAPI regen
- ae-6ov: Frontend Edit History panel
- ae-bwk: Frontend restore action + confirmation dialog
- ae-kv9: Frontend read-only snapshot viewer

## Accompanying PR:
https://github.com/arthur-ai/arthur-common/pull/121


iabouhashish:
Comment from @notthattal 
Add in the migration to set all the previously existing transforms are v1.

iabouhashish:
Errors should be resolved when: https://github.com/arthur-ai/arthur-common/pull/121 is merged and approved

genai-engine/src/db_models/transform_models.py
```diff
@@ -35,3 +37,40 @@ class DatabaseTraceTransform(Base):
...
+        default=uuid.uuid4,
+    )
+    transform_id: Mapped[uuid.UUID] = mapped_column(
+        UUID,
+        ForeignKey("trace_transforms.id", ondelete="CASCADE"),
+        index=True,
+        nullable=False,
+    )
+    version_number: Mapped[int] = mapped_column(Integer, nullable=False)
+    config_snapshot: Mapped[Dict[str, Any]] = mapped_column(
```
notthattal: when you switch to not storing the definition on the trace transform table anymore and only in this table it would probably save you a lot of headache to just rename this as "definition" since that's what it was on the top-level table and is how we reference it throughout the codebase

genai-engine/src/db_models/transform_models.py
```diff
@@ -35,3 +37,40 @@ class DatabaseTraceTransform(Base):
...
+        postgresql.JSON,
+        nullable=False,
+    )
+    author: Mapped[Optional[str]] = mapped_column(String, nullable=True)
```
notthattal: Spoke about it offline but just adding the comment that this should be removed

genai-engine/src/repositories/trace_transform_repository.py
```diff
@@ -116,16 +126,64 @@ def list_transforms(
...
+    def _create_version(
+        self,
+        db_transform: DatabaseTraceTransform,
+        author: Optional[str] = None,
+    ) -> DatabaseTraceTransformVersion:
+        """Create an immutable version snapshot for the given transform.
+
+        Retries once on IntegrityError to handle concurrent version number assignment.
+        """
+        for attempt in range(2):
```
notthattal: @mcgrawia this would be another case we should be using with_for_update() if we're worried about concurrency issues here right?

genai-engine/src/repositories/trace_transform_repository.py
```diff
@@ -164,15 +223,109 @@ def update_transform(
...
+            transform_id=db_version.transform_id,
+            version_number=db_version.version_number,
+            config_snapshot=TraceTransformDefinition.model_validate(
+                db_version.config_snapshot,
+            ),
+            author=db_version.author,
+            created_at=db_version.created_at,
+        )
+
+    def restore_version(
```
notthattal: this can be removed when we remove the definition from the top-level transform because users should just be able to select which version they want to use

genai-engine/src/routers/v1/transform_routes.py
```diff
@@ -182,6 +189,83 @@ def update_transform(
...
+                detail=f"Transform {transform_id} not found",
+            )
+        return repo.get_version_by_id(transform_id, version_id)
+    except HTTPException:
+        raise
+    except Exception as e:
+        raise HTTPException(status_code=500, detail=str(e))
+
+
+@transform_routes.post(
```
notthattal: this could also be removed with the update

genai-engine/alembic/versions/2026_03_23_1200-989b179b86b3_create_transform_versions_table.py
```diff
@@ -0,0 +1,89 @@
...
+                    (id, transform_id, version_number, config_snapshot, author, created_at)
+                VALUES
+                    (:id, :transform_id, :version_number, CAST(:config_snapshot AS json), :author, :created_at)
+                """,
+            ),
+            rows,
+        )
+
+
+def downgrade() -> None:
```
notthattal, line 83: when you upgrade to removing the definition on the transform to the version in this downgrade you're going to want to add probably the latest version's definition (config_snapshot) as the definition back in the top level transform table

genai-engine/src/repositories/trace_transform_repository.py
```diff
@@ -116,16 +126,64 @@ def list_transforms(
...
+                if attempt == 1:
+                    raise HTTPException(
+                        status_code=409,
+                        detail="Version creation conflict due to concurrent request, please retry.",
+                    )
+        raise HTTPException(
+            status_code=500,
+            detail="Unexpected error creating version.",
+        )
+
```
notthattal, line 175: We probably also want to have the ability to delete a transform version and clear its data similar to how prompts and evals also allow you to do that

genai-engine/src/repositories/trace_transform_repository.py
mcgrawia: in this case I think it's ok because there's a unique constraint on the table for (transform_id, version_num) so if there's a race condition, the first one to commit will win. 

That said, I'm a bit confused by the `begin_nested()` transaction below. If there is a conflict, I don't think we'll know about it until `.commit()` is called in the caller? Unless does the conflict appear on `.flush()`?

mcgrawia: Did some claude digging and I think we're ok here. It's quite a complex race condition but the flush should cause the inserts to happen, and postgres will validate the insert even against other uncommitted inserts. The loser transaction will raise an error and revert back to the point in the transaction before the begin_nested() call, and which point it will try the loop again and it will see the newer version that was just committed when it queries for the max

https://claude.ai/share/f756ad81-f41a-4d5c-87ff-fe8f1fc3440b

genai-engine/src/repositories/trace_transform_repository.py
```diff
@@ -70,13 +79,27 @@ def _get_db_transform_by_id(
...
+        return TraceTransformDefinition.model_validate(version.definition)
+
     def get_transform_by_id(self, transform_id: UUID) -> TraceTransform | None:
         db_transform = self._get_db_transform_by_id(transform_id)
 
         if not db_transform:
             return None
 
-        return TraceTransform.from_db_model(db_transform)
+        definition = self._get_latest_definition(transform_id)
```
notthattal: definition shouldn't be on the TraceTransform object anymore since we removed it from the db model. They should just get whatever fields are on the tracetransform db model and if they want to see the definition they can get the transform version

genai-engine/src/repositories/trace_transform_repository.py
```diff
@@ -136,7 +230,7 @@ def create_transform(
                 headers={"full_stacktrace": "false"},
             )
 
-        return TraceTransform.from_db_model(db_transform)
+        return TraceTransform.from_db_model(db_transform, trace_transform.definition)
```
notthattal: same thing here

genai-engine/src/routers/v1/continuous_eval_routes.py
```diff
@@ -371,6 +396,19 @@ def update_continuous_eval(
...
+                update_request.transform_version_id
+                if "transform_version_id" in update_request.model_fields_set
+                else existing_eval.transform_version_id
+            )
+            if effective_version_id is not None:
+                pinned_version = transform_repo.get_version_by_id(
+                    transform_id,
+                    effective_version_id,
+                )
+                transform.definition = pinned_version.definition
```
notthattal: wouldn't this not work anymore because we removed the definition of the transform?

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
```diff
@@ -196,8 +196,18 @@ def _execute_job(self, job: ContinuousEvalJob) -> None:
...
 
+            # Use pinned version's definition if set, otherwise use current definition
+            if db_continuous_eval.transform_version_id is not None:
+                pinned_version = trace_transform_repository.get_version_by_id(
+                    db_continuous_eval.transform_id,
+                    db_continuous_eval.transform_version_id,
+                )
+                transform_definition = pinned_version.definition
+            else:
+                transform_definition = trace_transform.definition
```
notthattal: same here

notthattal (approved):

PR #1565, opened 2026-04-07
UP-4109 eval bulk testing

mcgrawia:
## Overview
1. Adds new UX and backend functionality for executing a CE over a set of trace IDs
2. adds a loading state to the traces view table so it doesn't show "no traces" when first loading

## Demo Vid

https://github.com/user-attachments/assets/3f916429-3282-47a3-a673-23cf72b83938


genai-engine/tests/unit/routes/tasks/test_continuous_eval_test_run_routes.py
```diff
@@ -0,0 +1,701 @@
...
+            client, task.id, llm_eval, transform,
+        )
+        assert status_code == 200
+
+        status_code, response = client.create_test_run(
+            eval_id=str(continuous_eval.id),
+            trace_ids=[],
+        )
+        assert status_code == 400
+    finally:
```
xV3rt: Nitpicking, but the docstring suggests we return 422 while in reality we return 400

genai-engine/src/schemas/enums.py
```diff
@@ -59,6 +59,12 @@ class ClaimClassifierResultEnum(str, Enum):
     DIALOG = "dialog"
 
 
+class TestRunStatus(str, Enum):
+    RUNNING = "running"
+    COMPLETED = "completed"
+    PARTIAL_FAILURE = "partial_failure"
```
notthattal, line 65: Seems like we set to PARTIAL_FAILURE if we spot any issues but wouldn't we want an ERROR status if every test failed due to issues?

xV3rt (dismissed):
The frontend part looks good to me, not seeing any issues

genai-engine/src/repositories/continuous_eval_test_run_repository.py
```diff
@@ -0,0 +1,262 @@
...
+            )
+            self.db_session.add(annotation)
+            annotations.append(annotation)
+
+        # Commit all annotations so worker threads can see them
+        self.db_session.commit()
+
+        # Now enqueue jobs — annotations are visible to workers
+        for annotation in annotations:
+            assert annotation.trace_id is not None
```
notthattal, line 128: Did we decide against changing it to make trace_id a mandatory parameter? Because this wouldn't need to be here if we switched it right?

notthattal:
Instead of having to exclude test runs explicitly from agentic annotations, would it make sense for it to live in its own table? I know it would be extra work having to migrate functionality from the queue service too since that tries to pull from the agentic annotations table. Just noticed we were having to explicitly exclude it from the relationship at the db level and was curious if this other approach was considered

genai-engine/src/repositories/continuous_eval_test_run_repository.py
mcgrawia: Since this MR was already sizable, I left it as it is for now

genai-engine/tests/unit/routes/tasks/test_continuous_eval_test_run_routes.py
mcgrawia: thanks will fix

mcgrawia:
@notthattal I think since it ultimately is the same resource, AgenticAnnotation, it belongs in the same table. Like you mentioned, in the future we may have annotations on a number of things, so I think it's best to build on top of the existing foundation instead of making a new one

notthattal:
> @notthattal I think since it ultimately is the same resource, AgenticAnnotation, it belongs in the same table. Like you mentioned, in the future we may have annotations on a number of things, so I think it's best to build on top of the existing foundation instead of making a new one

Yeah true it only caught my eye because of the explicit exception we had to make for the relationship on trace metadata but the solution makes sense to me

genai-engine/alembic/versions/2026_04_07_1523-6939d74b3750_add_continuous_eval_test_table.py
```diff
@@ -0,0 +1,91 @@
...
+revision = "6939d74b3750"
+down_revision = "04c5e8528072"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.create_table(
+        "continuous_eval_test_runs",
```
videetparekh, line 22: This runs on a list of traces right? Does it make sense to maintain a list of trace IDs so we know which ones passed and failed?

genai-engine/src/schemas/enums.py
mcgrawia: just pushed changes for adding an error status if everything fails/skips

notthattal (approved):
Changes look good to me

genai-engine/alembic/versions/2026_04_07_1523-6939d74b3750_add_continuous_eval_test_table.py
videetparekh: As far as I can tell there's no way to identify if a trace passed or failed a continuous eval run today.

notthattal: The result lives in the agentic annotations table with a pass/failed/skipped/errored result

videetparekh: Oop, thanks

PR #1567, opened 2026-04-08
feat: ML eval support with continuous eval integration and unified evals API

iabouhashish:
## Summary

This PR introduces first-class support for **ML evaluators** (built-in model-based scorers) alongside the existing LLM evaluators, with full CRUD management, continuous eval integration, and a unified evals API.

**ML Evaluators Backend**
- New `DatabaseMLEval` + `DatabaseMLEvalVersionTag` DB models with composite PK `(task_id, name, version)`
- `MLEvalsRepository` — CRUD with version management, soft-delete, tag support, and lazy scorer initialization via `get_ml_scorer()` registry
- `MLEvaluator` — `BaseEvaluator` implementation that dispatches to ML scorer wrappers

**New API Routes**
- Full ML evals CRUD under `/api/v2/tasks/{task_id}/ml_evals/...`
- Unified `/api/v2/tasks/{task_id}/evals` endpoint returning both LLM and ML evals in a single paginated list

**Continuous Eval Integration**
- `ContinuousEvalCreateRequest` updated to support `eval_type: llm_eval | ml_eval` with appropriate field validation
- `get_continuous_eval_variables_and_mappings` updated to serve both eval types
- LLM-first scheduling in the continuous eval queue (ML evals load local models; LLM evals are fast API calls)

**ML Scorer Wrappers** (`scorer/ml_scorers.py`)
- `PIIScorerV2` — wraps GLiNER + Presidio (BinaryPIIDataClassifier)
- `PIIScorerV1` — wraps Presidio-only (BinaryPIIDataClassifierV1)
- `ToxicityMLScorer` — wraps RoBERTa toxicity classifier
- `PromptInjectionMLScorer` — wraps DebertaV3 prompt injection classifier

**Frontend**
- New `MLEvaluators` panel with `MLEvalFormModal` (create/edit) and `MLEvalActions` (delete versions)
- `EvaluatorAccordionList` — unified accordion showing both LLM and ML evaluators
- `EvaluatorPipelinesPanel` / `MLEvaluatorPipelinesPanel` — pipeline views per eval type
- `CreateEvalTypeModal` — entry point for choosing LLM vs ML evaluator creation
- `useAllEvals`, `useDeleteMLEvalMutation`, `useDeleteMlEvalVersionMutation`, `useMlEvals`, `useMlEval`, `useMlEvalVersions`, `useCreateMlEvalMutation` hooks

---

## Architecture

```
API Layer (v2)
├── GET/POST/DELETE /tasks/{id}/ml_evals/{name}           → ml_eval_routes.py
├── GET /tasks/{id}/ml_evals/{name}/versions              → ml_eval_routes.py
├── POST /tasks/{id}/ml_evals/{name}/versions/{v}/run     → ml_eval_routes.py
└── GET /tasks/{id}/evals                                 → eval_routes.py (unified)

Repository Layer
├── MLEvalsRepository           → CRUD + version mgmt for DatabaseMLEval
│   ├── save_ml_eval()          → upsert with auto-increment version
│   ├── get_ml_eval()           → by name + version (supports "latest")
│   ├── list_versions()         → all versions with tag metadata
│   ├── get_all_metadata()      → summary list across all eval names
│   ├── delete_version()        → soft-delete by version
│   └── delete_all_versions()   → hard-delete all versions of an eval
└── MLEvaluator (BaseEvaluator) → run() dispatches to get_ml_scorer()

Scorer Registry (lazy init)
├── get_ml_scorer("pii")        → PIIScorerV2 (GLiNER + Presidio)
├── get_ml_scorer("pii_v1")     → PIIScorerV1 (Presidio only)
├── get_ml_scorer("toxicity")   → ToxicityMLScorer (RoBERTa)
└── get_ml_scorer("prompt_injection") → PromptInjectionMLScorer (DeBERTa)

DB Schema
├── ml_evals (task_id, name, version PK) + ml_eval_type, config, variables, created_at, deleted_at
├── ml_eval_version_tags (task_id, name, version, tag) + unique constraint (task_id, name, tag)
└── continuous_evals updated: + eval_type, ml_eval_name, ml_eval_version columns; dropped FK to llm_evals
```

---

## Design Changes

**Unified Evaluators UX** (`ui/src/components/evaluators/`)
- `Evaluators.tsx` now renders an `EvaluatorAccordionList` with sections for LLM and ML evaluators
- `CreateEvalTypeModal` lets users choose between creating an LLM evaluator (existing flow) or ML evaluator (new flow)
- `MLEvaluators.tsx` shows a table of ML evals with inline version management
- `MLEvalFormModal.tsx` — form for creating/editing ML evals, with type selector (PII v2, PII v1, toxicity, prompt injection) and config options

**Live Evals (Continuous Evals) Updates**
- `ContinuousEvalStepper` and `EvaluatorSelector` updated to support ML eval selection
- `useContinuousEvalVariableMapping` passes `eval_type` query param
- Edit/create forms pass `eval_type` through the full creation flow

---

### ML Eval CRUD

```
POST   /api/v2/tasks/{task_id}/ml_evals/{eval_name}
Body:  { "ml_eval_type": "pii" | "pii_v1" | "toxicity" | "prompt_injection", "config": {...} }
Returns: MLEval

GET    /api/v2/tasks/{task_id}/ml_evals/{eval_name}/versions/{eval_version}
       eval_version: "latest" | <int> | <tag>
Returns: MLEval

GET    /api/v2/tasks/{task_id}/ml_evals/{eval_name}/versions
Returns: { versions: MLVersionResponse[], count: int }

GET    /api/v2/tasks/{task_id}/ml_evals
Returns: { ml_metadata: MLGetAllMetadataResponse[], count: int }

DELETE /api/v2/tasks/{task_id}/ml_evals/{eval_name}                   → 204 (hard-delete all)
DELETE /api/v2/tasks/{task_id}/ml_evals/{eval_name}/versions/{v}      → MLEval (soft-delete)

POST   /api/v2/tasks/{task_id}/ml_evals/{eval_name}/versions/{v}/run
Body:  { "text": "string to evaluate" }
Returns: EvalRunResponse { passed: bool, reason: str, score: float | null }
```

### Unified Evals

```
GET /api/v2/tasks/{task_id}/evals
Returns: AllEvalsMetadataListResponse {
  evals: EvalMetadataItem[],  // eval_type: "llm" | "ml"
  count: int
}
```

### Continuous Eval Updates

```
POST /api/v1/tasks/{task_id}/continuous_evals
Body now supports:
{
  "eval_type": "llm_eval" | "ml_eval",          // NEW field
  "llm_eval_name": str | null,                  // required when eval_type="llm_eval"
  "llm_eval_version": str | int | null,
  "ml_eval_name": str | null,                   // required when eval_type="ml_eval"  (NEW)
  "ml_eval_version": str | int | null           // (NEW)
}

GET /api/v1/tasks/{task_id}/continuous_evals/{id}/variables_and_mappings?eval_type=llm_eval|ml_eval
  (eval_type query param is NEW — defaults to "llm_eval" for backward compat)
```

---

## Database Migration

Migration `f3d8c92e7a14` (`2026_04_06_1200-ml_evals_and_continuous_eval_ml_support`):

1. Creates `ml_evals` table with composite PK and FK to `tasks`
2. Creates `ml_eval_version_tags` table with unique constraint on `(task_id, name, tag)`
3. Updates `continuous_evals`:
   - Drops old FK constraint to `llm_evals` (was non-nullable)
   - Adds `eval_type` column (default `"llm_eval"`)
   - Adds `ml_eval_name` and `ml_eval_version` (nullable)
   - Converts `llm_eval_name` + `llm_eval_version` to nullable
   - Backfills `eval_type = "llm_eval"` for all existing rows

Downgrade reverses all changes.

## Test plan
- [ ] Run `uv run alembic upgrade head` to apply DB migration
- [ ] Verify `GET /api/v2/tasks/{id}/evals` returns both LLM and ML evals
- [ ] Create an ML eval via `POST /api/v2/tasks/{id}/ml_evals/my-pii-eval` with `ml_eval_type: "pii"`
- [ ] Run the eval via `POST .../run` with a text containing PII
- [ ] Create a continuous eval referencing an ML eval (`eval_type: "ml_eval"`)
- [ ] Verify `GET .../variables_and_mappings?eval_type=ml_eval` returns correct variables
- [ ] UI: Open Evaluators tab, verify LLM and ML evaluator sections render
- [ ] UI: Create an ML evaluator and verify it appears in the list


genai-engine/src/scorer/base_ml_scorer.py
```diff
@@ -0,0 +1,30 @@
...
+        """Score the given text and return a V2 result.
+
+        Args:
+            text: The text to evaluate.
+            config: Evaluator-specific configuration (thresholds, entity lists, etc.).
+
+        Returns:
+            MLScoreResult with passed, reason, and optional details.
+        """
+        ...
```
fryz: typically here you just `pass`

fryz:
Reviewed in-person - to summarize: 

* Reuse the same storage (the schema's will eventually end up being the same)
    * We will introduce the type of eval (eg: LLMJudge and ML) as well as a sub-type (eg: LLMJudge, PII, Toxicity, etc.) 
* Reusing the same request/response schemas - an Eval is an Eval 
* Reuse the same routes (eg: there should only be 1 load_evals endpoint and it should load all LLMJudge + ML evals since these are the same objects but with different type differentiator fields)

genai-engine/src/schemas/request_schemas.py
```diff
@@ -933,6 +933,34 @@ class TransformListFilterRequest(BaseModel):
...
+ML_EVAL_TYPE_TOXICITY = "toxicity"
+ML_EVAL_TYPE_PROMPT_INJECTION = "prompt_injection"
+
+
+class CreateMLEvalRequest(BaseModel):
+    """Request schema for creating a new ML eval version."""
+
+    ml_eval_type: str = Field(
+        description="Type of built-in ML scorer (e.g. 'pii', 'toxicity', 'prompt_injection').",
+    )
```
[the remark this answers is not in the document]
notthattal: Agree with this comment but it should be a non-optional Enum instead of literal at that point

genai-engine/src/db_models/llm_eval_models.py
```diff
@@ -36,12 +36,19 @@ class DatabaseLLMEval(Base):
     )
     name: Mapped[str] = mapped_column(String, primary_key=True)
 
-    # LLM Configuration
-    model_name: Mapped[str] = mapped_column(String, nullable=False)
-    model_provider: Mapped[str] = mapped_column(String, nullable=False)
+    # Eval type — "llm_as_a_judge" | "pii" | "toxicity" | "prompt_injection" | ...
+    eval_type: Mapped[str] = mapped_column(
```
notthattal: I would just make this an enum

genai-engine/src/db_models/llm_eval_models.py
```diff
@@ -120,10 +127,18 @@ class DatabaseContinuousEval(Base):
...
+
+    # Discriminator — "llm_eval" or "ml_eval"
+    eval_type: Mapped[str] = mapped_column(
+        String,
+        nullable=False,
+        server_default="llm_eval",
+        default="llm_eval",
+    )
+
+    llm_eval_name: Mapped[Optional[str]] = mapped_column(String, nullable=True)
```
notthattal: why are these optional now? Don't they still exist on ml evals?

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    @abstractmethod
+    def get_eval_variables(
+        self,
+        task_id: str,
+        eval_name: str,
+        eval_version: str,
+    ) -> List[str]:
+        """Return the list of variable names expected by the evaluator."""
+        ...
```
notthattal: agree with zach I would switch ... for pass

genai-engine/src/repositories/ml_evals_repository.py
```diff
@@ -0,0 +1,172 @@
...
+_ML_SCORER_REGISTRY: Dict[str, Any] = {}
+
+
+def get_ml_scorer(eval_type: str) -> Optional[Any]:
+    """Return a cached BaseMLScorer for the given eval_type, or None if unknown."""
+    if eval_type not in ML_EVAL_TYPES:
+        return None
+    if eval_type not in _ML_SCORER_REGISTRY:
+        if eval_type == "pii":
+            from scorer.ml_scorers import PIIScorerV2
```
notthattal: imports for all of these should be at the top and will PIIScorerV2 work when the env var GENAI_ENGINE_USE_PII_MODEL_V2 is False? i'm pretty sure we don't download the gliner model if it's false

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -33,12 +41,18 @@
...
 
+logger = logging.getLogger(__name__)
+
 llm_eval_routes = APIRouter(
     prefix="/api/v1",
     route_class=GenaiEngineRoute,
 )
 
 
+def _repo(db_session: Session) -> LLMEvalsRepository:
```
notthattal: This works but doesn't really follow convention for how we setup repository classes across our other routers. I would just instantiate the repo inside each route instead of having this helper

genai-engine/src/schemas/request_schemas.py
```diff
@@ -827,6 +827,62 @@ class CreateEvalRequest(BaseModel):
     )
 
 
+class CreateMLEvalRequest(BaseModel):
+    eval_type: Literal["pii", "pii_v1", "toxicity", "prompt_injection"] = Field(
+        description="Type of ML eval (e.g. 'pii', 'toxicity', 'prompt_injection')",
+    )
+    config: Optional[Any] = Field(
```
notthattal: This should be a strongly typed class instead of Any if we want to have configs

notthattal: It seems like we could also have different subclasses that are like PIIEvalRequest, ToxicityEvalRequest that would each have their own config and then this class could have eval and be one of the subclasses or something

genai-engine/src/schemas/request_schemas.py
```diff
@@ -827,6 +827,62 @@ class CreateEvalRequest(BaseModel):
...
+    )
+    model_provider: Optional[ModelProvider] = Field(
+        default=None,
+        description="LLM model provider — required for llm_as_a_judge evals",
+    )
+    instructions: Optional[str] = Field(
+        default=None,
+        description="Eval instructions — required for llm_as_a_judge evals",
+    )
+    config: Optional[Any] = Field(
```
notthattal: same thing here

genai-engine/src/scorer/ml_scorers.py
```diff
@@ -0,0 +1,120 @@
...
+        if not reason:
+            reason = "No PII detected." if passed else "PII detected."
+        return MLScoreResult(passed=passed, reason=reason)
+
+
+class ToxicityMLScorer(BaseMLScorer):
+    """Wraps ToxicityScorer."""
+
+    def __init__(self) -> None:
+        from scorer.checks.toxicity.toxicity import ToxicityScorer
```
notthattal: these should be at the top

genai-engine/src/scorer/ml_scorers.py
```diff
@@ -0,0 +1,120 @@
...
+        if not reason:
+            reason = "No PII detected." if passed else "PII detected."
+        return MLScoreResult(passed=passed, reason=reason)
+
+
+class PIIScorerV1(BaseMLScorer):
+    """Wraps BinaryPIIDataClassifierV1 (Presidio only)."""
+
+    def __init__(self) -> None:
+        from scorer.checks.pii.classifier_v1 import BinaryPIIDataClassifierV1
```
notthattal: this should also be at the top

genai-engine/src/scorer/ml_scorers.py
```diff
@@ -0,0 +1,120 @@
...
+        if not reason:
+            reason = "No toxicity detected." if passed else "Toxic content detected."
+        return MLScoreResult(passed=passed, reason=reason)
+
+
+class PromptInjectionMLScorer(BaseMLScorer):
+    """Wraps BinaryPromptInjectionClassifier."""
+
+    def __init__(self) -> None:
+        from scorer.checks.prompt_injection.classifier import (
```
notthattal: same here

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
```diff
@@ -221,40 +220,57 @@ def _execute_job(self, job: ContinuousEvalJob) -> None:
...
+                raise ValueError(
+                    f"Continuous eval {continuous_eval.id} has no eval name configured",
+                )
+            if continuous_eval.llm_eval_version is None:
+                raise ValueError(
+                    f"Continuous eval {continuous_eval.id} has no eval version configured",
+                )
+
+            # Look up eval_type from the llm_evals row so we dispatch to the right executor.
+            from db_models.llm_eval_models import DatabaseLLMEval
```
notthattal: and here

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,60 @@
...
+        """
+        ...
+
+
+def get_evaluator(
+    db_session: Session,
+    eval_type: str = "llm_as_a_judge",
+) -> "BaseEvaluator":
+    """Factory: return the right evaluator for the given eval_type."""
+    from repositories.ml_evals_repository import ML_EVAL_TYPES, MLEvaluator
```
notthattal: and this

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,60 @@
...
+
+    if eval_type in ML_EVAL_TYPES:
+        return MLEvaluator(db_session)
+
+    from repositories.llm_evaluator import LLMEvaluator
```
notthattal: here too

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
```diff
@@ -221,40 +220,57 @@ def _execute_job(self, job: ContinuousEvalJob) -> None:
...
+            db_llm_eval = (
+                db_session.query(DatabaseLLMEval)
+                .filter(
+                    DatabaseLLMEval.task_id == continuous_eval.task_id,
+                    DatabaseLLMEval.name == eval_name,
+                    DatabaseLLMEval.version == continuous_eval.llm_eval_version,
+                )
+                .first()
+            )
+            eval_type = db_llm_eval.eval_type if db_llm_eval else "llm_as_a_judge"
```
notthattal: shouldn't this not work if we can't get the llm eval from the db? seems weird that we are defaulting to an llm as a judge evaluator when we couldn't find an eval in the db

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
```diff
@@ -221,40 +220,57 @@ def _execute_job(self, job: ContinuousEvalJob) -> None:
...
+                )
+            if continuous_eval.llm_eval_version is None:
+                raise ValueError(
+                    f"Continuous eval {continuous_eval.id} has no eval version configured",
+                )
+
+            # Look up eval_type from the llm_evals row so we dispatch to the right executor.
+            from db_models.llm_eval_models import DatabaseLLMEval
+
+            db_llm_eval = (
```
notthattal: is this query not what we were doing before by calling llm_eval_repository.get_llm_item? Seems unnecessary to add this query here when we already had it working with that function

genai-engine/src/services/experiment_executor.py
```diff
@@ -690,29 +692,53 @@ def _execute_eval_with_variable_map(
...
-            llm_evals_repo.model_provider_repo = model_provider_repo
-
-            # Create completion request with variables
-            completion_request = BaseCompletionRequest(
-                variables=[
-                    VariableTemplateValue(name=k, value=v)
-                    for k, v in variable_map.items()
-                ],
+            # Look up the eval's type so we can dispatch to the right evaluator.
+            from db_models.llm_eval_models import DatabaseLLMEval
```
notthattal: same thing here for the imports and below with having an else for an eval that doesn't exist

genai-engine/src/services/experiment_executor.py
```diff
@@ -690,29 +692,53 @@ def _execute_eval_with_variable_map(
...
+                if isinstance(evaluator, MLEvaluator):
+                    # ML evals (pii, toxicity, prompt_injection) run through the scorer directly.
+                    eval_response = evaluator.run(
+                        task_id=experiment.task_id,
+                        eval_name=eval_score.eval_name,
+                        eval_version=str(eval_score.eval_version),
+                        variable_mapping=[],
+                        resolved_variables=variable_map,
+                    )
+                else:
```
notthattal: I would probably make this `elif isinstance(evaluator, LLMEval)` and have an else that raises an error in case we decide for some reason to add another type of eval in the future it doesn't go through this pipeline that fails

genai-engine/tests/unit/repositories/test_ml_evals_repository.py
```diff
@@ -0,0 +1,347 @@
...
+    assert result.config == {"threshold": 0.8}
+
+    ml_evals_repo.delete_llm_item(task_id, eval_name)
+
+
+@pytest.mark.unit_tests
+def test_save_ml_eval_rejects_unknown_type(ml_evals_repo, task_id):
+    # CreateMLEvalRequest.eval_type is a Literal — invalid values are rejected
+    # at Pydantic parse time (ValidationError), before the repository is called.
+    from pydantic import ValidationError
```
notthattal: same here for the imports

genai-engine/tests/unit/repositories/test_ml_evals_repository.py
```diff
@@ -0,0 +1,347 @@
...
+
+# ---------------------------------------------------------------------------
+# eval_types isolation — ML evals must not see llm_as_a_judge rows
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.unit_tests
+def test_ml_repo_does_not_see_llm_as_a_judge_evals(task_id):
+    """ML repo must not surface llm_as_a_judge rows; LLM repo surfaces all eval types."""
+    from arthur_common.models.llm_model_providers import ModelProvider
```
notthattal: and here

notthattal:
I would also test out your downgrades for your alembic revisions in this repo. I was running into a lot of issues when trying to downgrade back to the changes before this PR

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def get_evaluator(
+    db_session: Session,
+    eval_type: EvalType = EvalType.LLM_AS_A_JUDGE,
+) -> "BaseEvaluator":
+    """Factory: return the right evaluator for the given eval_type."""
+    # Imports are intentionally deferred to avoid circular imports:
+    # both MLEvaluator and LLMEvaluator import BaseEvaluator at module level.
+    from repositories.ml_evals_repository import ML_EVAL_TYPES, MLEvaluator
```
notthattal: We shouldn't have circular imports in the repo. This function probably shouldn't live in this file if it causes a circular import

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,61 @@
...
+
+    if eval_type in ML_EVAL_TYPES:
+        return MLEvaluator(db_session)
+
+    from repositories.llm_evaluator import LLMEvaluator
```
notthattal: there's a lot of spots where imports are in the middle of files. Can we move them all to the top

genai-engine/alembic/versions/2026_04_07_1257-711e7d7236ae_add_trace_end_time_index.py
```diff
@@ -5,24 +5,23 @@
...
     # ### commands auto generated by Alembic - please adjust! ###
-    op.create_index('idx_traces_end_time', 'trace_metadata', ['end_time'], unique=False)
+    op.create_index("idx_traces_end_time", "trace_metadata", ["end_time"], unique=False)
     # ### end Alembic commands ###
 
 
 def downgrade() -> None:
     # ### commands auto generated by Alembic - please adjust! ###
-    op.drop_index('idx_traces_end_time', table_name='trace_metadata')
+    op.execute("DROP INDEX IF EXISTS idx_traces_end_time")
```
notthattal: this seems unrelated to the ml eval stuff. Was there a reason this needed to be updated?

genai-engine/src/db_models/llm_eval_models.py
```diff
@@ -155,12 +178,3 @@ class DatabaseContinuousEval(Base):
         nullable=False,
     )
     enabled: Mapped[bool] = mapped_column(Boolean, default=True, nullable=False)
-
```
notthattal, line 158: why did we remove the FK constraint?

genai-engine/src/repositories/ml_evals_repository.py
```diff
@@ -0,0 +1,189 @@
...
+
+# Module-level scorer cache — scorers load heavyweight models on first use.
+_ML_SCORER_REGISTRY: Dict[str, Any] = {}
+
+
+def get_ml_scorer(eval_type: str) -> Optional[Any]:
+    """Return a cached BaseMLScorer for the given eval_type, or None if unknown."""
+    if eval_type not in ML_EVAL_TYPES:
+        return None
+    if eval_type not in _ML_SCORER_REGISTRY:
```
notthattal: same thing here with moving the imports to the top

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,61 @@
...
+    """Abstract base for all evaluator implementations."""
+
+    @abstractmethod
+    def get_eval_variables(
+        self,
+        task_id: str,
+        eval_name: str,
+        eval_version: str,
+    ) -> List[str]:
+        """Return the list of variable names expected by the evaluator."""
```
notthattal: we should raise a not implemented error here if we expect all subclasses to override this

genai-engine/src/repositories/base_evaluator.py
```diff
@@ -0,0 +1,61 @@
...
+    @abstractmethod
+    def run(
+        self,
+        task_id: str,
+        eval_name: str,
+        eval_version: str,
+        variable_mapping: List[ContinuousEvalTransformVariableMapping],
+        resolved_variables: dict[str, str],
+    ) -> EvalRunResponse:
+        """Execute the evaluator and return a unified response.
```
notthattal: same with this one

genai-engine/src/repositories/llm_evaluator.py
```diff
@@ -0,0 +1,51 @@
...
+from schemas.internal_schemas import ContinuousEvalTransformVariableMapping
+from schemas.request_schemas import BaseCompletionRequest
+from schemas.response_schemas import EvalRunResponse
+
+
+class LLMEvaluator(BaseEvaluator):
+    def __init__(self, db_session: Session) -> None:
+        self._repo = LLMEvalsRepository(db_session)
+
+    def get_eval_variables(
```
notthattal, line 28: both this one and the ml one do the same thing just with a different repo it seems like it could just live in the parent class

genai-engine/src/repositories/ml_evals_repository.py
```diff
@@ -0,0 +1,189 @@
...
+        )
+        actual_version_num = str(db_item.version)
+        self.soft_delete_llm_item_version(task_id, eval_name, version)
+        return cast(
+            MLEval,
+            super().get_llm_item(task_id, eval_name, actual_version_num),
+        )
+
+
+class MLEvaluator(BaseEvaluator):
```
notthattal, line 106: It seems like we don't need a new parent class and subevaluators for this? Both the ML and LLM Evaluators do the same thing with a different repository for get_eval_variables. It seems like we could just add that to the BaseLLMRepository or create a BaseLLMEvalRepository which inherits from BaseLLMRepository and then the MlEvalRepository and LLMEvalRepository could inherit from that. In the BaseLLMEvalRepository it could just do 

```python
def get_eval_variables(
        self,
        task_id: str,
        eval_name: str,
        eval_version: str,
    ) -> List[str]:
        eval = self..get_llm_item(task_id, eval_name, eval_version)
        return eval.variables
```

and then instantiate the run function as an abstract raising a non implemented error. Then in the sub-repositories it would override the run function and then you don't need to add these new classes

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -53,26 +63,22 @@ def get_llm_eval(
     eval_name: Annotated[str, Path(), AfterValidator(decode_path_param)],
     eval_version: str = Path(
         ...,
-        description="The version of the llm eval to retrieve. Can be 'latest', a version number (e.g. '1', '2', etc.), an ISO datetime string (e.g. '2025-01-01T00:00:00'), or a tag.",
-        title="LLM Eval Version",
+        description="'latest', version number, ISO datetime, or tag.",
```
notthattal: we should probably leave the examples that we had in the description

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -166,18 +168,13 @@ def get_all_llm_evals(
 def run_saved_llm_eval(
     completion_request: BaseCompletionRequest,
     eval_name: Annotated[str, Path(), AfterValidator(decode_path_param)],
-    eval_version: str = Path(
```
notthattal, line 173: same thing here

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -234,22 +234,26 @@ def delete_llm_eval(
     task: Task = Depends(get_validated_task),
 ) -> Response:
     try:
-        llm_eval_service = LLMEvalsRepository(db_session)
-        llm_eval_service.delete_llm_item(task.id, eval_name)
+        LLMEvalsRepository(db_session).delete_llm_item(task.id, eval_name)
+        disabled = ContinuousEvalsRepository(
```
notthattal: this wouldn't be necessary if we kept the FK constraint above

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -259,37 +263,40 @@ def delete_llm_eval(
 @permission_checker(permissions=PermissionLevelsEnum.TASK_WRITE.value)
 def soft_delete_llm_eval_version(
     eval_name: Annotated[str, Path(), AfterValidator(decode_path_param)],
-    eval_version: str = Path(
```
notthattal, line 269: another description that was removed

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -259,37 +263,40 @@ def delete_llm_eval(
...
-        llm_eval_service = LLMEvalsRepository(db_session)
-        llm_eval_service.soft_delete_llm_item_version(
+        LLMEvalsRepository(db_session).soft_delete_llm_item_version(
             task.id,
             eval_name,
             eval_version,
         )
-
+        try:
+            version_int = int(eval_version)
```
notthattal: we should probably just grab the eval before soft deleting it and using that eval version instead of trying and failing because someone used latest

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -259,37 +263,40 @@ def delete_llm_eval(
...
-            raise HTTPException(status_code=400, detail=str(e))
+        raise HTTPException(status_code=400, detail=str(e))
     except Exception as e:
         raise HTTPException(status_code=500, detail=str(e))
 
 
 @llm_eval_routes.get(
     "/tasks/{task_id}/llm_evals/{eval_name}/versions/tags/{tag}",
     summary="Get an llm eval by name and tag",
-    description="Get an llm eval by name and tag",
```
notthattal, line 299: would add this desc back

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -303,96 +310,71 @@ def get_llm_eval_by_tag(
...
     summary="Add a tag to an llm eval version",
-    description="Add a tag to an llm eval version",
     response_model=LLMEval,
     response_model_exclude_none=True,
     tags=["LLMEvals"],
 )
 @permission_checker(permissions=PermissionLevelsEnum.TASK_WRITE.value)
 def add_tag_to_llm_eval_version(
     eval_name: Annotated[str, Path(), AfterValidator(decode_path_param)],
-    eval_version: str = Path(
```
notthattal, line 346: same here

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -303,96 +310,71 @@ def get_llm_eval_by_tag(
...
-            raise HTTPException(status_code=400, detail=str(e))
+        raise HTTPException(status_code=400, detail=str(e))
     except Exception as e:
         raise HTTPException(status_code=500, detail=str(e))
 
 
 @llm_eval_routes.delete(
     "/tasks/{task_id}/llm_evals/{eval_name}/versions/{eval_version}/tags/{tag}",
     summary="Remove a tag from an llm eval version",
-    description="Remove a tag from an llm eval version",
```
notthattal, line 380: and here

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -303,96 +310,71 @@ def get_llm_eval_by_tag(
...
-    responses={
-        status.HTTP_204_NO_CONTENT: {"description": "LLM eval version deleted."},
-    },
+    responses={status.HTTP_204_NO_CONTENT: {"description": "Tag removed."}},
     tags=["LLMEvals"],
 )
 @permission_checker(permissions=PermissionLevelsEnum.TASK_WRITE.value)
 def delete_tag_from_llm_eval_version(
     eval_name: Annotated[str, Path(), AfterValidator(decode_path_param)],
-    eval_version: Annotated[
```
notthattal, line 391: and here

genai-engine/src/routers/v2/ml_eval_routes.py
```diff
@@ -0,0 +1,56 @@
...
+    eval_name: Annotated[str, Path(), AfterValidator(decode_path_param)],
+    db_session: Session = Depends(get_db_session),
+    current_user: User | None = Depends(multi_validator.validate_api_multi_auth),
+    task: Task = Depends(get_validated_task),
+) -> MLEval:
+    try:
+        repo = MLEvalsRepository(db_session)
+        return repo.save_ml_eval(task.id, eval_name, create_request)
+    except ValueError as e:
+        from fastapi import HTTPException
```
notthattal: import should be at the top

genai-engine/src/schemas/request_schemas.py
```diff
@@ -831,6 +832,120 @@ class CreateEvalRequest(BaseModel):
...
+    )
+
+
+class PromptInjectionEvalConfig(BaseModel):
+    """Configuration for prompt injection detection evals."""
+
+
+# Map from EvalType to its typed config model for validation.
+# Config is stored as a raw dict in the DB; these models define the valid fields.
+_ML_EVAL_CONFIG_MODELS: Dict[EvalType, type] = {
```
notthattal: this doesn't seem to be used anywhere?

genai-engine/src/schemas/request_schemas.py
```diff
@@ -831,6 +832,120 @@ class CreateEvalRequest(BaseModel):
...
+}
+
+
+class CreateMLEvalRequest(BaseModel):
+    eval_type: EvalType = Field(
+        description="Type of ML eval (e.g. 'pii', 'toxicity', 'prompt_injection')",
+    )
+    # Config is a free-form dict whose valid fields are defined by the typed models above
+    # (PIIEvalConfig, ToxicityEvalConfig, PromptInjectionEvalConfig).
+    config: Optional[Dict[str, Any]] = Field(
```
notthattal: seems like this should be a UNION of PII, ToxicityConfig, Prompt Injection Config etc. instead of Dict[str,Any]

genai-engine/src/schemas/request_schemas.py
```diff
@@ -831,6 +832,120 @@ class CreateEvalRequest(BaseModel):
...
+    )
+    model_provider: Optional[ModelProvider] = Field(
+        default=None,
+        description="LLM model provider — required for llm_as_a_judge evals",
+    )
+    instructions: Optional[str] = Field(
+        default=None,
+        description="Eval instructions — required for llm_as_a_judge evals",
+    )
+    config: Optional[Dict[str, Any]] = Field(
```
notthattal: same here on what the type should be

genai-engine/src/schemas/request_schemas.py
```diff
@@ -831,6 +832,120 @@ class CreateEvalRequest(BaseModel):
...
+        default=None,
+        description=(
+            "Optional configuration. For llm_as_a_judge: LLM settings (temperature, etc.). "
+            "For ML evals: see PIIEvalConfig, ToxicityEvalConfig, PromptInjectionEvalConfig."
+        ),
+    )
+
+    @model_validator(mode="after")
+    def validate_for_type(self) -> "CreateAnyEvalRequest":
+        if self.eval_type == EvalType.LLM_AS_A_JUDGE:
```
notthattal: seems like a config should be required if it's an ML Eval no?

genai-engine/src/schemas/response_schemas.py
```diff
@@ -755,6 +790,10 @@ class LLMEvalRunResponse(BaseModel):
     cost: str = Field(..., description="Cost of this llm completion")
 
 
+# Alias — ML evals share the same run response shape as LLM evals
+EvalRunResponse = LLMEvalRunResponse
```
notthattal: seems like we could just rename LLMEvalRunResponse -> EvalRunResponse if it's the exact same thing

genai-engine/src/scorer/ml_scorers.py
```diff
@@ -0,0 +1,123 @@
...
+
+
+class PIIScorerV2(BaseMLScorer):
+    """Wraps BinaryPIIDataClassifier (GLiNER + Presidio, V2)."""
+
+    def __init__(self) -> None:
+        # Deferred import: scorer modules load heavyweight ML models (GLiNER, Presidio,
+        # Transformers) on import. Keeping these inside __init__ ensures models are only
+        # loaded when the scorer is first instantiated, not at server startup.
+        from scorer.checks.pii.classifier import BinaryPIIDataClassifier
```
notthattal: should move to the top

genai-engine/src/scorer/ml_scorers.py
```diff
@@ -0,0 +1,123 @@
...
+
+from typing import Any
+
+from arthur_common.models.enums import RuleResultEnum, RuleType
+
+from schemas.scorer_schemas import ScoreRequest
+from scorer.base_ml_scorer import BaseMLScorer, MLScoreResult
+
+
+class PIIScorerV2(BaseMLScorer):
```
notthattal: Instead of recreating all these scorers why not just call the score methods with the params directly? Similar to how the ScorerClient just calls scorer_obj.score(score_request)?

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
```diff
@@ -233,40 +233,49 @@ def _execute_job(self, job: ContinuousEvalJob) -> None:
...
+
+            # Validate that the resolved eval vars match what the evaluator expects
+            expected_vars = set(
+                evaluator.get_eval_variables(
+                    continuous_eval.task_id,
+                    eval_name,
+                    eval_version,
+                ),
+            )
+            mapped_eval_vars = set(resolved_variables.keys())
```
notthattal, line 267: Most of the variables changes above seem like they don't really change anything functionally. Wondering why the switch? get_llm_item should already raise an error so I think adding the try just to re-raise it wouldn't make sense no?

genai-engine/src/services/experiment_executor.py
```diff
@@ -690,29 +694,58 @@ def _execute_eval_with_variable_map(
...
+                        eval_name=eval_score.eval_name,
+                        eval_version=str(eval_score.eval_version),
+                        variable_mapping=[],
+                        resolved_variables=variable_map,
+                    )
+                elif isinstance(evaluator, LLMEvaluator):
+                    # LLM-as-a-judge evals run through the chat completion pipeline.
+                    model_provider_repo = ModelProviderRepository(db_session)
+                    llm_evals_repo = LLMEvalsRepository(db_session)
+                    llm_evals_repo.model_provider_repo = model_provider_repo
```
notthattal: this is already instantiated in the class when you create it no need to reset it here

genai-engine/src/scorer/ml_scorers.py
iabouhashish: If we later swap out or extend the underlying implementation, we only touch the wrapper MLEvaluator and run_ml_scorer stay unchanged. Calling scorer_obj.score(request) directly would couple all callers to the V1 scorer internals.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Fixed — moved `get_evaluator` to a new `evaluator_factory.py` module to break the circular import.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Moved all non-lazy imports to the top across all affected files. The deferred imports in `ml_scorers.py` are intentional — they lazy-load heavyweight models (GLiNER, Transformers, Presidio) to avoid pulling them in at server startup.

genai-engine/alembic/versions/2026_04_07_1257-711e7d7236ae_add_trace_end_time_index.py
iabouhashish: This was a bug in the PR — `downgrade()` was incorrectly calling `op.drop_table` instead of `op.drop_index`. Fixed.

genai-engine/src/db_models/llm_eval_models.py
iabouhashish: Restored with `ondelete="CASCADE"`. Also removed the now-redundant explicit `disable_continuous_evals_by_eval_name` call from the hard-delete path — the DB handles cleanup automatically.

genai-engine/src/repositories/ml_evals_repository.py
iabouhashish: Moved to the top.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Added `raise NotImplementedError` to both abstract methods.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Done.

genai-engine/src/routers/v1/llm_eval_routes.py
iabouhashish: Restored full descriptions with examples on all `Path()` version parameters.

genai-engine/src/routers/v1/llm_eval_routes.py
iabouhashish: Done.

genai-engine/src/routers/v1/llm_eval_routes.py
iabouhashish: Removed — the restored FK constraint with `ondelete="CASCADE"` handles this automatically.

genai-engine/src/schemas/request_schemas.py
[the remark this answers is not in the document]
iabouhashish: `eval_type` is now a required `EvalType` field (non-optional, non-string).

genai-engine/src/scorer/base_ml_scorer.py
iabouhashish: Used `raise NotImplementedError` instead — more explicit that subclasses must implement.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Used `raise NotImplementedError` for the same reason.

genai-engine/src/schemas/request_schemas.py
iabouhashish: Replaced with a typed `MLEvalConfig` union (`PIIEvalConfig | ToxicityEvalConfig | PromptInjectionEvalConfig`) with a `model_validator` that coerces the raw dict based on `eval_type`.

iabouhashish: Added `PIIEvalConfig`, `ToxicityEvalConfig`, and `PromptInjectionEvalConfig` typed models, with a `model_validator(mode="before")` on `CreateMLEvalRequest` that coerces the incoming dict to the correct type based on `eval_type`.

genai-engine/src/schemas/request_schemas.py
iabouhashish: Done — `CreateAnyEvalRequest.config` now coerces to the typed config union as well.

genai-engine/src/scorer/ml_scorers.py
iabouhashish: The heavy model imports (`TOXICITY_MODEL`, `BinaryPIIDataClassifier`, etc.) remain deferred inside `get_ml_scorer()` intentionally — lazy-loading avoids pulling heavyweight models into memory at startup. Non-model imports are now at the top.

genai-engine/src/scorer/ml_scorers.py
iabouhashish: Same — kept deferred for lazy-loading. See above.

genai-engine/src/scorer/ml_scorers.py
iabouhashish: Same.

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
iabouhashish: Moved to the top.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Moved to the top.

genai-engine/src/repositories/base_evaluator.py
iabouhashish: Done.

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
iabouhashish: Removed — `get_llm_item` now raises directly on failure. The fallback to LLM-as-a-judge was wrong behavior.

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
iabouhashish: Simplified — removed the redundant query and restored the direct `get_llm_item` call.

genai-engine/src/services/experiment_executor.py
iabouhashish: Moved imports to the top.

genai-engine/src/services/experiment_executor.py
iabouhashish: Done — `if isinstance(evaluator, MLEvaluator)` / `elif isinstance(evaluator, LLMEvaluator)` / `else: raise ValueError(...)`.

genai-engine/src/routers/v1/llm_eval_routes.py
iabouhashish: Removed the helper — each route now instantiates `LLMEvalsRepository(db_session)` directly, consistent with convention.

genai-engine/src/repositories/llm_evaluator.py
iabouhashish: Kept the wrapper pattern intentionally. The separation of concerns is deliberate — repositories own data access, evaluators own execution logic. Merging them would make it harder to swap the underlying scorer/LLM client, add a non-DB-backed evaluator (e.g. mock or remote), or extract evaluation into a separate service later without touching the persistence layer. Added a docstring to both `LLMEvaluator` and `MLEvaluator` explaining this.

genai-engine/src/repositories/ml_evals_repository.py
iabouhashish: Kept the wrapper pattern intentionally. The separation of concerns is deliberate — repositories own data access, evaluators own execution logic. Merging them would make it harder to swap the underlying scorer/LLM client, add a non-DB-backed evaluator (e.g. mock or remote), or extract evaluation into a separate service later without touching the persistence layer. Added a docstring to both `LLMEvaluator` and `MLEvaluator` explaining this.

genai-engine/src/routers/v1/llm_eval_routes.py
iabouhashish: Restored — the full description with examples is back on all `eval_version` Path parameters in `soft_delete_llm_eval_version` and the other routes.

genai-engine/src/routers/v1/llm_eval_routes.py
iabouhashish: Done — we now resolve the version via `get_llm_item` before calling `soft_delete_llm_item_version`, so passing `'latest'` works correctly.

genai-engine/src/schemas/request_schemas.py
iabouhashish: Removed — `_ML_EVAL_CONFIG_MODELS` is gone. Config validation now uses the typed `PIIEvalConfig`, `ToxicityEvalConfig`, and `PromptInjectionEvalConfig` models directly via the `model_validator`.

genai-engine/src/services/continuous_eval/continuous_eval_queue_service.py
iabouhashish: The try/except is removed — `get_llm_item` now raises directly. The variable structure change (list of `VariableTemplateValue` → `dict[str, str]`) was necessary because `evaluator.run()` takes `resolved_variables: dict[str, str]`; the dict form is what the `BaseEvaluator` interface expects.

genai-engine/src/services/experiment_executor.py
iabouhashish: Removed — `LLMEvalsRepository.__init__` already sets `self.model_provider_repo`, so the manual reassignment was redundant and is now gone.

genai-engine/src/scorer/ml_scorers.py
notthattal: the models are already loaded on startup regardless of if we import it here or not. I still think these should move to the top

genai-engine/src/scorer/ml_scorers.py
notthattal: wait wdym? Because we have a function get_scorer_client() that already splits this up based on the name version mapping. If we wanted to give the option to use both v1 and v2 we could just create a similar helper that allows for both of them and then calls .score(request) on any of our guardrails?

genai-engine/src/routers/v1/llm_eval_routes.py
```diff
@@ -269,19 +274,27 @@ def soft_delete_llm_eval_version(
...
-        llm_eval_service = LLMEvalsRepository(db_session)
-        llm_eval_service.soft_delete_llm_item_version(
-            task.id,
-            eval_name,
-            eval_version,
-        )
-
+        repo = LLMEvalsRepository(db_session)
+        # Resolve version (e.g. "latest") to an actual integer before soft-deleting
+        try:
```
notthattal: you don't need the try/except here bc we already raise an error in get_llm_item

genai-engine/src/scorer/ml_scorers.py
```diff
@@ -0,0 +1,139 @@
...
+from typing import Any, Dict, Optional
+
+from arthur_common.models.enums import RuleResultEnum, RuleType
+
+from schemas.enums import EvalType
+from schemas.response_schemas import EvalRunResponse
+from schemas.scorer_schemas import ScoreRequest
+from scorer.base_ml_scorer import BaseMLScorer
+
+ML_EVAL_TYPES = [
```
notthattal: this global is identical to _ML_EVAL_TYPES we should reuse one of them instead of recreating it

genai-engine/src/repositories/ml_evals_repository.py
notthattal: still don't see anything blocking the pii v2 scorer when that flag is false

genai-engine/src/scorer/ml_scorers.py
iabouhashish: Oh I meant without making any changes we would have to couple to v1 scorer. But more importantly, we might change the way we utilize scorers, so didn't want to have a dependency on them directly.

genai-engine/src/repositories/ml_evals_repository.py
iabouhashish: the conditional import was removed, so it always loads both and lets the user choose.

notthattal (dismissed):

videetparekh (approved):

PR #1568, opened 2026-04-09
feat(ui): add continuous eval name to AnnotationCell schema and table

iabouhashish:
Summary
                                                                                                                                              
  - Adds continuous_eval_name to the ContinuousEvalAnnotation Zod schema so the field is parsed from the API response                         
  - Adds a new "Continuous Eval Name" column as the second column in the annotations table dialog, displayed only for continuous eval rows    
                                                                                                                                              
  Notes                                                                                                                                     
                                                                                                                                              
  - continuous_eval_name is already returned by the backend on AgenticAnnotationResponse — this is a frontend-only change with no DB or API modifications


notthattal (approved):
lgtm

PR #1569, opened 2026-04-10
[no-ref] handle unknown model cost gracefully

mcgrawia:
Previously if a user used a model that LiteLLM didn't know the cost for, the API call for generation would fail:

```
  File "/vagrant/genai-engine/src/clients/llm/llm_client.py", line 224, in completion
    cost_float = completion_cost(response)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/cost_calculator.py", line 1622, in completion_cost
    raise e
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/cost_calculator.py", line 1615, in completion_cost
    raise e
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/cost_calculator.py", line 1494, in completion_cost
    ) = cost_per_token(
        ^^^^^^^^^^^^^^^
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/cost_calculator.py", line 506, in cost_per_token
    return openai_cost_per_token(
           ^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/llms/openai/cost_calculation.py", line 35, in cost_per_token
    return generic_cost_per_token(
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/litellm_core_utils/llm_cost_calc/utils.py", line 640, in generic_cost_per_token
    model_info = get_model_info(model=model, custom_llm_provider=custom_llm_provider)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/utils.py", line 5979, in get_model_info
    _model_info = _get_model_info_helper(
                  ^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/vagrant/.venv/lib/python3.12/site-packages/litellm/utils.py", line 5892, in _get_model_info_helper
    raise Exception(
Exception: This model isn't mapped yet. model=gpt-5.4-mini-2026-03-17, custom_llm_provider=openai. Add it here - https://github.com/BerriAI/litellm/blob/main/model_prices_and_context_window.json.
```


notthattal:
I think there is a missed reference on this line that would cause this to fail: https://github.com/arthur-ai/arthur-engine/blob/dev/genai-engine/src/services/prompt/chat_completion_service.py#L398. AgenticPromptRunResponse doesn't make cost optional

mcgrawia:
hmm good find @notthattal. how is this working today and passing mypy? The cost could be null just a few lines above it: https://github.com/arthur-ai/arthur-engine/blob/8711d4d15111241e1e3348a35b9df1d58df18b7a/genai-engine/src/services/prompt/chat_completion_service.py#L376-376

Maybe mypy is smart enough to know `completion_cost` never returns null, so the cost will never be None in this case?

Looks like the default cost is set to "0.00" already here, I can change it to that

mcgrawia:
@notthattal just pushed a fix, let me know what you think

notthattal:
> hmm good find @notthattal. how is this working today and passing mypy? The cost could be null just a few lines above it: https://github.com/arthur-ai/arthur-engine/blob/8711d4d15111241e1e3348a35b9df1d58df18b7a/genai-engine/src/services/prompt/chat_completion_service.py#L376-376
> 
> Maybe mypy is smart enough to know `completion_cost` never returns null, so the cost will never be None in this case?
> 
> Looks like the default cost is set to "0.00" already here, I can change it to that



> hmm good find @notthattal. how is this working today and passing mypy? The cost could be null just a few lines above it: https://github.com/arthur-ai/arthur-engine/blob/8711d4d15111241e1e3348a35b9df1d58df18b7a/genai-engine/src/services/prompt/chat_completion_service.py#L376-376
> 
> Maybe mypy is smart enough to know `completion_cost` never returns null, so the cost will never be None in this case?
> 
> Looks like the default cost is set to "0.00" already here, I can change it to that

yeah it looks like that might be the case I inspected the function it shouldn't ever return None. I think the fix makes sense though

notthattal (dismissed):

notthattal (approved):

PR #1575, opened 2026-04-13
Track violation count per alert rule in compliance metrics

videetparekh:
## Summary
- Update `policy_alert_rule_check_count` metric to emit the actual number of alert violations per rule instead of a hardcoded `1.0`
- Add violation counting in `_check_alert_rules` — counts all alerts per rule in the 24h window instead of just keeping the first
- Add better logging to alert rule fetching flow (model ID, assignment ID, page counts, rule IDs)

## Test plan
- [x] All 13 unit tests pass (including 2 new: multiple violations, zero violations)
- [x] Linted with `scripts/lint.sh` (black, isort, autoflake, mypy)
- [x] Verified against live scope-v4.dev environment — ml-engine picks up jobs and writes 5 metrics correctly


ml-engine/pyproject.toml
```diff
@@ -61,8 +61,6 @@ linters = [
 ]
 
 [tool.uv]
-exclude-newer = "3 days"
-exclude-newer-package = { arthur-client = false }
```
mcgrawia, line 65: @videetparekh this should still be in the file

videetparekh: Done!

mcgrawia (dismissed):

mcgrawia (approved):

PR #1587, opened 2026-04-15
Automate release PR creation and main→dev sync (#release-process)

jdbarthur:
- Add create-release-pr.yml: workflow_dispatch to create dev→main PR with the correct trigger title; guards against same-version and duplicate PRs
- Add sync step to version-workflow.yml: merges origin/main before each version bump so dev stays in sync after production squash merges


jdbarthur:
no idea what happened to the history on this branch. creating a new one.

PR #1588, opened 2026-04-15
[no-ref] update traces table to allow selecting rows to test with an eval

mcgrawia:
# Dependencies

Blocked on the changes to the shared-packages library here: https://gitlab.com/ArthurAI/unify-frontend/-/merge_requests/1263

# Demo

https://github.com/user-attachments/assets/a76d6994-640a-4af5-91f6-f5f132b26625


genai-engine/ui/src/components/traces/components/tables/TraceLevel.tsx
```diff
@@ -199,24 +230,41 @@ export const TraceLevel = memo(({ welcomeDismissed }: TraceLevelProps) => {
...
+      {testRunConfig && (
+        <TestRunDialog
+          open={!!testRunConfig}
+          onClose={handleTestRunClose}
+          evalId={testRunConfig.evalId}
+          evalName={testRunConfig.evalName}
+          taskId={task?.id ?? ""}
+          initialTraceIds={selectedTraceIds.slice(0, 50)}
+        />
+      )}
```
xV3rt: I think we technically don't need the `open` prop if we are conditionally rendering here. We could either always render the Dialog and control it through the prop, or use conditional rendering to display it at the right times.

genai-engine/ui/src/components/traces/components/tables/components/SelectionActionBar.tsx
```diff
@@ -0,0 +1,41 @@
+import { Close, PlayArrow } from "@mui/icons-material";
+import { Button, Chip, Collapse, Paper, Stack, Typography } from "@mui/material";
+
+const MAX_TRACES = 50;
```
xV3rt: Could we move that to constants? We also use that variable in `<TestRunDialog />`. We could import it from a common source in both files

genai-engine/ui/src/components/live-evals/components/TestRunDialog.tsx
```diff
@@ -68,6 +71,22 @@ export const TestRunDialog = ({ open, onClose, evalId, evalName, taskId }: Props
...
+        .mutateAsync(unique)
+        .then((result) => {
+          setTestRunId(result.id);
+        })
+        .catch(() => {
+          autoStarted.current = false;
+        });
+    }
+  }, [open, initialTraceIds, retryCount]); // eslint-disable-line react-hooks/exhaustive-deps
+
```
xV3rt: I think we can skip this effect entirely. The mutation is really triggered by the user clicking "Run" in the eval picker, so it belongs in that click handler, not in an effect watching for initialTraceIds to appear. If we hoist useCreateTestRun up to TraceLevel and only pass testRunId down once it exists, the autoStarted ref, the retryCount state, and the exhaustive-deps disable all fall out naturally. They seem to only be here because the parent allocates a fresh initialTraceIds array every render and we need to make sure the effect wont rerun.

```
// TestRunDialog.tsx — props shrink to just identity + optional testRunId
type Props = {
  open: boolean;
  onClose: () => void;
  evalId: string;
  evalName: string;
  taskId: string;
  testRunId?: string; // when present, skip setup and go straight to monitoring
};
// no initialTraceIds, no autoStarted ref, no retryCount, no useEffect

// TraceLevel.tsx — the mutation fires in the click handler
const createTestRun = useCreateTestRun(); // { evalId, traceIds } passed at mutate time

const handleEvalSelected = async (eval_: { id: string; name: string }) => {
  setEvalPickerOpen(false);
  const unique = [...new Set(selectedTraceIds)].slice(0, MAX_TRACES_PER_TEST_RUN);
  if (unique.length === 0) return;
  try {
    const result = await createTestRun.mutateAsync({ evalId: eval_.id, traceIds: unique });
    setTestRunConfig({ evalId: eval_.id, evalName: eval_.name, testRunId: result.id });
  } catch {
    // error snackbar is already handled inside useCreateTestRun
  }
};
```

genai-engine/ui/src/components/traces/components/tables/TraceLevel.tsx
```diff
@@ -93,6 +99,31 @@ export const TraceLevel = memo(({ welcomeDismissed }: TraceLevelProps) => {
     queryFn: () => getFilteredTraces(api, params),
   });
 
+  const selectedTraceIds = useMemo(() => Object.keys(rowSelection).filter((k) => rowSelection[k]), [rowSelection]);
+
+  // Clear selection when pagination, filters, sorting, or time range change
+  useEffect(() => {
+    setRowSelection({});
+  }, [pagination.pageIndex, pagination.pageSize, filters, timeRange, sorting]);
```
xV3rt: I don't think we need an effect here, we can also move row selection clearance to an event handler. 

For sorting and pagination we own the setters locally, so wrapping them to also call setRowSelection({}) keeps cause and effect in the same event. 

For filters and timeRange we can't touch the setters (they're on a shared zustand store), but the React docs have a pattern for exactly this case: ["adjusting some state when a prop changes"](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes), which stores the previous value and compares during render. That also avoids the extra render cycle an effect-based reset would cause.

```
// Local state we own — wrap the setters so the event itself clears selection
const handleSortingChange = useCallback<OnChangeFn<SortingState>>((updater) => {
  setSorting(updater);
  setRowSelection({});
}, []);

const handlePaginationChange = useCallback<OnChangeFn<PaginationState>>(
  (updater) => {
    props.onPaginationChange(updater);
    setRowSelection({});
  },
  [props]
);

// External zustand state we can't reach
const [prevFilters, setPrevFilters] = useState(filters);
const [prevTimeRange, setPrevTimeRange] = useState(timeRange);
if (prevFilters !== filters || prevTimeRange !== timeRange) {
  setPrevFilters(filters);
  setPrevTimeRange(timeRange);
  setRowSelection({});
}

// useEffect(() => setRowSelection({}), [...]) — gone
```

genai-engine/ui/src/components/traces/components/tables/components/EvalPickerDialog.tsx
```diff
@@ -0,0 +1,189 @@
...
+      if (trimmed) {
+        setFilters([{ name: "name", operator: TextOperators.CONTAINS, value: trimmed }]);
+      } else {
+        setFilters([]);
+      }
+      pagination.resetPage();
+    }, DEBOUNCE_MS);
+
+    return () => clearTimeout(debounceTimer.current);
+  }, [searchInput]); // eslint-disable-line react-hooks/exhaustive-deps
```
xV3rt: I think we can use the existing useDebouncedValue hook instead of hand-rolling this with setTimeout + ref. Then filters can become a useMemo derived from the debounced value rather than a second piece of state. The resetPage() call can move into the input's onChange handler since we want that to happen as soon as the user starts typing anyway, not after the debounce.

```
const [searchInput, setSearchInput] = useState("");
const debouncedSearch = useDebouncedValue(searchInput, 300);

// filters derived
const filters = useMemo(() => {
  const trimmed = debouncedSearch.trim();
  return trimmed ? [{ name: "name", operator: TextOperators.CONTAINS, value: trimmed }] : [];
}, [debouncedSearch]);

// page reset happens in the event
const handleSearchChange = (value: string) => {
  setSearchInput(value);
  pagination.resetPage();
};

// <TextField ... onChange={(e) => handleSearchChange(e.target.value)} />

// the setTimeout/ref/useEffect/exhaustive-deps-disable block goes away
```

genai-engine/ui/src/components/traces/components/tables/components/SelectionActionBar.tsx
xV3rt: I think we also want the same value here, not 100% sure, but 50 is hardcoded atm

```
<TestRunDialog
    open={!!testRunConfig}
    onClose={handleTestRunClose}
    evalId={testRunConfig.evalId}
    evalName={testRunConfig.evalName}
    taskId={task?.id ?? ""}
    initialTraceIds={selectedTraceIds.slice(0, 50)}
/>
```

genai-engine/ui/src/components/traces/components/tables/TraceLevel.tsx
mcgrawia: ok these should all be addressed in my latest commits, let me know 

https://github.com/arthur-ai/arthur-engine/pull/1588/changes/1b902240af3cbb42aaebd789e5f48c7f163c5568

https://github.com/arthur-ai/arthur-engine/pull/1588/changes/1cc09b545a8daad031dfbfa3e623797fbff11d70

xV3rt (approved):
Looks great now! 🚀

PR #1599, opened 2026-04-17
Add guardrails to customer support agent demo

snpranav:
Added PII / Toxicity guardrails + option to simulate a violation to the customer support agent demo.

This is how the traces look like when `ARTHUR_GUARDRAILS_ENABLED=true` and `SIMULATE_INCORRECT_RESPONSE=false`
Screenshot 2026-04-17 at 11 10 52 AM


genai-engine/examples/agents/customer-support-agent/.env.example
```diff
@@ -15,6 +15,11 @@ GITHUB_TOKEN=your-github-token-here
 # Logging Configuration (optional)
 LOG_LEVEL=info
 
+# Guardrails Configuration
+ARTHUR_GUARDRAILS_ENABLED=false
```
fryz, line 19: would like this to default true

genai-engine/examples/agents/customer-support-agent/.env.example
```diff
@@ -15,6 +15,11 @@ GITHUB_TOKEN=your-github-token-here
...
+# Simulate Incorrect Response to test the guardrail implementation
+SIMULATE_INCORRECT_RESPONSE=false
```
fryz, line 21: think we can leave this out entirely, but you can leave it in either way (i probably won't use this for the stuff im demoing)

PR #1606, opened 2026-04-23
UP-3149 Follow-up

martin-arthur-ai:
## Summary
Closes the observability gap in the UP-3149 trace retention service and makes the run interval configurable for high-volume customers.

## Changes

**Observability (always-on):**

- Replaced the conditional if total_deleted > 0 log with an unconditional Trace retention run complete: deleted N traces (cutoff=..., retention=N days) on every success path, including zero-deletion runs.
- Added Next trace retention run scheduled at <ISO> after every enqueue (initial and subsequent).
- Standby log now includes ; next leadership check at <ISO> so non-leader replicas also surface liveness.
- Added a startup line Trace retention interval: N hour(s) confirming the resolved env-var value.

**Configurable interval:**

- New env var TRACE_RETENTION_INTERVAL_HOURS (integer hours). Default 24, minimum 1. Sub-minimum values are clamped up, non-integer values fall back to the default, both with a WARNING.
- The existing ONE_DAY_SECONDS module constant is preserved unchanged and continues to serve as the default.
- Read once at service construction (via a small _resolve_interval_seconds() helper), so changing the value requires a process restart — matches the convention used by other *_ENV_VARs in this repo.
**Docs:**
- genai-engine/src/services/README.md gains a TRACE_RETENTION_INTERVAL_HOURS entry under Configuration and a new Observability table listing every log string operators can grep for.

**Tests:**

- Added caplog-based tests for the new run-complete log (zero-deletion and non-zero-deletion), the next-run-scheduled log, the standby log suffix, and the four env-var branches (unset / valid override / sub-minimum clamp / invalid fallback).


genai-engine/src/services/trace_retention_service.py
```diff
@@ -136,12 +168,13 @@ def _execute_job(self, job: TraceRetentionJob) -> None:
...
-                        total_deleted,
-                        cutoff.isoformat(),
-                    )
+                logger.info(
+                    "Trace retention run complete: deleted %d traces "
+                    "(cutoff=%s, retention=%d days)",
+                    total_deleted,
+                    cutoff.isoformat(),
+                    config.trace_retention_days,
+                )
```
mcgrawia, line 178: thanks for adding this. I think the changes look good but we should also add a log line at the start of this function so we know when the job starts and if it somehow got stuck before reaching the end

martin-arthur-ai: Add a log for when the job starts

mcgrawia (approved):

PR #1607, opened 2026-04-23
feat(ml-engine): bake compliance policy check into alert check job

videetparekh:
## Summary

- Compliance policy check now runs at the tail of every `ALERT_CHECK` chain as a queued job, sharing the alert window. Replaces the old fixed 24h scheduler.
- Per-rule verdict is a **sliding 2-interval window**: query `(window_end - 2*rule.interval, window_end]` and report non-compliant iff any alert lands there. Independent of `window_start` — the verdict is a fixed-point latest-bucket check, not a scan of the user's lookback. Fixes the bug where stale alerts in a long lookback could pin a rule as non-compliant.
- Alert filter switched from `created_at_*` to `time_*` (alert bucket time, not insertion time) so backfills attribute correctly.
- `CompliancePolicyCheckJobSpec` carries `check_range_start/end_timestamp` and `policy_assignment_id`; `execute()` raises if windows are missing.
- Deleted `schedule_compliance_jobs_executor.py` and its dispatch case.
- ml-engine now stamps the assignment's chain progress: after spawning AlertCheck the metrics-calc executor PATCHes `alerts_check_job_id`, after spawning CompliancePolicyCheck the alert-check executor PATCHes `compliance_job_id`. Skipped for model-wide schedule runs (no `policy_assignment_id`). Powers the new per-stage chain status widget in the FE.

## Why 2 intervals, not 1

Alert timestamps land at fixed bucket boundaries; `window_end` is arbitrary. Looking back exactly 1 interval misses the latest completed bucket whenever `window_end` is mid-bucket (worst case: the latest bucket boundary is almost 2 intervals before `window_end`). 2 intervals always catches it, and at most one bucket fits in a 2-interval window, so we can't accidentally pick up a stale older bucket either.

## Metric change: `policy_alert_rule_check_count`

The `value` field becomes a **0/1 violation flag** instead of a full-window alert count. No `status` dimension added.

**Why 0/1, not a status dimension:** existing dashboard SQL reads `value` directly for the alert branch. Adding a status dim would have required hunting down and updating every consumer. Value-as-flag keeps the SQL contract identical — `SUM(value)` still equals "checks where this rule was violating at this time slot."

**Why drop the count:** the verdict now reflects only the latest bucket, so a stored full-window count tells a different story than the verdict. The count was also already unstable upstream of this change (window varies with cron cadence after the scope MR shipped). 0/1 stabilizes it and aligns with the verdict.

**Trade-off:** chart shape changes from continuous density to discrete violation events.

## Metrics now use `triggereing_alert.timestamp` as the key

`policy_alert_rule_check_count` is a `0/1` violation flag that dashboards `SUM` over time to chart per-rule violations. Today it's keyed to `_align_to_5min(now)` — the time the compliance check ran — not to the alert that caused the violation.

That means re-running compliance against the same persistent non-compliant alert produces a fresh metric row each time, with a different timestamp on each run. Five consecutive runs over the same ongoing alert get summed as "5 violations" on the dashboard when reality is one violation that simply hasn't cleared.

### The fix

For violation samples, the metric is now keyed to the alert, not the check run:
  - `timestamp = _align_to_5min(triggering_alert.timestamp)`
  - new `alert_id` dimension carrying `triggering_alert.id`


mcgrawia (dismissed):

ml-engine/src/ml_engine/job_executors/_chain_utils.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def _list_assignment_ids_for_model(
+    policies_client: PoliciesV1Api,
+    model_id: str,
+) -> List[str]:
+    ids: List[str] = []
+    page = 1
+    while True:
+        resp = policies_client.list_model_policy_assignments(
```
mcgrawia, line 49: just to confirm, the engine's role allows listing policy assignments on the model right?

mcgrawia (approved):

ml-engine/src/ml_engine/job_executors/_chain_utils.py
videetparekh: Yeah, DATA_PLANE_PROJECT_ROLE has `PermissionName.MODEL_LIST_POLICY_ASSIGNMENTS,`

PR #1632, opened 2026-05-08
Add engine configuration endpoint and demo mode feature flag

fryz:
## Summary
This PR introduces a new engine configuration system that exposes engine-level settings to the frontend via a public API endpoint. It includes a demo mode feature flag that can be controlled via environment variable.

## Key Changes

**Backend:**
- Added `GENAI_ENGINE_DEMO_MODE` environment variable support in `config.py` with case-insensitive "ENABLED" check
- Created new `/api/v2/engine-config` public endpoint (no authentication required) that returns engine configuration
- Added `EngineConfigResponse` schema with `demo_mode` boolean field
- Registered the new `engine_config_routes` router in all app initialization paths (production, test, and dev)

**Frontend:**
- Created `useEngineConfig()` hook to fetch engine configuration from the API with React Query
- Added `EngineConfigContext` and `EngineConfigProvider` for sharing demo mode state across the application
- Exported `useDemoMode()` hook for consuming the demo mode value in components
- Wrapped the app with `EngineConfigProvider` in `App.tsx`

## Implementation Details

- The demo mode feature flag defaults to `false` when the environment variable is not set or has any value other than "ENABLED"
- The endpoint is marked as public using the `@public_endpoint` decorator, allowing access without authentication
- The React Query hook uses `staleTime: Infinity` to cache the configuration indefinitely since it's unlikely to change during runtime
- The context provides both the `demoMode` boolean and `isLoading` state for UI components to handle loading states appropriately

https://claude.ai/code/session_012vccdYu323PLsUVScA5jKy


.claude/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Write|Edit",
+        "hooks": [
+          {
+            "type": "command",
+            "command": "{ read -r json; f=$(echo \"$json\" | jq -r '.tool_input.file_path // .tool_response.filePath // empty'); if echo \"$f\" | grep -qE 'genai-engine/ui/src/.*\\.(ts|tsx|js|jsx)$'; then cd /home/user/arthur-engine/genai-engine/ui && node_modules/.bin/eslint --fix \"$f\" 2>/dev/null; node_modules/.bin/prettier --write \"$f\" 2>/dev/null; fi; } 2>/dev/null || true"
```
fryz: this makes the linter run every time Claude touches FE code

videetparekh (dismissed):

genai-engine/ui/src/App.tsx
```diff
@@ -58,123 +59,125 @@ function App() {
             >
               <>
                 <ReactQueryDevtools />
-                <AuthProvider>
```
fryz, line 61: idk what the heck this is all about, but looks like this was a linter making a whitespace/formatting change?

genai-engine/uv.lock
xV3rt, line 1: Are those changes intended? Seems like `uv` was ran using a newer version than previously?

.claude/settings.json
```diff
@@ -0,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
+      {
+        "matcher": "Write|Edit",
```
xV3rt: I think we are missing `MultiEdit` here which is also used to make many edits within a single file at once

.claude/settings.json
xV3rt: the path here should probably come from `$CLAUDE_PROJECT_DIR` variable instead of being hardcoded. We aren't handling errors in any way, so if a path won't exist, the hook will never actually run

.claude/settings.json
```diff
@@ -0,0 +1,15 @@
...
+        "hooks": [
+          {
+            "type": "command",
+            "command": "{ read -r json; f=$(echo \"$json\" | jq -r '.tool_input.file_path // .tool_response.filePath // empty'); if echo \"$f\" | grep -qE 'genai-engine/ui/src/.*\\.(ts|tsx|js|jsx)$'; then cd /home/user/arthur-engine/genai-engine/ui && node_modules/.bin/eslint --fix \"$f\" 2>/dev/null; node_modules/.bin/prettier --write \"$f\" 2>/dev/null; fi; } 2>/dev/null || true"
+          }
+        ]
+      }
+    ]
+  }
+}
```
xV3rt, line 15: Now that I think of it, the approach of running linting after every write/edit might not be the best, as during tasks that include many file edits it will actually add quite a lot of latency. Wouldn't it be more efficient to run it once on "stop" after all the edits are done? Was thinking about something like this (also with `--cache`)
```
{
  "hooks": {
    "Stop": [
      {
        "hooks": [
          {
            "type": "command",
            "command": "cd \"$CLAUDE_PROJECT_DIR/genai-engine/ui\" && { yarn lint --fix --cache --cache-location node_modules/.cache/eslint && yarn format; } 1>&2 || exit 2"
          }
        ]
      }
    ]
  }
}
```

genai-engine/uv.lock
fryz: Yes, this is a benign side effect of running `uv sync` with a newer version of uv locally. The changes are purely metadata — a `revision = 3` counter bump and `upload-time` fields added to each package entry. No package versions or hashes changed. Happy to revert the `uv.lock` commit if you'd prefer to keep it consistent with the version used in CI.

---
_Generated by [Claude Code](https://claude.ai/code)_

.claude/settings.json
fryz: this should be g2g now

xV3rt (approved):

PR #1633, opened 2026-05-11
Add a new validate endpoint that enables easy prompt injection checks

xV3rt:
Adds `POST /api/v2/validate` - a stateless endpoint for running Arthur's built-in guardrail checks against arbitrary prompt/response without creating a task, rule, or persisted inference.


genai-engine/src/routers/v2/builtin_validate_routes.py
```diff
@@ -0,0 +1,90 @@
...
+
+@builtin_validate_routes.post(
+    "/validate",
+    description=(
+        "Stateless validation of arbitrary text against Arthur's built-in checks "
+        "(prompt_injection, toxicity, pii). Does NOT persist results, does NOT "
+        "create an inference, and is task-less. Intended for ad-hoc checks such as "
+        "validating tool-call output for prompt injection.\n\n"
+        "Notes:\n"
+        "- Configurable checks (regex, keyword, sensitive_data, hallucination) are "
```
mcgrawia: couple of questions:

1. why does the hallucination rule require configuration? I thought you could just turn it on and use it?
2. I had thought part of the spec for this endpoint was that any guardrails that require configuration would allow the user to pass the config in the request? Is there a reason we can't do that so all guardrails are usable via this endpoint?

genai-engine/src/schemas/builtin_validate_schemas.py
```diff
@@ -0,0 +1,59 @@
...
+
+BUILTIN_CHECK_TO_RULE_TYPE: dict[BuiltinCheckName, RuleType] = {
+    BuiltinCheckName.PROMPT_INJECTION: RuleType.PROMPT_INJECTION,
+    BuiltinCheckName.TOXICITY: RuleType.TOXICITY,
+    BuiltinCheckName.PII: RuleType.PII_DATA,
+}
+
+
+class BuiltinValidationRequest(BaseModel):
+    text: str = Field(
```
mcgrawia: I believe the current endpoints accept `prompt`, `response`, and `context`, is there a reason we want to collapse that to just a single field? In order to support rules like hallucination, I think we'll need them separate?

xV3rt:
Refactored the endpoint to accept prompt/response/context and support all guardrails

genai-engine/src/routers/v2/builtin_validate_routes.py
```diff
@@ -0,0 +1,165 @@
...
+    ValidationRequest,
+)
+from schemas.rules_schema_utils import CONFIG_CHECKERS
+from scorer.score import ScorerClient
+from utils.users import permission_checker
+
+builtin_validate_routes = APIRouter(
+    prefix="/api/v2",
+    route_class=GenaiEngineRoute,
+)
```
fryz: Can we add these routes to the existing validate_routes codebase? Want to keep things in the same place so that when we go to do an eventual API migration we can deprecate all the things in one place.

genai-engine/src/routers/v2/builtin_validate_routes.py
```diff
@@ -0,0 +1,165 @@
...
+        if not request.response:
+            raise HTTPException(
+                status_code=400,
+                detail="Check 'hallucination' requires `response` to be provided.",
+            )
+        if not request.context:
+            raise HTTPException(
+                status_code=400,
+                detail="Check 'hallucination' requires `context` to be provided.",
+            )
```
fryz: fwiw, think we can get rid of this validation - in agentic world, the whole idea of prompt-/response-only checking doesn't make sense anymore

genai-engine/src/schemas/builtin_validate_schemas.py
```diff
@@ -0,0 +1,148 @@
...
+    free-string customer eval names scoped to a task.
+    """
+
+    PROMPT_INJECTION = "prompt_injection"
+    TOXICITY = "toxicity"
+    PII = "pii"
+    HALLUCINATION = "hallucination"
+    REGEX = "regex"
+    KEYWORD = "keyword"
+    SENSITIVE_DATA = "sensitive_data"
```
fryz: ideally, we wouldn't have to maintain a separate list of checks that were supported and could instead use/derive this list from elsewhere 

eg: I believe we have some similar checking/validation in the scorer module

genai-engine/src/schemas/builtin_validate_schemas.py
```diff
@@ -0,0 +1,148 @@
...
+
+BUILTIN_CHECK_TO_RULE_TYPE: dict[BuiltinCheckName, RuleType] = {
+    BuiltinCheckName.PROMPT_INJECTION: RuleType.PROMPT_INJECTION,
+    BuiltinCheckName.TOXICITY: RuleType.TOXICITY,
+    BuiltinCheckName.PII: RuleType.PII_DATA,
+    BuiltinCheckName.HALLUCINATION: RuleType.MODEL_HALLUCINATION_V2,
+    BuiltinCheckName.REGEX: RuleType.REGEX,
+    BuiltinCheckName.KEYWORD: RuleType.KEYWORD,
+    BuiltinCheckName.SENSITIVE_DATA: RuleType.MODEL_SENSITIVE_DATA,
+}
```
fryz: same here - would be great if we could re-use existing abstractions 

this will make it easier for us eventually when we deprecate this API and move to a new API

genai-engine/src/schemas/builtin_validate_schemas.py
```diff
@@ -0,0 +1,148 @@
...
+        default=None,
+        max_length=100_000,
+        description="LLM response to validate.",
+    )
+    context: Optional[str] = Field(
+        default=None,
+        max_length=1_000_000,
+        description="Grounding context for response-side checks (hallucination, sensitive_data).",
+    )
+    checks: List[BuiltinCheck] = Field(
```
fryz: ultimately, seems like this is the place where the seam exists in your implementation 

this is where i'd take a dependency on what already exists so that it couples your implementation to the old API which we'll hope to eventually swap out

genai-engine/src/routers/v2/builtin_validate_routes.py
```diff
@@ -0,0 +1,165 @@
...
+
+
+def _build_synthetic_rule(
```
fryz: see feedback from the schema module, but think if we re-use existing abstractions, we probably don't need this translation layer

fryz (changes requested):
LMK if you have questions

genai-engine/src/schemas/builtin_validate_schemas.py
xV3rt: Done - one caveat: the public type values are now PascalCase ("PromptInjectionRule", "ToxicityRule", etc.) instead of snake_case ("prompt_injection")

genai-engine/src/schemas/builtin_validate_schemas.py
xV3rt: Done, NewRuleRequest uses type: str + an untyped config: Optional[Union[...]] catch-all, so we lose the per-check Pydantic discrimination we had. A caller can send {"type": "RegexRule", "config": <ToxicityConfig fields>} and it'll parse, the mismatch only surfaces at CONFIG_CHECKERS dispatch. Might be worth to tighten NewRuleRequest to a discriminated union in `arthur-common` in the future

fryz (approved):
this lgtm :) thanks for accommodating the feedback

PR #1634, opened 2026-05-11
feat(UP-4385): Add search to tasks list + memoize TaskCard

TheLukaszNs-Blazity:
[JIRA Task](https://arthurai.atlassian.net/browse/UP-4385)

### DIFF
--- a/genai-engine/ui/src/components/AllTasks.tsx
+++ b/genai-engine/ui/src/components/AllTasks.tsx
@@ -22,41 +22,38 @@ import {
   Tooltip,
   Typography,
 } from "@mui/material";
+import { useQueryClient } from "@tanstack/react-query";
 import React, { useState, useEffect, useMemo, useCallback, useRef } from "react";
 import { useNavigate } from "react-router-dom";
 
 import { ArthurLogo } from "./common/ArthurLogo";
+import { SearchBar } from "./common/SearchBar";
 import { CreateTaskForm } from "./CreateTaskForm";
 import { TaskCard } from "./TaskCard";
 
 import { SettingsMenuButton } from "@/components/settings/SettingsMenuButton";
-import { useApi } from "@/hooks/useApi";
-import { TaskResponse } from "@/lib/api";
+import { useDebouncedValue } from "@/hooks/useDebouncedValue";
+import { useActiveTasksQuery, useArchivedTasksQuery } from "@/hooks/useTasksList";
+import { queryKeys } from "@/lib/queryKeys";
 import { type InactiveDays, type SortBy, useTaskListStore } from "@/stores/task-list.store";
-
-const PAGE_SIZE = 50;
 
 export const AllTasks: React.FC = () => {
   const navigate = useNavigate();
-  const api = useApi();
-  const [tasks, setTasks] = useState<TaskResponse[]>([]);
-  const [totalCount, setTotalCount] = useState(0);
-  const [currentPage, setCurrentPage] = useState(0);
-  const [isLoading, setIsLoading] = useState(true);
-  const [isFetchingMore, setIsFetchingMore] = useState(false);
-  const [error, setError] = useState<string | null>(null);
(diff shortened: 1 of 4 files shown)


genai-engine/ui/src/hooks/useTasksList.ts
```diff
@@ -0,0 +1,77 @@
...
+      const response = await api.searchTasksApiV2TasksSearchPost(
+        { page_size: ARCHIVED_PAGE_SIZE, page: pageParam as number },
+        { only_archived: true }
+      );
+      return response.data;
+    },
+    enabled: !!api && enabled,
+    initialPageParam: 0,
+    getNextPageParam: () => undefined,
+  });
```
xV3rt, line 72: Is it intended that we use `useInfiniteQuery()` here but `getNextPageParam()` function is a noop?

xV3rt (dismissed):

genai-engine/ui/src/components/AllTasks.tsx
```diff
@@ -111,6 +121,25 @@ export const AllTasks: React.FC = () => {
...
+          archivedFetchNextPage();
+        }
+      },
+      { root, threshold: 0.1 }
+    );
+
+    observer.observe(sentinel);
+    return () => observer.disconnect();
+  }, [archivedDialogOpen, archivedHasNextPage, archivedIsFetchingNextPage, isLoadingArchived, archivedIsError, archivedFetchNextPage]);
+
```
xV3rt, line 142: Considering this logic is largely duplicated (we need one effect for regular list, one for archived list), could we create a reusable hook that handles infinite scroll? We don't have one like that within the repo (unless you want to use external library).

Something like that:
```
import { useCallback, useRef, type RefObject } from "react";

type UseInfiniteScrollSentinelOptions = {
  hasNextPage: boolean;
  isFetchingNextPage: boolean;
  fetchNextPage: () => void;
  isLoading?: boolean;
  isError?: boolean;
  enabled?: boolean;
  root?: RefObject<Element | null> | null;
  rootMargin?: string;
  threshold?: number | number[];
};

export function useInfiniteScrollSentinel({
  hasNextPage,
  isFetchingNextPage,
  fetchNextPage,
  isLoading = false,
  isError = false,
  enabled = true,
  root = null,
  rootMargin,
  threshold = 0.1,
}: UseInfiniteScrollSentinelOptions) {
  const stateRef = useRef({ hasNextPage, isFetchingNextPage, fetchNextPage, isLoading, isError });
  stateRef.current = { hasNextPage, isFetchingNextPage, fetchNextPage, isLoading, isError };

  return useCallback(
    (node: HTMLElement | null) => {
      if (!node || !enabled) return;

      const observer = new IntersectionObserver(
        ([entry]) => {
          const s = stateRef.current;
          if (entry.isIntersecting && s.hasNextPage && !s.isFetchingNextPage && !s.isLoading && !s.isError) {
            s.fetchNextPage();
          }
        },
        { root: root?.current ?? null, rootMargin, threshold }
      );

      observer.observe(node);
      return () => observer.disconnect();
    },
    [enabled, root, rootMargin, threshold]
  );
}
```

And then we could use it in the following way:
```
const sentinelRef = useInfiniteScrollSentinel({
    hasNextPage,
    isFetchingNextPage,
    fetchNextPage,
    isLoading,
    isError,
  });

  const archivedSentinelRef = useInfiniteScrollSentinel({
    hasNextPage: archivedHasNextPage,
    isFetchingNextPage: archivedIsFetchingNextPage,
    fetchNextPage: archivedFetchNextPage,
    isLoading: isLoadingArchived,
    isError: archivedIsError,
    enabled: archivedDialogOpen,
    root: archivedScrollRef,
  });
```

I find it cleaner and its very likely we will need it in other pages in the future

xV3rt (approved):

genai-engine/ui/src/components/AllTasks.tsx
TheLukaszNs-Blazity: Let's work on this refactor later

PR #1646, opened 2026-05-14
add ability to create demo task

notthattal:
## Description
- Adds the ability to create a demo task if the FF is enabled
- Instantiates the demo task with:
        1. a prompt
        2. two CEs (one that passes and one that is intentionally incorrect to give a user the ability to fix it)
        3. An example dataset with some general knowledge question and answer pairs
        4. 3 Traces based on the dataset that also has the CE results
- Changes the way chatbots function to hold history on the FE and pass in the entire history to the BE instead of having an in-memory cache
- Creates the demo chatbot (disables the normal chatbot in the demo chatbot tab as it was obstructing the send/reset buttons)
- Adds the demo env var to the .env and cft templates

## FE Images
Screenshot 2026-05-21 at 11 30 48 AM
Screenshot 2026-05-21 at 11 31 26 AM
Screenshot 2026-05-21 at 11 31 43 AM


genai-engine/src/config/config.py
```diff
@@ -89,3 +89,11 @@ def audit_log_dir(cls) -> str:
...
 
         return os.path.join(os.path.dirname(__file__), "..", "..", "audit_logs")
+
+    @classmethod
+    def demo_mode_enabled(cls) -> bool:
+        demo_mode_enabled = get_env_var(
+            constants.GENAI_ENGINE_DEMO_MODE_ENABLED_ENV_VAR,
+            default="false",
+        )
+        return demo_mode_enabled.lower() == "true"
```
fryz: @notthattal - we already introduced this FF in https://github.com/arthur-ai/arthur-engine/pull/1632

notthattal: Looks like it wasn't added to the env

notthattal: But I will update

notthattal: Updated to use the existing one: https://github.com/arthur-ai/arthur-engine/pull/1646/changes/35a0dd5ef6dc66e0cc27ab0b47bc597347ccef1e

notthattal:
Fixed all CCR comments in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1646/changes/d22ec9ba9f4210378ba411b780a493064254a0b4)

> ## Code review
> Found 2 issues during review.
> 
> (1) Bug: broken Anthropic to OpenAI fallback in _get_model_provider_and_name -- see
> 
> https://github.com/arthur-ai/arthur-engine/blob/9d6907958d6d6e40e0ac6fb8788cba56aa55842f/genai-engine/src/repositories/demo_task_repository.py#L52-L66
> 
> ModelProviderRepository.get_model_provider_client never returns None; it either returns an LLMClient or raises HTTPException(status_code=400). So when Anthropic is not configured, the first call raises immediately and the OpenAI fallback branch is unreachable. The route catches the HTTPException, archives the partial task, and surfaces a 400 about Anthropic even on deployments where only OpenAI is configured. The is-not-None checks are dead code. Wrap each call in try/except HTTPException (or use a non-raising existence check) to implement the intended fallback.
> 
> (2) CLAUDE.md compliance: test does not clean up state it creates -- test_create_demo_task_archives_task_when_demo_items_raise -- see https://github.com/arthur-ai/arthur-engine/blob/9d6907958d6d6e40e0ac6fb8788cba56aa55842f/genai-engine/tests/unit/routes/tasks/test_task_crud.py#L251-L280
> 
> The route writes a real task via tasks_repo.create_task(...) before create_demo_items_for_task is invoked, so a DB row is created on every parametrize iteration. On the mocked exception the route calls archive_task(task.id), which only flips the archived flag -- the row remains. The test ends after asserting get_task returns 404 but never deletes the archived task. The sibling success test does call client.delete_task(task.id).
> 
> This leaves 3 orphan archived task rows per run. Per genai-engine/CLAUDE.md (https://github.com/arthur-ai/arthur-engine/blob/9d6907958d6d6e40e0ac6fb8788cba56aa55842f/genai-engine/CLAUDE.md):
> 
> > Make sure any unit tests are deterministic. They shouldnt rely on database ordering of results and should clean up any state they create
> 
> Suggested fix: after the assertions, call client.delete_task(captured_task_ids[0]) to remove the archived row.
> 
> Apologies for the earlier stray test comments, that was while probing tool permissions; the dedicated inline-comment tool was not loaded in this session.

genai-engine/src/routers/v1/demo_task_routes.py
```diff
@@ -0,0 +1,186 @@
...
+    except Exception as e:
+        tasks_repo.archive_task(task.id)
+        raise HTTPException(
+            status_code=500,
+            detail=f"Failed to create demo task: {e}",
+        )
+
+
+@demo_task_routes.post(
+    "/tasks/{task_id}/demos/chatbot/stream",
```
mcgrawia, line 25: @notthattal slightly confused by this, the MR description says the agent is still TODO? Has that been done now?

notthattal: ohp forgot to update the description. Yeah it's done now. Originally this was just going to be the setup for the demo task but since we wanted the real traces it ended up being combined

genai-engine/src/repositories/demo_task_repository.py
```diff
@@ -0,0 +1,428 @@
...
+
+        Prefers Anthropic, falls back to OpenAI. get_model_provider_client
+        raises an 400 err when a provider is not configured, so we
+        catch it to fall through to the next provider.
+
+        Raises a 400 err if neither provider is configured.
+        """
+        for provider, model_name in (
+            (ModelProvider.ANTHROPIC, "claude-sonnet-4-6"),
+            (ModelProvider.OPENAI, "gpt-4o"),
```
mcgrawia: should we use a newer model from open AI like gpt-5-mini?

genai-engine/src/services/chatbot/demo_chatbot_service.py
```diff
@@ -0,0 +1,191 @@
...
+from services.chatbot.base_chatbot_service import BaseChatbotService
+from services.prompt.chat_completion_service import ChatCompletionService
+from services.trace.internal_trace_service import TraceSpanBuilder
+from utils.sse_events import format_sse_json
+from utils.wikipedia_tools import wikipedia_fetch, wikipedia_search
+
+logger = logging.getLogger(__name__)
+
+
+DEMO_CONVERSATION_HISTORIES: MutableMapping[Tuple[str, str], List[OpenAIMessage]] = (
```
mcgrawia: @notthattal how is this expected to function? Since this is stored in memory it seems like there may be a case where a user's subsequent request is sent to a different replica than the first which wouldn't have the user's convo history?

Can users only have one convo at a time with the demo chatbot? 

I'm curious, why does this need to be stored on the BE anyway? Shouldn't the FE send the message history in each request to the BE?

genai-engine/src/server.py
```diff
@@ -614,6 +617,8 @@ def get_app() -> FastAPI:
     )
     if extra_feature_config.CHATBOT_ENABLED:
         add_routers(app, [chatbot_routes])
+    if Config.demo_mode():
+        add_routers(app, [demo_task_routes])
```
mcgrawia, line 624: one comment here, I notice you also added code in the endpoints to check the feature flag inside the endpoint. Since it looks like the endpoints are conditionally added based on the flag, those checks may never be hit because the router will return a 404 because it doesn't know about those paths

genai-engine/src/services/chatbot/demo_chatbot_service.py
mcgrawia: I'm also wondering, if the user enters some chats in the UI, then closes it and comes back later, does the UI re-load the history? I don't see an API endpoint that retrieves the history, so it seems the user will see an empty chat on the FE, but the BE will append the message to the previous chat history?

genai-engine/src/repositories/demo_task_repository.py
notthattal: yeah we can

genai-engine/src/services/chatbot/demo_chatbot_service.py
notthattal: Yeah I had it so a user can only have one conversation with the chatbot per task and if they want to have a new one they should just reset the convo. The reason it's stored on the backend is because after a `GENAI_ENGINE_CHATBOT_MAX_ITERATIONS` number of times the chat is summarized for better context management. But on the frontend we don't show that summary. I can add an endpoint to retrieve the history, but it would also show the summary when the user reloads it so it might be better to clear it when switching out of a task. 

I can also change it to not summarize for the demo. I guess it just depends how much we care about the token cost. I added the summarizer for the regular chatbot since that cost would be paid by the user, but, if we don't really care about it since we'd be fronting the cost, we could just skip that portion of it.

genai-engine/src/server.py
notthattal: I can remove the check inside the route

genai-engine/src/services/chatbot/demo_chatbot_service.py
mcgrawia: ok a couple of questions/comments/concerns with this approach:
1. why is the convo summarized after `GENAI_ENGINE_CHATBOT_MAX_ITERATIONS`? I thought that param limited the number of iterations in a single response? I would expect that the convo is summarized after a certain conversation length, not necessarily the number of iterations the agent took to response to a single request
2. if `GENAI_ENGINE_CHATBOT_MAX_ITERATIONS` controls the summarization, what happens if an agent response never hits that `GENAI_ENGINE_CHATBOT_MAX_ITERATIONS` limit while responding to a single request? It seems like the conversation will grow indefinitely? 
3. I think this needs some re-thinking from the user perspective. If this chat bot only supports a single conversation at a time, and the UI doesn't display the conversation history, but takes into account previous messages when responding, the users is going to be confused. From their perspective, they will think they're starting a blank chat but the BE has stored the history from their last chat
4. I believe there is a bug here if there's multiple replicas of the BE. The chat history will only store in memory of one replica, so each time a user sends a chat message, it may be added to the history of a different replica. That will cause odd responses because in the current implementation, the FE only sends the most recent message to the BE? So if the user sends two messages, the first message may route to one replica, then the second may route to another, and the agent would respond as if it had no idea about the first message because it was in the history of another replica.
5. Similar to 4, since the history is stored per replica, the clear history endpoint will only clear the history on whatever replica receives the request, which means there may be other history on other replicas.
6. If users have a long running chat session and the server restarts/updates, the agent will suddenly lose all memory of the conversation

genai-engine/src/server.py
mcgrawia: I think it's probably ok to leave it? Just in case someone removes the flag to add the endpoints?

Also, how is the FE aware of this feature flag? Won't it need to know if the flag is enabled to show/hide the create demo agent button?

mcgrawia: @TheLukaszNs-Blazity this might be a question for you^? How are you currently determining when to show/hide the sign up button in the UI?

genai-engine/src/services/chatbot/demo_chatbot_service.py
notthattal: I'm sorry I think I copied the wrong env var. I meant to use GENAI_ENGINE_CHATBOT_MAX_HISTORY_SIZE

notthattal: It sounds like you're suggesting we store the messages on the frontend? If so, do we still want to summarize? If we do summarize, would we want the summary displayed to the user instead of the previous chat history or keep a copy of both and only add the newest message to what's sent to the BE?

genai-engine/src/server.py
notthattal: I see this within `genai-engine/ui/src/contexts/EngineConfigContext.tsx`

```javascript
import { createContext, ReactNode, useContext } from "react";

import { useEngineConfig } from "@/hooks/useEngineConfig";

interface EngineConfigContextValue {
  demoMode: boolean;
  isLoading: boolean;
}

const EngineConfigContext = createContext<EngineConfigContextValue>({
  demoMode: false,
  isLoading: true,
});

export function EngineConfigProvider({ children }: { children: ReactNode }) {
  const config = useEngineConfig();
  return <EngineConfigContext.Provider value={config}>{children}</EngineConfigContext.Provider>;
}

export function useDemoMode(): EngineConfigContextValue {
  return useContext(EngineConfigContext);
}

```

genai-engine/src/repositories/demo_task_repository.py
notthattal: Updated in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1646/changes/810a72b9ed93c9b50d895bf21dd56391787f5e31)

genai-engine/src/services/chatbot/demo_chatbot_service.py
mcgrawia: Ah ok makes sense then on the `MAX_HISTORY_SIZE` being the summarization control 👍 

And I don't necessarily have a preference, I think they could be stored on the BE or FE. If they're stored on the BE they should be in the DB though.

Ultimately, I think it comes down to what the UX to the users should be. It we're giving them the appearance of an ephemeral chat that disappears as soon as they leave, storing and putting it on the FE is probably sufficient? Then the BE simply becomes a completion call? If we want to persist chats, we'll need to store them in the DB

notthattal: I think it's fine to keep it on the FE I don't think the chats necessarily need to persist. I guess I'm just not sure what to do about the summarization portion of it. I think skipping summarization for the demo chatbot might be okay. But for the regular chatbot that a user might need to iterate on multiple times do we store two copies of the chat on the FE the summarized one that gets sent back to the BE and the one that the user can see in the chat history as they scroll? Or, do we replace the previous history as they're talking with the summary?

genai-engine/src/server.py
mcgrawia: does the useEngineConfig call an API on the BE? Or how does that state get to the FE?

notthattal: yeah it looks like it calls `/api/v2/engine-config` which just returns this:

```python
return EngineConfigResponse(demo_mode=Config.demo_mode())
```

mcgrawia: thanks for confirming, I didn't know that endpoint existed. just wanted to make sure it was something all users (admin and tenant) can access

genai-engine/src/services/chatbot/demo_chatbot_service.py
mcgrawia: hmm, I think when Claude Code does compaction the chat history is replaced with a single summary message? Seems like that would be fine for now?

notthattal: okay sounds good. So I'll switch it to store on the FE but have the BE return the summary as well so we can update the chats with the summary

notthattal: Updated in these commits:
- https://github.com/arthur-ai/arthur-engine/pull/1646/changes/0fe130d70b2df3089fc890310326e719b25b26fe
- https://github.com/arthur-ai/arthur-engine/pull/1646/changes/247c24c081d5b3dc8bd8615298e96f4f44a2e7a5

deployment/cloudformation/genai-engine/arthur-genai-engine-ecs-task-definition-gpu.yml
ntatsumi, line 1: Need to expose the new param on the root templates.
* [root-arthur-engine-cpu.yml](https://github.com/arthur-ai/arthur-engine/blob/dev/deployment/cloudformation/root-arthur-engine-cpu.yml)
* [root-arthur-engine-gpu.yml](https://github.com/arthur-ai/arthur-engine/blob/dev/deployment/cloudformation/root-arthur-engine-gpu.yml)
* [root-arthur-genai-engine-cpu.yml](https://github.com/arthur-ai/arthur-engine/blob/dev/deployment/cloudformation/root-arthur-genai-engine-cpu.yml)
* [root-arthur-genai-engine-gpu.yml](https://github.com/arthur-ai/arthur-engine/blob/dev/deployment/cloudformation/root-arthur-genai-engine-gpu.yml)

notthattal: Updated in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1646/changes/3eea70ab24196fe1f1d6fcbec26797496c363149)

genai-engine/docs/audit-logging.md
```diff
@@ -0,0 +1,174 @@
+# Audit Logging
```
mcgrawia, line 1: @notthattal looks like some merge artifacts here or was this intentional?

genai-engine/docs/MULTI_TENANCY_DESIGN.md
```diff
@@ -231,7 +231,7 @@ POST /api/v2/tenant/signup       (no auth required)
 
 ### Feature flag
 
-The endpoint exists in every deployment but only responds when `GENAI_ENGINE_DEMO_MODE` is `true` in `src/utils/config.py`. Default: `false`. With the flag off, the endpoint returns 404 — clients cannot distinguish "feature unavailable" from "endpoint doesn't exist," which is the desired behavior on production customer deployments.
+The endpoint exists in every deployment but only responds when `ENABLE_PUBLIC_TENANT_SIGNUP` is `true` in `src/utils/config.py`. Default: `false`. With the flag off, the endpoint returns 404 — clients cannot distinguish "feature unavailable" from "endpoint doesn't exist," which is the desired behavior on production customer deployments.
```
mcgrawia: @notthattal same here looks like merge artifacts? I think this got reverted

genai-engine/src/routers/v2/tenant_signup_routes.py
```diff
@@ -104,15 +105,28 @@ def create_tenant_signup(
 
         api_key = ApiKeyRepository(db_session).create_api_key(
             description=f"demo signup for org {db_org.name}",
-            roles=[APIKeysRolesEnum.TENANT_USER],  # type: ignore[attr-defined]
+            roles=[APIKeysRolesEnum.TENANT_USER],
             org_id=db_org.id,
             commit=False,
         )
 
-        db_session.commit()
```
mcgrawia, line 112: @notthattal I'm not seeing a commit in this endpoint anymore. how does this work without it?

mcgrawia: it looks like there's actually multiple commits happening throughout the repo functions.. we should edit those to add optional commit flags so there's only one commit at the end and everything is committed or aborted together

genai-engine/docs/MULTI_TENANCY_DESIGN.md
notthattal: redid the rebase and this should be fixed

genai-engine/docs/audit-logging.md
notthattal: redid the rebase and this should be fixed

genai-engine/tests/clients/base_test_client.py
```diff
@@ -524,6 +524,51 @@ def create_task(
             ),
         )
 
+    def create_demo_task(self) -> tuple[int, TaskResponse]:
+        resp = self.base_client.post(
+            "/api/v1/tasks/demos",
```
videetparekh: This endpoint doesn't exist anymore

genai-engine/tests/clients/base_test_client.py
```diff
@@ -524,6 +524,51 @@ def create_task(
...
+                TaskResponse.model_validate(resp.json())
+                if resp.status_code == 200
+                else None
+            ),
+        )
+
+    def clear_demo_chatbot_history(self, task_id: str) -> tuple[int, str | None]:
+        resp = self.base_client.delete(
+            f"/api/v1/tasks/{task_id}/demos/chatbot/history",
+            headers=self.authorized_user_api_key_headers,
```
videetparekh: Same here I think you removed this?

genai-engine/src/routers/v2/tenant_signup_routes.py
notthattal: added in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1646/changes/de404062246ea2d12080f888b206744995bca8ef)

genai-engine/tests/clients/base_test_client.py
notthattal: yup noticed that too. I fixed it in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1646/changes/3fefddf44f09e51b0919c9c96ef94a36d069565c)

genai-engine/tests/clients/base_test_client.py
notthattal: same thing fixed in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1646/changes/3fefddf44f09e51b0919c9c96ef94a36d069565c)

videetparekh (approved):

PR #1648, opened 2026-05-14
UP-4388 multi-tenancy design doc

mcgrawia:


mcgrawia:
Thanks for the review. Addressed in 33e5f3eb. Status of each item:

**Issue 1 — `@enforce_task_scope` cannot scope body `task_id` — already addressed in revision 2.** The reviewed commit (11d81774) was the v1 of the doc, which had only one decorator. Revision 2 (6f1b3b30) introduced four enforcement patterns:
- Pattern A — `@enforce_task_scope` for path-task_id
- Pattern B — **`@enforce_body_task_scope`** for body-task_id (the case you flagged)
- Pattern C — repository-level filter for resource-id lookups
- Pattern D — `@enforce_query_task_scope` for `?task_ids=` query params

The fuzz coverage test was also expanded (lines 776-778 of the current revision) to assert each pattern: it inspects the handler's body model for a `task_id` field and asserts `@enforce_body_task_scope` is present. So the gap you described is closed, and the test catches future drift.

**Issue 2 — `tasks.is_autocreated` migration — column already exists.** Defined at `src/db_models/task_models.py:26` and added by migration `alembic/versions/2026_02_09_1744-843e2d3f46d5_service_name_map_and_default_task_.py:25`. The signup INSERT will succeed against the current schema. To remove ambiguity for future readers, I added an "Existing columns reused (no new migration needed)" subsection in Schema Changes that names the migration and confirms the column's shape.

**Issue 3 — `model_providers` enumeration leak — fixed.** You're right that per-provider `available_models` lets a tenant probe `anthropic`, `openai`, `azure`, etc. and infer the configured set from 200 vs 404, defeating the list endpoint's block. Both are now ❌ for tenant keys (Pattern E). The integration test list was extended to include `GET /api/v1/model_providers/{p}/available_models` for known provider IDs to catch regressions. Tenants who eventually need a "what models can my task use" view should get a separate endpoint that returns model names without revealing provider configuration — flagged inline in the doc.

mcgrawia:
Follow-up on the model_providers fix: blocking those endpoints leaves tenants with no way to discover models, so I added a "Tenant Model Access" section (47c1d312) enumerating three paths and a v1 recommendation.

**The three options, in short:**

- **Option A — Arthur hosts, sanitized `GET /api/v2/models` endpoint.** Tenants get a curated model list with no provider config. Arthur eats LLM costs. ~1-2 days.
- **Option B — BYOK with per-tenant model providers.** Tenants configure their own provider credentials per task. Big schema change (`task_id` on `model_providers` and `secret_storage`) and new CRUD endpoints. ~1-2 weeks.
- **Option C — Hybrid.** Ship A in v1; layer B on top in v2 without breaking the response shape.

**Recommendation in the doc: Option A for v1**, with C as the long-term path. Reasoning is in the section, but the short version: the demo env's target user is evaluating genai-engine, friction matters more than billing flexibility, and option A is forward-compatible.

**Three open questions for the team flagged in the doc:**
1. Is "Arthur eats LLM costs for demo tenants" acceptable financially? If not, jump straight to option C with BYOK as the v1 path for cost-sensitive deployments.
2. Do paying multi-tenant customers ever expect to BYOK? If yes, schedule option B sooner.
3. Should the v1 `/api/v2/models` response include `context_window` and other model capabilities, or stay minimal (just `id` + `display_name`)?

Doc updates also include the `MODELS_READ` permission frozenset, the new endpoint in Phase 2, an integration test for sanitization, and a future-work note in "What This Doc Is Not Doing" for BYOK.

mcgrawia:
Reversing course on the model_providers fix after reading the actual response schemas (5d04f216).

The read endpoints already return:
- \`ModelProviderResponse = {provider, enabled}\` (\`src/schemas/response_schemas.py:334\`)
- \`ModelProviderModelList = {provider, available_models}\` (\`src/schemas/response_schemas.py:347\`)

No credentials, no endpoints, no deployment names. FastAPI's \`response_model\` enforcement guarantees nothing else leaks. The list of supported \`ModelProvider\` values is already part of the public OpenAPI enum (it appears in path params and request bodies customers consume), so it's not a meaningful secret.

The bot review's "leaks infra" rationale was based on a wrong premise: there's no infrastructure config in the response payloads, just the public enum + a boolean. Blocking these endpoints was solving a problem that didn't exist, and it forced a new endpoint + UI rewrite for no real security gain.

**New v1 path: Option 0** — add \`TENANT-USER\` to the \`MODEL_PROVIDER_READ\` frozenset. That's it. Writes stay admin-only. Zero UI changes, no new endpoint, no new permission. The existing UI flow that calls \`GET /model_providers\` then \`GET /model_providers/{p}/available_models\` continues to work as-is.

Option A (new sanitized \`/api/v2/models\` endpoint) is retained in the doc as a fallback if a future deployment judges "tenants learn which providers are enabled on this deployment" as a meaningful leak. Option B (BYOK) and the hybrid Option C are unchanged, except C now layers BYOK on top of Option 0 instead of Option A — same endpoints, merged tenant+global responses in v2.

Open questions for the team are unchanged:
1. Is "tenants learn which providers a customer deployment has configured" acceptable? If not, fall back to Option A and budget the UI work.
2. Do paying multi-tenant customers expect to BYOK? If yes, schedule Option B sooner.
3. Is Arthur eating LLM costs for demo tenants acceptable financially? If not, consider Option B for v1 cost-sensitive deployments.

genai-engine/docs/MULTI_TENANCY_DESIGN.md
```diff
@@ -0,0 +1,1047 @@
...
+| 6 | Documents/embeddings = admin-only in v1 | Owner-scoped today, no clean tenant story without bigger schema work. Defer. |
+| 7 | Add a single new role: `TENANT-USER` | Existing role system already supports adding roles; the permission frozensets just need this role added where appropriate. |
+| 8 | `task_id` on `api_keys` is **nullable** | Null = cross-task admin key (existing behavior). Non-null = tenant-scoped key. Clean backwards-compat. |
+| 9 | Four enforcement patterns, not one | Audit revealed task is identified four different ways: path, body, resource-id (via FK), query param. Each needs its own mitigation. A single decorator can't cover all four. |
+| 10 | Body-`task_id` mismatch returns 404 | Consistent with path-mismatch behavior. Prevents enumeration. |
+| 11 | Trace upload: strict reject on `arthur.task` mismatch | Honest failure beats silent rewrite. Easier to debug client misconfigurations. |
+| 12 | Service-name → task mapping disabled for tenant keys | Implicit task routing is too easy to abuse. Tenant keys must send explicit `arthur.task`. Admin keys keep existing fallback. |
```
fryz: believe Tal's PR uses this

mcgrawia:
Pushed rev 4 (bffc8030) — major restructure around your feedback. Six items addressed.

**1. Organization is now the tenant boundary, not task.** An org can contain multiple tasks. A tenant API key carries `org_id`, and any task in that org is reachable. The single-task-per-tenant case is just an org with one task.

**2. Migrations** — four new ones:
- Migration 0: create `organizations` table + seed `default` + `system` orgs
- Migration 1: add `tasks.org_id` (NOT NULL after backfill)
- Migration 2: add `api_keys.org_id` (NULL = admin, non-NULL = tenant)
- Migration 3: denorm both `task_id` AND `org_id` onto rule-result / feedback / annotation tables

**3. `POST /api/v2/tasks` extended** — admin can optionally pass `org_id` (defaults to `default` org) and `create_tenant_key` (mints an org-scoped tenant key for the admin in the same response). Preserves backwards compat: existing admin clients that don't pass `org_id` keep working, their tasks land in the `default` org.

**4. Auth checks resolve task → org.** New `request.state.org_scope` (replaces `task_scope`). All four enforcement decorators were renamed (`@enforce_org_scope`, `@enforce_body_org_scope`, `@enforce_query_org_scope`) and now do task→org resolution before comparing to the caller's `org_scope`. Pattern C repositories filter on the denormalized `org_id` column on hot tables. Decorators cache the resolved task on `request.state` so handlers don't re-fetch.

**5. `POST /api/v1/traces` is admin-only in v1.** Cleanest fix per your feedback. Drops the rev-2/3 mitigation logic (strict reject on `arthur.task` mismatch + disable service-name fallback for tenants) since the endpoint is now blocked entirely for tenant keys. Tenant keys still read traces via the GET endpoints (Patterns C and D). `TRACES_WRITE` permission no longer includes `TENANT-USER`. Re-enabling for tenants is documented as v2 work in "What This Doc Is Not Doing."

**6. System tasks → `system` org.** Migration 1 backfill routes everything with `is_system_task=True` into the `system` org. A partial unique index on `organizations.is_system` enforces "at most one system org." Tenant keys can never have `org_id = system_org.id` (the signup flows always create new orgs with `is_system=false`), so they cannot resolve system tasks via any path. Admin keys see them via the existing cross-org behavior.

**7. Item 7 in your list was empty.** Flagged in the doc's "Open Questions / Risks" section so the team can fill it in.

**Other notable changes:**
- Multi-task-per-tenant moves from non-goal to default behavior (a tenant key naturally sees all tasks in its org)
- New admin endpoints under `POST/GET/DELETE /api/v2/organizations*`
- New tenant signup endpoint creates `(org, task, api_key)` in one transaction
- Estimate revised to ~10-14 days human / ~8-10 hours CC (up from rev 3 due to the task→org resolution layer)
- Architecture section has a new ASCII diagram of the org → task → resource hierarchy
- Tenant Model Access (Option 0) wording updated — same conclusion, BYOK in v2 still merges into the same `/api/v1/model_providers*` endpoints but now scoped by org instead of task

Doc length: 999 lines.

mcgrawia:
Pushed rev 5 (3b8ac347) — a full rewrite of the design doc as a clean forward-looking spec rather than an iteratively revised journal.

**Structural changes:**
- 14 numbered sections with an explicit table of contents at the top
- Removed the design-decisions changelog and all rejected alternatives — what's in the doc is what's being built
- Section 8 (Endpoint Policy) trimmed to "one example per pattern" plus a compact summary table, instead of the prior ~80-row breakdown

**Behavior changes since rev 4:**
- `POST /api/v2/tasks` is now open to tenants (creates a task in the caller's org). No request body changes — caller identity determines org routing.
- `DELETE /api/v2/tasks/{id}` and `POST /api/v2/tasks/{id}/unarchive` also open to tenants for their own org's tasks (Pattern A guards cross-org access).
- All admin org-management endpoints are removed. Provisioning is exclusively via `POST /api/v2/tenant/signup`. Admins inspect orgs via the DB if needed.
- No rate limiting in v1.
- Hot-table denormalization is `org_id` only (not also `task_id`). Rationale captured in section 13 Future Work — keeps write cost down without solving any v1 problem; adding `task_id` later is a one-column migration if a future workload needs it.
- New `GET /users/me` endpoint for UI identity introspection.

Doc lands at 915 lines (down from 1047 in rev 4) despite covering more decisions. The 14-section structure should be much easier to skim and navigate.

genai-engine/docs/MULTI_TENANCY_DESIGN.md
```diff
@@ -0,0 +1,915 @@
...
+
+`org_id` only is the minimum needed for tenant isolation. If a future workload demonstrates that fast per-task queries on these tables are a bottleneck, adding `task_id` is a single-column migration with the same shape as Migration 3.
+
+---
+
+## 14. Open Questions
+
+Items that need a team decision before or during implementation. None gate the design — they're choices best made at review or implementation time.
+
+1. **Should `org_id` appear in `GET /api/v2/tasks*` responses for tenant callers?** After Migration 1, every task carries `org_id` and the existing response models will pick it up automatically. A tenant's response will include their own org's id on every task. Not a cross-org leak (they only ever see their own org), but it's a new field in the response shape. Options: (a) include unconditionally — simplest; (b) include for admin only — keeps the tenant API surface minimal; (c) include as a nested `org: {id, name}` for readability. Default in this doc is (a).
```
notthattal: This seems like it would be a pretty minimal change and wouldn't hurt to include in the response

mcgrawia: yeah agreed, will update to include it

genai-engine/docs/MULTI_TENANCY_DESIGN.md
```diff
@@ -0,0 +1,915 @@
...
+The following tables today reach `task_id` via 1-3 hop joins. Multi-tenancy enforcement needs a direct filter on org membership, so we denormalize `org_id` onto each of them. We do **not** also denormalize `task_id`: existing per-task queries already work via single-hop joins to the parent tables, and adding both columns doubles write cost without solving a tenant-isolation problem (see Future Work for the full rationale).
+
+| Table | Today reaches `task_id` via | Backfill source |
+|---|---|---|
+| `inference_feedback` | `inferences.task_id` (1 hop) | `inferences.task_id → tasks.org_id` |
+| `prompt_rule_results` | `inference_prompts → inferences.task_id` (2 hop) | same chain |
+| `response_rule_results` | `inference_responses → inferences.task_id` (2 hop) | same chain |
+| `rule_result_details` | via prompt/response rule_result (3 hop) | same chain |
+| `agentic_annotations` | `trace_metadata.task_id` (1 hop) | `trace_metadata.task_id → tasks.org_id` |
+| `hallucination_claims`, `pii_entities`, `keyword_entities`, `regex_entities`, `toxicity_scores` | same as `rule_result_details` (3 hop) | same chain |
```
videetparekh, line 198: We should consider doing the same for the `trace_metadata` table. Users will click into a trace by `My Org -> Observe -> list_traces (from trace metadata) -> get_trace (from span table)`. Adding org ID to the trace metadata will cost one less join. The org ID is resolvable by the same task ID lookup we do today. 

Possibly also worth considering `spans`.

genai-engine/docs/MULTI_TENANCY_DESIGN.md
```diff
@@ -0,0 +1,915 @@
...
+
+The endpoint exists in every deployment but only responds when `ENABLE_PUBLIC_TENANT_SIGNUP` is `true` in `src/utils/config.py`. Default: `false`. With the flag off, the endpoint returns 404 — clients cannot distinguish "feature unavailable" from "endpoint doesn't exist," which is the desired behavior on production customer deployments.
+
+### Request and response
+
+The request body is empty. The endpoint generates all identifiers server-side.
+
+```
+POST /api/v2/tenant/signup
+Body: (none)
```
videetparekh, line 242: Curious about how we would capture prospects and leads? Don't we need some signup body for that? Are we sending everything to Hubspot or something from the FE?

genai-engine/docs/MULTI_TENANCY_DESIGN.md
```diff
@@ -0,0 +1,915 @@
...
+
+201 Created
+{
+  "org_id":    "uuid",
+  "task_id":   "uuid",
+  "task_name": "demo-a3f9b2c1",
```
videetparekh, line 248: Personalized org and task naming based on signup body

genai-engine/docs/MULTI_TENANCY_DESIGN.md
mcgrawia: Can you explain this more? 

`My Org -> Observe`

What is the my org page? For now I was expecting users to enter the same way to the Observe page via a single task

genai-engine/docs/MULTI_TENANCY_DESIGN.md
mcgrawia: Good question, I didn't focus on that part for this design doc but we might want to add it to the body here and store it in the DB. Not sure what the FE is thinking or if that will go directly to Hubspot

genai-engine/docs/MULTI_TENANCY_DESIGN.md
videetparekh: Aah you're right, this is already pre-filtered to a task. Ignore this comment.

videetparekh (approved):

PR #1655, opened 2026-05-18
add documentation for audit log system

notthattal:
## Description

Adds documentation for the audit log system

### DIFF
--- a/genai-engine/docs/audit-logging.md
+++ b/genai-engine/docs/audit-logging.md
@@ -1,0 +1,174 @@
+# Audit Logging
+
+Arthur GenAI Engine includes built-in audit logging that records every authenticated API request as a structured JSON event. Logs are written to rotating files on disk.
+
+## Enabling Audit Logging
+
+Audit logging is **enabled by default**. Control it with these environment variables:
+
+| Variable | Default | Description |
+|---|---|---|
+| `AUDIT_LOG_ENABLED` | `"true"` | Set to `"false"` to disable audit logging entirely |
+| `AUDIT_LOG_RETENTION_DAYS` | `365` | Number of daily log files to retain before rotation deletes them |
+| `AUDIT_LOG_OVERRIDE_PATH` | *(unset)* | Custom directory for audit log files. Defaults to `<project-root>/audit_logs/` |
+
+When enabled, the server automatically:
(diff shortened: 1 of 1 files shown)


genai-engine/docs/audit-logging.md
```diff
@@ -0,0 +1,174 @@
...
+Arthur GenAI Engine includes built-in audit logging that records every authenticated API request as a structured JSON event. Logs are written to rotating files on disk.
+
+## Enabling Audit Logging
+
+Audit logging is **enabled by default**. Control it with these environment variables:
+
+| Variable | Default | Description |
+|---|---|---|
+| `AUDIT_LOG_ENABLED` | `"true"` | Set to `"false"` to disable audit logging entirely |
+| `AUDIT_LOG_RETENTION_DAYS` | `365` | Number of daily log files to retain before rotation deletes them |
```
mcgrawia, line 12: just thinking this feels quite high for a default. 365 days of high volume API requests could be huge files. Would 14 or 30 days make more sense?

genai-engine/docs/audit-logging.md
```diff
@@ -0,0 +1,174 @@
...
+2. Writes one JSON object per line to `audit.log`, rotating daily
+
+### Skipped Endpoints
+
+The following paths are **not** audit-logged:
+
+- `/health`
+- `/docs`, `/redoc`, `/openapi.json`
+- `**/tasks/{id}/chatbot/stream` (streaming responses)
+- `**/completions`
```
mcgrawia, line 26: why don't we log completions?

genai-engine/docs/audit-logging.md
mcgrawia: ah looks like below each day is written to a separate file? I think that's probably fine then

mcgrawia (approved):

genai-engine/docs/audit-logging.md
notthattal: yup it rotates on a daily schedule

genai-engine/docs/audit-logging.md
notthattal: we don't log completions for the same reason we don't log chatbot/stream which is that it can return a streaming response

PR #1656, opened 2026-05-19
feat(UP-4411): persist onboarding try-it-out submissions with session tokens

TheLukaszNs-Blazity:
Add onboarding_sessions and onboarding_submissions tables, public v2 API for issuing single-use tokens and submitting form data, plus UI wiring with session hook and submit flow.


genai-engine/src/routers/v2/onboarding_routes.py
```diff
@@ -0,0 +1,72 @@
...
+    session, plaintext_token = repository.create_session()
+    return OnboardingSessionResponse(
+        session_token=plaintext_token,
+        expires_at=session.expires_at,
+    )
+
+
+@onboarding_routes.post(
+    "/onboarding/submissions",
+    description="Submit try-it-out onboarding form data using a single-use session token.",
```
mcgrawia: repeating my comment from slack: https://arthur-ai.slack.com/archives/C095P231EEB/p1779210360737869?thread_ts=1779199451.942569&cid=C095P231EEB

@TheLukaszNs-Blazity curious, why do we need to introduce the separate onboarding sessions / tokens and separate endpoints? Can we make this simpler and just accept a JSON payload of data with the call to /signup that contains this form info?

mcgrawia: For reference, @notthattal 's MR creating the endpoint is here: https://github.com/arthur-ai/arthur-engine/pull/1646/changes#diff-9efbe130b1327af8da5299d4f2fec6551c160bfc5c99e93aa9016ce020b2b012R29

I was wondering if we could just extend the body of that request to accept and store the form info?

genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
```diff
@@ -0,0 +1,93 @@
+"""create organizations table
```
mcgrawia, line 1: @TheLukaszNs-Blazity it looks like there's a lot of duplicate changes in here, did this rebase correctly?

genai-engine/src/schemas/request_schemas.py
```diff
@@ -1387,6 +1387,33 @@ class SyntheticDataGenerationRequest(BaseModel):
...
+    email: str = Field(min_length=1)
+    job_title: str = Field(min_length=1)
+    company: str = Field(min_length=1)
+    maturity: str = Field(min_length=1)
+    brings: str = Field(min_length=1)
+    brings_other: str = ""
+    competitors: list[str] = Field(min_length=1)
+    competitor_other: str = ""
+    attribution: str = Field(min_length=1)
+    attribution_other: str = ""
```
mcgrawia: @TheLukaszNs-Blazity we should still use `Field()` values for these, but you can specify a `default=` value if they're not required.

If they're not required though, why not make them `Optional[str]` so they can be null/None?

genai-engine/src/schemas/request_schemas.py
```diff
@@ -1387,6 +1387,33 @@ class SyntheticDataGenerationRequest(BaseModel):
     )
 
 
+class OnboardingTryItOutFormData(BaseModel):
+    """Try-it-out onboarding form fields (matches UI TryItOutSubmission)."""
+
+    first_name: str = Field(min_length=1)
+    last_name: str = Field(min_length=1)
+    email: str = Field(min_length=1)
```
mcgrawia, line 1395: I might be remembering wrong, but I think there's an `EmailStr` type you can use that will validate it's a proper email address? Although the downside there is the request will be rejected if it's not a valid email, so not sure if it's better to accept invalid emails, or make the user specify a valid one?

genai-engine/tests/unit/routes/conftest.py
```diff
@@ -19,6 +21,22 @@
 from tests.clients.unit_test_client import get_genai_engine_test_client
 
 
+@pytest.fixture
+def tracked_onboarding_submissions() -> Generator[list[uuid.UUID], None, None]:
+    submission_ids: list[uuid.UUID] = []
```
mcgrawia, line 26: how does this work exactly? it seems like the list of submission_ids will always be empty, and this will never delete anything?

genai-engine/tests/unit/routes/test_tenant_signup.py
```diff
@@ -328,12 +357,61 @@ def create_organization(self, **kwargs):
...
+
+    try:
+        with patch.dict(os.environ, {"GENAI_ENGINE_DEMO_MODE": "ENABLED"}):
+            test_client = TestClient(app)
+            response = _signup(test_client)
+
+        assert response.status_code == 200
+
+        from db_models.onboarding_models import DatabaseOnboardingSubmission
+        from tests.clients.base_test_client import override_get_db_session
```
mcgrawia: these imports should go at the top of the file

genai-engine/src/schemas/request_schemas.py
TheLukaszNs-Blazity: We'd need to install `pydantic[email]`. Fwiw we already do email validation client-side, which should be enough in our case

genai-engine/tests/unit/routes/conftest.py
TheLukaszNs-Blazity: The fixture yields a mutable list object. The test appends the created submission ID to that same list, and pytest resumes the fixture teardown after the test, so the cleanup loop sees the appended IDs

genai-engine/src/schemas/request_schemas.py
mcgrawia: ok cool that works for me

genai-engine/tests/unit/routes/conftest.py
mcgrawia: Ah whoops I was thinking of Go's slice passing.. nvm, makes sense!

genai-engine/src/schemas/request_schemas.py
mcgrawia: @TheLukaszNs-Blazity is there a reason `brings_other` wasn't also updated?

PR #1659, opened 2026-05-19
UP-4435 Integrate engine onboarding flow with amplitude

xV3rt:


xV3rt:
Closing, we ended up implementing onboarding tour with an end-to-end custom solution

PR #1660, opened 2026-05-19
Feat/onboarding walkthrough

xV3rt:


xV3rt:
Closing, we actually implemented the tour using our own custom solution instead of `react-joyride`

PR #1664, opened 2026-05-20
feat(multitenancy): enforce org scope on routes (UP-4425)

mcgrawia:
## Summary

Implements API-level enforcement of organization scope for genai-engine routes. Tenant-scoped callers (`request.state.org_scope`, set by `multi_validator` from `api_key.org_id`) are blocked from touching another org's tasks at the path and query layers. Admin callers (`org_scope is None`) pass through unchanged.

## Decorators (`src/utils/users.py`)

- **`@enforce_org_scope()`** — Pattern A. Resolves the path's `{task_id}` to its `org_id` and 404s on mismatch.
- **`@enforce_query_org_scope()`** — Pattern D. Validates that supplied `task_ids` belong to the caller's org (403 on outside-org IDs); when absent, injects the full set of the caller's org's task IDs so downstream filters are transparent.

Both 404 (not 403) on miss/mismatch — `404` chosen to avoid cross-org enumeration oracles.

The `enforce_body_org_scope` (Pattern B) decorator was implemented but ultimately deleted in this PR — no current endpoint takes `task_id` in the request body (every `task_id`-using route puts it in the path).

## Coverage

- **Pattern A**: applied to all **60** path-`{task_id}` endpoints across 17 router files (task_management, validate, rag, rag_setting, agent_polling, agentic_experiment, agentic_notebook, agentic_prompt, chatbot, continuous_eval, llm_eval, notebook, prompt_experiment, rag_experiment, rag_notebook, transform, dataset_management).
- **Pattern D**: applied to all **11** query-`task_ids` endpoints (trace_api, legacy_span, query_routes, feedback, task_management/search).
- `get_validated_task` dependency was extended with an `org_scope` check so every route already using that dependency is covered automatically.

## Route-handler behavior changes

- `POST /api/v2/tasks` now routes by `org_scope`: admin → `DEFAULT_ORG_ID`, tenant → caller's org.
- `POST /api/v1/traces` remains admin-only (no change).

## Well-known org UUIDs

`DEFAULT_ORG_ID` and `SYSTEM_ORG_ID` live in `utils.constants` as the single source of truth (stable `…0001` / `…0002` UUIDs). All 4 multi-tenancy migrations and all app code reference them by import rather than duplicating UUID literals.

## Cross-dialect UUID columns

`sqlalchemy.Uuid(as_uuid=True)` replaces `postgresql.UUID(as_uuid=True)` on org_id columns so SQLite-backed unit tests round-trip UUIDs.

## Write-path `org_id` propagation

`org_id` is denormalized onto child rows by repositories at write time (needed to keep this PR's tests passing on top of UP-4424's NOT NULL constraints; overlaps with UP-4429's full Pattern C scope but is the minimum needed here):

- `InferenceRepository.save_prompt` / `save_response` stamp `org_id` onto prompt/response rule results and rule_details from the owning task.
- `FeedbackRepository.create_feedback` looks up `org_id` via `inference → task`.
- `ContinuousEvalsRepository`, `ContinuousEvalTestRunRepository`, `TraceAnnotationService` stamp `org_id` from the owning task on `agentic_annotations`.
- System task creators (`system_task_repository`, `services/system_tasks_service`, `global_agent_polling_service`, `tasks_repository.create_auto_task`) tag new tasks with `SYSTEM_ORG_ID` / `DEFAULT_ORG_ID` as appropriate.

The repeated `select(Task.org_id) ... or DEFAULT_ORG_ID` pattern is centralized in a single `lookup_org_id(session, query, default=...)` helper in `organizations_repository.py`. It uses `@overload` so the return type narrows to `UUID` when a non-None default is supplied.

## Review fixes folded in

- **Tenant isolation bypass on `inference_id`** — `query_inferences`'s `inference_id` shortcut now consults the decorator-rewritten `task_ids` (caller's org tasks) and treats a non-matching inference as "not found" for tenant callers.
- **404 enumeration oracle in `get_validated_task`** — both 404 paths (task absent vs. task in another org) now return identical `"Task not found"` bodies for tenant callers.
- **Inline imports** — moved out of function bodies except for one documented case in `utils.utils.seed_database` (now resolved by moving constants to `utils.constants`).

## Test plan

- [x] `uv run pytest -m unit_tests` — **1624 passed, 0 failed**
- [x] 51 new decorator unit tests cover path/query patterns + admin bypass + 404 on mismatch + permission_checker matrix for `TENANT-USER`
- [x] mypy: 0 errors (261 files)
- [x] routes_security_check: 0 unprotected (215 endpoints, 203 protected, 12 explicitly public)
- [ ] Integration tests on `UP-4387-multitenancy` after merge
- [ ] Manual smoke: hit `/api/v2/tasks/{id}` with a tenant key whose org owns the task → 200; with a different org → 404

## Scope of this PR (and what's not)

In scope: Patterns A, D, route-handler tweaks, well-known UUID constants, repository helpers for org_id stamping.

Out of scope (handled by sibling tickets):
- Pattern C (resource-id-scoped repo filtering for `{trace_id}`, `{span_id}`, `{inference_id}`, `{annotation_id}`, `{dataset_id}`, `{test_run_id}` paths) → **UP-4429**
- Full body-`task_id` decorator (Pattern B) → no endpoints need it today; decorator removed

### DIFF
--- a/genai-engine/alembic/versions/2026_05_19_1510-3abf881da275_denormalize_org_id_to_task_scoped_tables.py
+++ b/genai-engine/alembic/versions/2026_05_19_1510-3abf881da275_denormalize_org_id_to_task_scoped_tables.py
@@ -146,7 +146,7 @@ def upgrade() -> None:
         """)
     op.execute("""
         UPDATE agentic_annotations
-           SET org_id = (SELECT id FROM organizations WHERE name = 'system')
+           SET org_id = '00000000-0000-0000-0000-000000000002'  -- system org
          WHERE org_id IS NULL
         """)
 
--- a/genai-engine/alembic/versions/2026_05_19_1510-3abf881da275_denormalize_org_id_to_task_scoped_tables.py
+++ b/genai-engine/alembic/versions/2026_05_19_1510-3abf881da275_denormalize_org_id_to_task_scoped_tables.py
@@ -31,6 +31,7 @@ CK constraints on that table.
 import sqlalchemy as sa
 
 from alembic import op
+from utils.constants import SYSTEM_ORG_ID
 
 # revision identifiers, used by Alembic.
 revision = "3abf881da275"
@@ -144,11 +145,12 @@ def upgrade() -> None:
          WHERE aa.continuous_eval_id = ce.id
            AND aa.org_id IS NULL
         """)
-    op.execute("""
-        UPDATE agentic_annotations
-           SET org_id = '00000000-0000-0000-0000-000000000002'  -- system org
-         WHERE org_id IS NULL
-        """)
+    op.execute(
+        sa.text(
+            "UPDATE agentic_annotations SET org_id = :system_id "
+            "WHERE org_id IS NULL"
+        ).bindparams(system_id=str(SYSTEM_ORG_ID))
+    )
 
     # Step C: SET NOT NULL + index + FK on every table. If any row's parent
     # chain was broken (orphan data not covered by the fallbacks above),
--- a/genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
+++ b/genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
@@ -5,14 +5,18 @@ Revises: b3f7a2c1d9e0
 Create Date: 2026-05-19 15:10:05.190824
 
 Multi-tenancy step 1 of 5. Creates the `organizations` table that owns
-tasks and (transitively) every task-scoped resource. Seeds two rows:
+tasks and (transitively) every task-scoped resource. Seeds two rows
+with well-known UUIDs so application code references them by id rather
+than by name (name is likely to become user-editable in v2):
 
-  - `default` — the bucket for pre-multi-tenancy tasks and any admin-
-    created task without tenant context. Tenants never carry org_id = default.
-  - `system`  — internal tasks (is_system_task=True) live here. Tenants
-    never carry org_id = system.
+  - `default` — id `00000000-0000-0000-0000-000000000001`. The bucket
+    for pre-multi-tenancy tasks and any admin-created task without
+    tenant context. Tenants never carry org_id = default.
+  - `system`  — id `00000000-0000-0000-0000-000000000002`. Internal
+    tasks (is_system_task=True) live here. Tenants never carry
+    org_id = system.
 
-Subsequent migrations backfill `tasks.org_id` from these orgs.
+Subsequent migrations backfill `tasks.org_id` from these orgs by id.
 
 Partial unique index `uq_organizations_is_system_true` enforces "at most
 one row with is_system=TRUE."
@@ -66,9 +70,14 @@ def upgrade() -> None:
         postgresql_where=sa.text("is_system = TRUE"),
     )
 
+    # Seed with well-known UUIDs so application code can reference them
+    # by id without a name lookup. These ids are also asserted in
+    # `repositories/organizations_repository.py` and used by the
+    # backfill step in the next migration.
     op.execute(
-        "INSERT INTO organizations (name, is_system) VALUES "
-        "('default', FALSE), ('system', TRUE) "
+        "INSERT INTO organizations (id, name, is_system) VALUES "
+        "('00000000-0000-0000-0000-000000000001', 'default', FALSE), "
+        "('00000000-0000-0000-0000-000000000002', 'system',  TRUE) "
         "ON CONFLICT (name) DO NOTHING"
     )
 
--- a/genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
+++ b/genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
@@ -29,6 +29,7 @@ doesn't double-insert.
 import sqlalchemy as sa
 
 from alembic import op
+from utils.constants import DEFAULT_ORG_ID, SYSTEM_ORG_ID
 
 # revision identifiers, used by Alembic.
 revision = "514464b8ca3d"
@@ -71,14 +72,19 @@ def upgrade() -> None:
     )
 
     # Seed with well-known UUIDs so application code can reference them
-    # by id without a name lookup. These ids are also asserted in
-    # `repositories/organizations_repository.py` and used by the
-    # backfill step in the next migration.
+    # by id without a name lookup. UUIDs are sourced from `utils.constants`
+    # — the single source of truth shared with app code and the
+    # task / annotation backfill migrations below.
     op.execute(
-        "INSERT INTO organizations (id, name, is_system) VALUES "
-        "('00000000-0000-0000-0000-000000000001', 'default', FALSE), "
-        "('00000000-0000-0000-0000-000000000002', 'system',  TRUE) "
-        "ON CONFLICT (name) DO NOTHING"
+        sa.text(
+            "INSERT INTO organizations (id, name, is_system) VALUES "
+            "(:default_id, 'default', FALSE), "
+            "(:system_id,  'system',  TRUE) "
+            "ON CONFLICT (name) DO NOTHING"
+        ).bindparams(
+            default_id=str(DEFAULT_ORG_ID),
+            system_id=str(SYSTEM_ORG_ID),
+        )
     )
 
 
--- a/genai-engine/src/repositories/feedback_repository.py
+++ b/genai-engine/src/repositories/feedback_repository.py
@@ -10,8 +10,9 @@ from opentelemetry import trace
 from sqlalchemy import asc, desc
 from sqlalchemy.orm import Session
 
-from db_models import DatabaseInference, DatabaseInferenceFeedback
+from db_models import DatabaseInference, DatabaseInferenceFeedback, DatabaseTask
 from dependencies import get_db_session
+from repositories.organizations_repository import DEFAULT_ORG_ID
 
 logger = logging.getLogger()
 tracer = trace.get_tracer(__name__)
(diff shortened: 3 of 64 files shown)


mcgrawia:
Thanks for the review — pushed [894acbfa](https://github.com/arthur-ai/arthur-engine/pull/1664/commits/894acbfa) addressing all three.

**1) Tenant isolation bypass — `inference_id` shortcut**
Fixed in [`query_routes.py`](https://github.com/arthur-ai/arthur-engine/blob/894acbfa/genai-engine/src/routers/v2/query_routes.py). The `inference_id` branch now consults the decorator-rewritten `task_ids` (already constrained to the caller's org by `@enforce_query_org_scope`) and treats a non-matching inference as "not found" for tenant callers. Admin callers (`org_scope is None`) keep cross-org reads.

**2) 404 enumeration oracle in `get_validated_task`**
Fixed in [`dependencies.py`](https://github.com/arthur-ai/arthur-engine/blob/894acbfa/genai-engine/src/dependencies.py). For tenant callers, both 404 paths now return identical `"Task not found"` bodies. Admin callers still see the original (more informative) `"Task <UUID> not found."` from the repository.

**3) Imports inside function bodies**
Moved three of the four inline imports to module-top: `tasks_repository.py`, `tests/conftest.py`, `tests/unit/routes/inferences/conftest.py`.

`utils/utils.py::seed_database` is intentionally kept lazy with a comment: `db_models.base` imports `utils.utils` for `get_env_var`, so a module-top import of `repositories.organizations_repository` (which transitively imports `db_models`) creates a real circular import at module load. The lazy import is the documented exception.

Full unit suite green: 1627 passed.

genai-engine/alembic/versions/2026_05_19_1510-3abf881da275_denormalize_org_id_to_task_scoped_tables.py
```diff
@@ -146,7 +146,7 @@ def upgrade() -> None:
         """)
     op.execute("""
         UPDATE agentic_annotations
-           SET org_id = (SELECT id FROM organizations WHERE name = 'system')
+           SET org_id = '00000000-0000-0000-0000-000000000002'  -- system org
```
videetparekh: We can set these as values in `constants.py`, as we do with our task defaults. That way we have one UIUD source of truth referenced across all these migrations.

genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
```diff
@@ -66,9 +70,14 @@ def upgrade() -> None:
...
+    # Seed with well-known UUIDs so application code can reference them
+    # by id without a name lookup. These ids are also asserted in
+    # `repositories/organizations_repository.py` and used by the
+    # backfill step in the next migration.
     op.execute(
-        "INSERT INTO organizations (name, is_system) VALUES "
-        "('default', FALSE), ('system', TRUE) "
+        "INSERT INTO organizations (id, name, is_system) VALUES "
+        "('00000000-0000-0000-0000-000000000001', 'default', FALSE), "
+        "('00000000-0000-0000-0000-000000000002', 'system',  TRUE) "
```
mcgrawia: @videetparekh I noticed in the code we're looking up orgs by name, we should instead give these well-known orgs stable IDs so we can find them reliably if we want users to be able to name orgs in the future

genai-engine/src/repositories/feedback_repository.py
```diff
@@ -29,6 +30,14 @@ def create_feedback(
...
         user_id: str | None,
     ) -> DatabaseInferenceFeedback:
+        org_id = (
+            self.db_session.query(DatabaseTask.org_id)
+            .join(DatabaseInference, DatabaseInference.task_id == DatabaseTask.id)
+            .filter(DatabaseInference.id == inference_id)
+            .scalar()
+        )
+        if org_id is None:
+            org_id = DEFAULT_ORG_ID
```
videetparekh: This seems like a common usage pattern (also used in inference_repository.py). Do you think it makes sense to convert this into a helper that can be used downstream?
```
def get_org_id(
    session: Session,
    query: Select[tuple[UUID]],
    default: UUID | None = None,
) -> UUID | None:
    org_id = session.execute(query).scalar_one_or_none()
    return org_id if org_id is not None else default
```

Invoked by:
```
query = (
    select(DatabaseTask.org_id)
    .join(DatabaseInference, DatabaseInference.task_id == DatabaseTask.id)
    .where(DatabaseInference.id == inference_id)
)
org_id = get_org_id(self.db_session, query, default=DEFAULT_ORG_ID)
```

genai-engine/src/repositories/inference_repository.py
```diff
@@ -318,6 +358,18 @@ def save_prompt(
...
+
+        org_id = DEFAULT_ORG_ID
+        if task_id is not None:
+            task_org_id = (
+                self.db_session.query(DatabaseTask.org_id)
+                .filter(DatabaseTask.id == task_id)
+                .scalar()
+            )
+            if task_org_id is not None:
+                org_id = task_org_id
```
videetparekh: See comment above.

genai-engine/src/repositories/inference_repository.py
```diff
@@ -342,8 +394,18 @@ def save_response(
...
             db_inference = self.get_inference(inference_id)
+            org_id = DEFAULT_ORG_ID
+            if db_inference is not None and db_inference.task_id is not None:
+                task_org_id = (
+                    self.db_session.query(DatabaseTask.org_id)
+                    .filter(DatabaseTask.id == db_inference.task_id)
+                    .scalar()
+                )
+                if task_org_id is not None:
+                    org_id = task_org_id
```
videetparekh: See comment above.

genai-engine/src/repositories/organizations_repository.py
```diff
@@ -5,6 +5,13 @@
 
 from db_models import DatabaseOrganization
 
+# Well-known UUIDs for the seeded organizations. Mirrored in the
+# `create_organizations_table` migration's seed INSERT and in the
+# task / annotation backfills. Application code should reference these
+# constants (not the names) — names may eventually be user-editable.
+DEFAULT_ORG_ID: uuid.UUID = uuid.UUID("00000000-0000-0000-0000-000000000001")
+SYSTEM_ORG_ID: uuid.UUID = uuid.UUID("00000000-0000-0000-0000-000000000002")
```
videetparekh: This should be in constants.py

notthattal:
If we are requiring the decorators to be added to all task endpoints should we add a check similar to the `genai-engine-routes-security-check` to validate that any new routes added with a task endpoint are properly decorated?

mcgrawia:
> If we are requiring the decorators to be added to all task endpoints should we add a check similar to the `genai-engine-routes-security-check` to validate that any new routes added with a task endpoint are properly decorated?

@notthattal yup, that is coming next in ticket https://arthurai.atlassian.net/browse/UP-4432

genai-engine/alembic/versions/2026_05_19_1510-514464b8ca3d_create_organizations_table.py
videetparekh: Agreed with this.

mcgrawia:
@videetparekh thanks for the review — pushed [ba42c9cb](https://github.com/arthur-ai/arthur-engine/pull/1664/commits/ba42c9cb) addressing both points.

**1) Well-known org UUIDs → `utils.constants`**
`DEFAULT_ORG_ID` and `SYSTEM_ORG_ID` now live in [`genai-engine/src/utils/constants.py`](https://github.com/arthur-ai/arthur-engine/blob/ba42c9cb/genai-engine/src/utils/constants.py) as the single source of truth. The 4 migrations now `from utils.constants import …` and bind the values as SQL params (no more raw UUID literals embedded in the SQL strings). All app code and tests repointed.

**2) `lookup_org_id` helper**
Added in [`organizations_repository.py`](https://github.com/arthur-ai/arthur-engine/blob/ba42c9cb/genai-engine/src/repositories/organizations_repository.py). Five call sites refactored: `feedback_repository`, `inference_repository` (×2), `continuous_evals_repository`, `continuous_eval_test_run_repository`, `trace_annotation_service`.

I used `@overload` to give it precise typing — when a non-None `default` is supplied the return type narrows to `UUID` (not `Optional[UUID]`), so callers needing a guaranteed UUID for the `_stamp_org_id_on_*` helpers don't need extra asserts.

**Bonus** — dropped the lazy-import workaround in `utils.utils.seed_database`. The original circular path went `organizations_repository → db_models → utils.utils`; now that the constants live in `utils.constants` (sibling of `utils.utils`), a module-top import works cleanly.

mypy: 0 errors. routes_security_check: 0 unprotected. Unit tests: 1624 passing.

genai-engine/src/utils/users.py
```diff
@@ -66,3 +69,179 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+    org_id = db_session.execute(
+        select(DatabaseTask.org_id).where(DatabaseTask.id == str(task_id)),
+    ).scalar_one_or_none()
+    return None if org_id is None else str(org_id)
+
+
+async def _call(func: FunctionT, *args: Any, **kwargs: Any) -> Any:
+    if iscoroutinefunction(func):
+        return await func(*args, **kwargs)
+    return func(*args, **kwargs)
```
videetparekh, line 100: @claude What's the purpose of this `await _call(func, *args, **kwargs)` paradigm? Why do we invoke `_call(func,...)` instead `func(...)`?

genai-engine/src/utils/users.py
```diff
@@ -66,3 +69,179 @@ async def wrapper(*args: Any, **kwargs: Any) -> Any:
...
+        return kwargs, field, kwargs.get(field)
+
+    # Nested on a Pydantic model passed via Depends/Body
+    for value in kwargs.values():
+        if value is None:
+            continue
+        if isinstance(value, (str, bytes, int, float, bool, list, dict)):
+            # Avoid scanning primitive containers
+            continue
+        if hasattr(value, field):
```
videetparekh, line 172: Wouldn't injected Pydantic objects still be `BaseModel` types? If so, we can skip all the other checks and just look for any `BaseModel` instances.

videetparekh:
It seems like Pattern B and Pattern D were merged. Could you also update the design doc for this?

genai-engine/src/utils/users.py
mcgrawia: tbh I'm not sure every object here is guaranteed to by a pydantic object.. @claude do you know?

mcgrawia: @videetparekh I think I agree with claude's read here that if someone accidentally adds a non pydantic dependency in the future, we should leave this check agnostic so it still catches and enforced the tenancy. Thoughts?

videetparekh: Yeah, sounds good.

videetparekh (approved):

PR #1665, opened 2026-05-21
feat(multitenancy): public tenant signup endpoint (UP-4430)

videetparekh:
## Summary

Implements [UP-4430](https://arthurai.atlassian.net/browse/UP-4430): `POST /api/v2/tenant/signup` mints a fresh `(org, task, TENANT-USER api_key)` triple in a single transaction, gated by `GENAI_ENGINE_DEMO_MODE` (404 when disabled). Anonymous; the raw API key is the only place a tenant ever sees their token.

The demo *content* (rules fixtures, demo items, chatbot) is intentionally not created here — that lives on the in-flight #1646 (`feat/create-demo-task`) and will layer in when that branch merges.

## What's in here

**Endpoint** — `src/routers/v2/tenant_signup_routes.py`:
- `POST /api/v2/tenant/signup` via `@public_endpoint`; 404 when `GENAI_ENGINE_DEMO_MODE != ENABLED`.
- Single outer `try/except` wraps the whole flow: org create → task create → api_key mint → `db_session.commit()`. Any failure → `db_session.rollback()` + re-raise. The org-name retry loop has its own inner `except IntegrityError` because SQLAlchemy requires an explicit rollback between failed inserts before the next attempt can flush.
- Org name uses `demo-{secrets.token_hex(4)}`; collision → single retry with a fresh suffix; second failure → 500. The demo task reuses the org's `demo-<hex>` name so the two correlate by eye.

**Repository changes** (all backward-compatible, default `commit=True` keeps existing call sites untouched):
- `OrganizationsRepository.create_organization(name, is_system=False, commit=True)` — new.
- `tasks_repository.create_task(... commit=True)` — added kwarg.
- `api_key_repository.create_api_key(... commit=True)` — added kwarg.
- The signup route passes `commit=False` to all three for one atomic transaction.

**Org model portability** — `src/db_models/organization_models.py`:
- Python-side `default=` for `id`/`is_system`/`created_at` instead of Postgres-only `gen_random_uuid()`/`now()` server defaults. The Postgres migration still has the server-side defaults; ORM inserts never fall through to them.
- `sqlite_where` added alongside `postgresql_where` on the `is_system=TRUE` partial unique index. Without it SQLite treats the index as unconditional and refuses every `is_system=False` insert after the first.
- Needed to make the new tests run through the real test DB; UP-4422 missed these because nothing exercised real org INSERTs.

**Schema** — `src/schemas/response_schemas.py`:
- `DemoTaskSignupResponse {org_id: UUID, task_id: str, task_name: str, api_key: str}`.

**Design doc** — `docs/MULTI_TENANCY_DESIGN.md`:
- Literal swap of `ENABLE_PUBLIC_TENANT_SIGNUP` → `GENAI_ENGINE_DEMO_MODE`. No other doc changes.

**Spec + changelog** — regenerated `staging.openapi.json` + `src/api_changelog.md` (single line: `/api/v2/tenant/signup endpoint added`).

## Tests

`tests/unit/routes/test_tenant_signup.py` — 8 tests through the real `TestClient(app)` (same pattern as `test_engine_config_routes.py`):

**Behavioral (no mocking):**
- [x] `demo_mode` off → 404
- [x] `demo_mode` on → 200 with the four-field shape; `task_name` matches `demo-<8 hex>`
- [x] No `Authorization` header required
- [x] Returned bearer token authenticates against `/users/me` with `org_scope` populated to the new org and `TENANT-USER` in roles (end-to-end validation that the key is fully wired)
- [x] Each call mints a distinct `(org, task, key)` triple

**Error paths (monkeypatch the route's repository imports through TestClient):**
- [x] First `create_organization` raises `IntegrityError`; retry succeeds → 200
- [x] Both attempts raise `IntegrityError` → 500
- [x] `create_api_key` fails after org+task flushed → 500 and the count of `demo-*` orgs in the DB is unchanged (proves the rollback actually rolled the org back, not just that the handler returned 500)

## Acceptance criteria

- [x] Endpoint reachable without authentication
- [x] `GENAI_ENGINE_DEMO_MODE != ENABLED` → 404
- [x] `GENAI_ENGINE_DEMO_MODE = ENABLED` → 200 with four-field response shape
- [x] New org has `name="demo-{8-char-hex}"`, `is_system=False`
- [x] New task has `org_id` set to the new org
- [x] New API key has `roles=[TENANT-USER]`, `org_id = new_org.id`
- [x] Returned `api_key` is the only place the raw token appears; DB stores only the bcrypt hash
- [x] Transaction rolls back cleanly on any failure (verified by the no-orphan-org assertion)
- [x] Unique-name collision retries once; second failure → 500
- [x] Bearer token authenticates against `/users/me` with the correct `org_scope` and `TENANT-USER` role
- [ ] Cross-org isolation assertions on `GET /api/v2/tasks` — deferred to UP-4432 (those assertions belong to the test ticket, not the implementation ticket)
- [ ] Demo task content (rules fixtures, demo items, chatbot) — will come back when #1646 merges

## Test plan
- [x] `uv run pytest tests/unit/routes/test_tenant_signup.py` (8/8 pass)
- [x] Smoke `tests/unit/utils/` + `tests/unit/routes/test_engine_config_routes.py` + new file (202 pass, 0 fail)
- [x] `uv run generate_changelog` — picked up exactly the new endpoint
- [x] `uv run black` + `uv run isort` clean on touched files
- [x] `uv run mypy` clean on touched files
- [x] Rebased onto latest `UP-4387-multitenancy` (includes UP-4425)

## Reviewer notes

When #1646 (`feat/create-demo-task`) lands on `UP-4387-multitenancy`, reconcile the route module that holds the demo signup endpoint — we own `tenant_signup_routes.py` (path `/api/v2/tenant/signup`), they own `demo_task_routes.py` (path `/api/v1/tasks/demos` + a streaming demo chatbot). Different paths → no FastAPI route collision. The `DemoTaskRepository.create_demo_items_for_task` call from #1646 needs to thread between our task creation and api-key creation to add demo content to the new org's task.


genai-engine/tests/unit/utils/test_tenant_signup.py
```diff
@@ -0,0 +1,226 @@
...
+from fastapi import HTTPException
+from sqlalchemy.exc import IntegrityError
+
+from routers.v2.tenant_signup_routes import create_tenant_signup
+
+
+def _call(db_session: MagicMock, application_config: MagicMock = None):
+    # create_tenant_signup is a sync handler wrapped by @public_endpoint;
+    # __wrapped__ is the original sync function.
+    return create_tenant_signup.__wrapped__(
```
mcgrawia: is there a reason we can't use the test client to call this API instead of having to mock it?

videetparekh: Good catch — no, there isn't. The original reason was the `tests/unit/routes/` conftest fixture was erroring on `tasks.org_id NOT NULL` before UP-4423/UP-4425 landed, which forced the `__wrapped__` workaround. Now that UP-4425's merged into UP-4387-multitenancy, the conftest works again.

Rewrote the tests in 961c6b03 using `TestClient(app)` (same pattern as `test_engine_config_routes.py`). The new tests cover the whole HTTP path: route registration, FastAPI DI, real DB transaction. Added a cross-check that the minted bearer token actually authenticates as TENANT-USER against `/users/me` with org_scope populated — much stronger than the previous mock-based assertions.

Moved the file to `tests/unit/routes/` while I was at it (proper home for route tests).

The error-path tests (retry/double-collision/rollback) still go through the TestClient but monkeypatch the route module's repository imports — IntegrityError on a known-not-yet-taken random hex is hard to force through the real DB.

Two small fixes to `organization_models.py` were required to make this work for SQLite (UP-4422 missed them; never noticed because no test had exercised real org creation): Python-side defaults for `id`/`is_system`/`created_at` instead of Postgres-only server-side `gen_random_uuid()`/`now()`, and a matching `sqlite_where` clause on the `is_system=TRUE` partial unique index. Both are folded into the same commit.

8/8 tests pass.

videetparekh: ^ "Claude be temperamental"

mcgrawia (approved):

PR #1668, opened 2026-05-21
Initial Design for Adding Images into Evaluators

iabouhashish:


genai-engine/docs/Image_Eval_Design.md
```diff
@@ -0,0 +1,326 @@
...
+---
+
+## What Images Look Like in the Trace (Today)
+
+An OpenInference span carrying an image message arrives as:
+
+```
+attributes.llm.input_messages.0.message.role = "user"
+attributes.llm.input_messages.0.message.contents.0.message_content.type = "image"
+attributes.llm.input_messages.0.message.contents.0.message_content.image.url = "https://..."
```
fryz, line 20: would be worth confirming that images come in via url or base64-encoded content (or some other encoding)

genai-engine/docs/Image_Eval_Design.md
```diff
@@ -0,0 +1,326 @@
...
+    model_provider: ModelProvider
+    instructions: str
+    config: Optional[LLMRequestConfigSettings] = None
+    image_variables: list[str] = Field(
+        default_factory=list,
+        description="Variable names whose values are image URLs. These are injected "
+                    "as image_url content items in the LLM message, not interpolated "
+                    "into the text prompt.",
+    )
+```
```
fryz, line 110: Now looking at this option, not a big fan of using "Approach A" and handling the image variables separately. 

For the BE implementation (and for clients), it's not clear why `image_variables` are special/different from `text_variables`, which it looks like are extracted from the `instructions` variable (and not separately sent/managed like `image_variables`). 

We're implicitly extracting `text_variables` while being explicit about `image_variables`, which isn't clear from the API definitions (will only know this from reviewing the implementation). 

I worry that this tribal knowledge will introduce bugs later down the line.

genai-engine/docs/Image_Eval_Design.md
```diff
@@ -0,0 +1,326 @@
...
+)
+```
+
+Without this line, `LLMEval.image_variables` is always `[]` regardless of what's in the DB — a silent runtime bug.
+
+**Alembic migration** — one `NOT NULL` JSONB column with server default `'[]'`. Follow the existing `variables` column pattern in `llm_eval_models.py`: use `JSON` imported from `sqlalchemy.dialects.postgresql` (not `JSON().with_variant(...)`). Match the column definition exactly:
+
+### 2\. Variable pipeline — full trace
+
+The eval pipeline has several layers that all need to handle `image_variables` correctly:
```
fryz, line 147: Same thing here - we will be introducing special handling in numerous places for `image_variables`, separately from how we handle `text_variables`. 

Overall think it will lead to more complicated and messy code.

genai-engine/docs/Image_Eval_Design.md
```diff
@@ -0,0 +1,326 @@
...
+10. Eval result (pass/fail + reason) stored as DatabaseAgenticAnnotation
+```
+
+---
+
+## Open Questions
+
+1. **Base64 images**: If the client sends a base64 data URI instead of a URL (some OpenInference instrumentation does this), does LiteLLM pass it through correctly? Test this. OpenAI accepts `data:image/jpeg;base64,...` in `image_url.url`.
+
+2. **Image URL expiry**: If the image URL is a short-lived pre-signed URL and the continuous eval is delayed (jobs queue under load), the URL may expire before the LLM call. No solution for this in v1 — document it. If it becomes a real issue, the storage design from `MULTIMEDIA_DESIGN.md` (now deprecated) is the v2 path.
```
fryz, line 288: believe we should hold a requirement that we are NOT storing image content 

if URLs expire for the image then we should return `skipped` for the eval evaluation

fryz (changes requested):
Overall: 

1. We should get some more information about how evals will be passed in. Will the customer pass in image URLs, base64 encoded images (or use some other encoding format), or both? 
2. It would be good to look into what the implementation plan would look like for "Approach B" (using type annotation on the mappings). This will allow for us to build generalized abstractions for eval variables, meaning that we can push down handling to specific implementing classes instead of the proposed special-case handling. It'll lead to cleaner, more maintainable and understandable code. 


Finally, it would be good to understand how this will impact variable handling on prompts as well. I'm not 100% sure, but I believe that some of the variable handling in prompts (esp. in the FE code), is shared between evals + prompts.

genai-engine/docs/Image_Eval_Design.md
iabouhashish: Customer mentioned that they are flexible, They can pass both the url or the base64 enconded content.
The images will be stored in GCS.

mcgrawia: if the images arrive as base64 encoded content, does all this still work exactly the same? or would we need to process encoded images differently than URLs?

genai-engine/docs/Image_Eval_Design.md
mcgrawia: I think a big difference that might justify the separation is that images aren't location specific within the prompt like the text placeholders. They just accompany the text to the LLM as separate content items in the message. I was thinking a separate variable set was justified.

would it make more sense if we made this `media_variables` instead of just images to make it more generic? I could see this being an issue in the future for documents like PDFs, and possibly videos too

genai-engine/docs/Image_Eval_Design.md
```diff
@@ -0,0 +1,326 @@
...
+                    "as image_url content items in the LLM message, not interpolated "
+                    "into the text prompt.",
+    )
+```
+
+**`src/db_models/llm_eval_models.py` — `DatabaseLLMEval`:**
+
+```py
+image_variables: Mapped[List[str]] = mapped_column(
+    JSON().with_variant(postgresql.JSONB, "postgresql"),
```
mcgrawia, line 116: does this need to be a JSONB column? Or can this be a list of varchar column to constrain it?

PR #1669, opened 2026-05-21
feat(multitenancy): Pattern C — org_scope filtering on read paths (UP-4429)

mcgrawia:
## Summary

Pattern C (`org_scope` filtering on task-scoped repository read paths) for the multi-tenancy work, plus all the follow-up gaps the original audit flagged. Tenant callers now get **404** (same body for cross-org or missing) instead of leaking another org's resource. Admin callers (`org_scope is None`) pass through unchanged.

## Final coverage — every endpoint accounted for

Audit script (`/tmp/audit_openapi.py` in CI artifacts) cross-references the live OpenAPI spec on `engine.development.arthur.ai` (193 operations) against the codebase. **Zero unprotected endpoints.**

| Pattern | Endpoints | What it means |
|---|---|---|
| **A** (`@enforce_org_scope`, path task_id) | **60** | Path has `{task_id}`; decorator resolves it and 404s on cross-org |
| **D** (`@enforce_query_org_scope`, query task_ids) | **11** | Query has `task_ids` / `task_id`; decorator rewrites or 403s |
| **C** (`org_scope` to repo) — *this PR* | **92** | Repository methods accept `org_scope` and filter the SQL |
| **E** (admin-only frozenset) | **19** | `TENANT-USER` not in the permission set; 403 from `@permission_checker` |
| Public | 3 | `@public_endpoint` or `/health`, `/openapi.json`, etc. |
| Design-tenant-callable (no tenant data) | 8 | Global/registry reads + stateless body-provided ops; documented allowlist |
| **Unprotected** | **0** | — |

### 1. Pattern C — `org_scope` filtering on every task-scoped repository

Every `_get_db_X` / public lookup method on a task-scoped repository now accepts `org_scope: UUID | None = None`. Tenant callers' result sets get filtered (404 on cross-org or missing); admins pass through.

Repositories touched (full list in commits):
- `inference_repository`, `feedback_repository`, `span_repository`, `services/trace/trace_annotation_service`
- `datasets_repository`, `continuous_eval_test_run_repository`, `continuous_evals_repository`
- `rag_providers_repository` (incl. version + tag helpers)
- `agentic_experiment_repository`, `agentic_notebook_repository`, `prompt_experiment_repository`
- `rag_experiment_repository`, `rag_notebook_repository`, `notebook_repository`
- `trace_transform_repository`

Handlers wired across 17 router files. For span/trace endpoints I added `_trace_belongs_to_org` / `_span_belongs_to_org` private helpers — single `SELECT 1` ownership check at the entry point, cheaper than threading `org_scope` through three service layers.

### 2. New admin-only permission frozensets (purpose-named)

Added in `src/schemas/enums.py` — not reusing `APP_CONFIG_*` for things that aren't config:

```python
CHATBOT_CONFIG_READ   = frozenset([ORG_ADMIN, ORG_AUDITOR])
CHATBOT_CONFIG_WRITE  = frozenset([ORG_ADMIN])
AGENT_POLLING_ADMIN   = frozenset([ORG_ADMIN, TASK_ADMIN])
TELEMETRY_ADMIN_READ  = frozenset([ORG_ADMIN, ORG_AUDITOR, TASK_ADMIN])
DEFAULT_VALIDATION_RUN = frozenset([ORG_ADMIN, TASK_ADMIN])
```

Applied to:
- `GET/PUT /api/v1/chatbot/config` and `DELETE /api/v1/chatbot/history/{conversation_id}` → chatbot-specific perms (system task; never tenant data)
- `POST /api/v1/agent-polling/execute-all` → `AGENT_POLLING_ADMIN` (system-wide trigger)
- `GET /api/v1/traces/spans/unregistered` → `TELEMETRY_ADMIN_READ` (orphaned-spans debug view)
- `POST /api/v2/validate_prompt` (deprecated) → `DEFAULT_VALIDATION_RUN` (task-less default-rules run)

### 3. Org-list filtering on list/search endpoints

Repositories that return lists got an `org_scope` filter:

- `tasks_repository.query_tasks` / `get_all_tasks` → tenant sees only their org's tasks
- `rules_repository.query_rules` → tenant sees default-scope rules **plus** task-scope rules linked to their org's tasks
- `usage_repository.get_tokens_usage` / `run_tokens_query` / `get_token_query_statement` → filters on the denormalized `DatabasePromptRuleResult.org_id` / `DatabaseResponseRuleResult.org_id` columns added by UP-4424

Handlers wired:
- `GET /api/v2/tasks`
- `GET /api/v2/agent-tasks`
- `POST /api/v2/rules/search`
- `GET /api/v2/usage/tokens`

### 4. Pattern C for sessions

`span_repository.get_session_traces` and `compute_session_metrics` accept `org_scope`. Sessions span multiple tasks, so we filter per-trace (only keep traces whose owning task is in the caller's org) rather than session-wholesale.

- `GET /api/v1/traces/sessions/{session_id}`
- `GET /api/v1/traces/sessions/{session_id}/metrics`

### 5. Write-path corrections folded in

- **Task-less rule_results now land in `SYSTEM_ORG_ID`**, not `DEFAULT_ORG_ID`. Migration `d894934c8994` backfills orphan inferences to `UNMAPPED_TASK_ID` (system org); their rule_results should match. Fixes drift in `inference_repository.save_prompt` / `save_response`.
- **Deleted unused `AgenticAnnotation.to_db_model()`** — the Pydantic model had no `org_id` field; would have NOT NULL'd on use.

## Design-tenant-callable allowlist (the 8 remaining)

These 8 endpoints don't expose tenant data; the allowlist is documented in the audit script:

| Endpoint | Why it's safe |
|---|---|
| `GET /api/v2/default_rules` | Global defaults visible to everyone by design |
| `POST /api/v2/validate` | Stateless; body provides everything |
| `POST /api/v1/completions` | Unsaved-prompt body; nothing persisted |
| `POST /api/v1/prompt_renders` | Same |
| `POST /api/v1/prompt_variables` | Same |
| `GET /api/v2/datasets/synthetic-data/prompt-status` | Reads the system task's prompt; not tenant data |
| `GET /api/v1/model_providers` | Read-only model registry |
| `GET /api/v1/model_providers/{provider}/available_models` | Same |

## Test plan

- [x] mypy: 0 errors (261 files)
- [x] routes_security_check: 0 unprotected (215 endpoints)
- [x] OpenAPI audit: 0 unprotected (193 operations)
- [x] `uv run pytest -m unit_tests` — green
- [ ] Integration: tenant K1 hits each direct-id read/write with a T2-owned resource → 404 across the board

### DIFF
--- a/genai-engine/src/repositories/span_repository.py
+++ b/genai-engine/src/repositories/span_repository.py
@@ -18,6 +18,7 @@ from db_models import DatabaseSpan
 from db_models.task_models import DatabaseTask
 from db_models.telemetry_models import DatabaseTraceMetadata
 from repositories.metrics_repository import MetricRepository
+from repositories.organizations_repository import lookup_org_id
 from repositories.tasks_metrics_repository import TasksMetricsRepository
 from schemas.internal_schemas import (
     AgenticAnnotation,
@@ -40,6 +41,7 @@ from services.trace.trace_ingestion_service import TraceIngestionService
 from services.trace.tree_building_service import TreeBuildingService
 from utils import trace as trace_utils
 from utils.constants import (
+    DEFAULT_ORG_ID,
     EXPECTED_SPAN_VERSION,
     SPAN_VERSION_KEY,
 )
@@ -243,25 +245,28 @@ class SpanRepository:
         )
 
     def _trace_belongs_to_org(self, trace_id: str, org_scope: UUID) -> bool:
-        """Return True if the trace's task belongs to org_scope."""
+        """Return True if the trace belongs to org_scope.
+
+        Uses the denormalized `trace_metadata.org_id` column (Migration 4) so
+        this is a single-column lookup rather than a join through `tasks`.
+        """
         row = self.db_session.execute(
-            select(DatabaseTraceMetadata.trace_id)
-            .join(DatabaseTask, DatabaseTask.id == DatabaseTraceMetadata.task_id)
-            .where(
+            select(DatabaseTraceMetadata.trace_id).where(
                 DatabaseTraceMetadata.trace_id == trace_id,
-                DatabaseTask.org_id == str(org_scope),
+                DatabaseTraceMetadata.org_id == org_scope,
             )
         ).first()
         return row is not None
 
     def _span_belongs_to_org(self, span_id: str, org_scope: UUID) -> bool:
-        """Return True if the span's task belongs to org_scope."""
+        """Return True if the span belongs to org_scope.
+
+        Uses the denormalized `spans.org_id` column (Migration 4).
+        """
         row = self.db_session.execute(
-            select(DatabaseSpan.span_id)
-            .join(DatabaseTask, DatabaseTask.id == DatabaseSpan.task_id)
-            .where(
+            select(DatabaseSpan.span_id).where(
                 DatabaseSpan.span_id == span_id,
-                DatabaseTask.org_id == str(org_scope),
+                DatabaseSpan.org_id == org_scope,
             )
         ).first()
         return row is not None
@@ -383,17 +388,14 @@ class SpanRepository:
         # Tenant callers: filter the trace_ids down to only those whose owning
         # task lives in the caller's org. Sessions can span multiple tasks, so
         # we filter per-trace rather than checking the session wholesale.
+        # Uses the denormalized trace_metadata.org_id column (Migration 4).
         if org_scope is not None:
             owned = {
                 row[0]
                 for row in self.db_session.execute(
-                    select(DatabaseTraceMetadata.trace_id)
-                    .join(
-                        DatabaseTask, DatabaseTask.id == DatabaseTraceMetadata.task_id
-                    )
-                    .where(
+                    select(DatabaseTraceMetadata.trace_id).where(
                         DatabaseTraceMetadata.trace_id.in_(trace_ids),
-                        DatabaseTask.org_id == str(org_scope),
+                        DatabaseTraceMetadata.org_id == org_scope,
                     )
                 ).all()
             }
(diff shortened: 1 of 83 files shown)


genai-engine/src/schemas/internal_schemas.py
```diff
@@ -836,21 +836,6 @@ def from_db_model(db_annotation: DatabaseAgenticAnnotation) -> "AgenticAnnotatio
             updated_at=db_annotation.updated_at,
         )
 
-    def to_db_model(self) -> DatabaseAgenticAnnotation:
```
videetparekh, line 839: Was deleting this intentional?

genai-engine/src/repositories/span_repository.py
```diff
@@ -227,18 +239,47 @@ def compute_trace_metrics(
...
+        """Return True if the trace's task belongs to org_scope."""
+        row = self.db_session.execute(
+            select(DatabaseTraceMetadata.trace_id)
+            .join(DatabaseTask, DatabaseTask.id == DatabaseTraceMetadata.task_id)
+            .where(
+                DatabaseTraceMetadata.trace_id == trace_id,
+                DatabaseTask.org_id == str(org_scope),
+            )
+        ).first()
+        return row is not None
```
videetparekh, line 259: IMO it makes sense to add `org_id` to DatabaseTraceMetadata because it's a one time cost that would incur 1 less join every time someone accesses one or more traces, which is a common access pattern. The argument against denormalization is that every time the trace (or its underlying task) changes the org id we need to update it in the trace meta, but in practice a trace or task will never change its underlying org.

videetparekh: We may want to consider the same for spans as well.

genai-engine/src/schemas/internal_schemas.py
mcgrawia: yeah it wasn't used

genai-engine/src/repositories/span_repository.py
mcgrawia: ok I can add

mcgrawia: should be updated now

videetparekh (approved):

PR #1670, opened 2026-05-22
add alert rule logging

notthattal:
## Description

Adds the checks if an alert rule is okay/fired/has no data to be logged in the platform

## Partner MR

- https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1743

## Jira Tickets

- https://arthurai.atlassian.net/browse/UP-4277
- https://arthurai.atlassian.net/browse/UP-4278
- https://arthurai.atlassian.net/browse/UP-4371

### DIFF
--- a/ml-engine/src/ml_engine/job_executors/alert_check_executor.py
+++ b/ml-engine/src/ml_engine/job_executors/alert_check_executor.py
@@ -51,6 +51,12 @@ ALERT_RULES_NON_ADDITIONAL_DIMENSION_FIELDS = [
     "dimensions",
 ]
 ALERT_RULES_SCALAR_TYPES = (str, int, float, bool, datetime, UUID)
+
+# TimescaleDB's time_bucket anchors buckets to a fixed origin that depends on
+# the bucket width. Midnight Jan 3, 2000 for buckets using seconds/minutes/
+# hours/days and midnight Jan 1, 2000 for buckets using month/year/century.
+# AlertRuleInterval only supports seconds/minutes/hours/days, so it will
+# always be anchored against Jan 3, 2000.
 TIME_BUCKET_ORIGIN = datetime(2000, 1, 3, tzinfo=timezone.utc)
 
 
--- a/ml-engine/src/ml_engine/job_executors/alert_check_executor.py
+++ b/ml-engine/src/ml_engine/job_executors/alert_check_executor.py
@@ -263,8 +263,8 @@ class AlertCheckExecutor:
 
     def _crosses_threshold(self, alert_rule: AlertRule, value: float) -> bool:
         if alert_rule.bound == AlertBound.UPPER_BOUND:
-            return value > alert_rule.threshold
-        return value < alert_rule.threshold
+            return bool(value > alert_rule.threshold)
+        return bool(value < alert_rule.threshold)
 
     def _query_model_metrics(
         self,
--- a/ml-engine/src/ml_engine/job_executors/alert_check_executor.py
+++ b/ml-engine/src/ml_engine/job_executors/alert_check_executor.py
@@ -7,6 +7,7 @@ from arthur_client.api_bindings import (
     AlertBound,
     AlertCheckJobSpec,
     AlertRule,
+    AlertRuleLogStatus,
     AlertRulesV1Api,
     AlertsV1Api,
     CompliancePolicyCheckJobSpec,
@@ -18,6 +19,8 @@ from arthur_client.api_bindings import (
     PoliciesV1Api,
     PolicyAssignmentJobChainPatch,
     PostAlert,
+    PostAlertRuleLog,
+    PostAlertRuleLogs,
     PostAlerts,
     PostJob,
     PostJobBatch,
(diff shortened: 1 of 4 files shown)


ml-engine/src/ml_engine/job_executors/alert_check_executor.py
```diff
@@ -173,11 +176,38 @@ def _process_alert_rule(
...
+            self._post_alerts(job_spec.scope_model_id, alert_rule.id, alerts)
+        else:
+            self.logger.info("No alerts found!")
+
+        if len(alerts) > 0:
+            status = AlertRuleLogStatus.FIRED
+        elif self._has_data(job_spec, alert_rule):
+            status = AlertRuleLogStatus.OKAY
+        else:
+            status = AlertRuleLogStatus.NO_DATA
```
fryz: its not immediately clear to me on reading this, but it seems that `alerts` might be a collection, whereas what we're uploading for status should be tied to a single alert (and not a collection of alerts?) 

How does this tie back into our tech design w.r.t. being able to sequence alert status over time?

notthattal: yeah so right now it is checking if for an alert rule if there was any data to be evaluated in the check window, if any alerts fired for that alert rule or if it was okay. When we post the log in the platform we can then query on the status per model and alert rule at a specific time if it was failing/passing/no data

notthattal: so when we're sequencing it over time we would be able to see, for example: 

"policy alert 'prompt_injection_alert_rule' was failing for 'agent_x' all of last wednesday, stopped failing last thursday when we paused it and has received no data since"

fryz: ok - im still a bit confused here though 

Here's what im thinking - let me know if I'm off here: 
* Alert Status is tied to an Alert Rule - so we have a status tied to some check period (beginning/end time)
* Inside of a check period for a single Alert Rule, we may have multiple alerts that fired, were ok, or had no data
    * Day 1 - Alert fired
    * Day 2 - Alert was ok
    * Day 3 - Alert Fired
    * Day 4 - No Data

With what we've represented here, my understanding is that the Alert Rule would be in the FIRED state, but what I would expect to see on the graph in the UI is 4 blocks: red/green/red/grey 

Am I missing something w/ how this works?

notthattal: you're correct and i get what you're saying now. okay so I think I ned to change it so it's an alert log not an alert rule log and then post logs at the specified interval if there was data or it exceeded/didn't exceed the threshold at each interval

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
```diff
@@ -47,6 +51,23 @@
...
+) -> List[datetime]:
+    bucket_ts = adjusted_start_time - ((adjusted_start_time - TIME_BUCKET_ORIGIN) % td)
+    if bucket_ts < adjusted_start_time:
+        bucket_ts += td
+
+    buckets = []
+    while bucket_ts <= adjusted_end_time:
+        buckets.append(bucket_ts)
+        bucket_ts += td
+    return buckets
```
mcgrawia, line 114: @notthattal can you explain what this function is doing? it seems like it's getting the buckets between start/end time? What happens if the alert rule's bucket isn't a day?

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
```diff
@@ -171,36 +201,84 @@ def _process_alert_rule(
...
-        td = alert_interval_to_timedelta(alert_rule.interval)
-        adjusted_start_time = job_spec.check_range_start_timestamp - td
-
-        # similarly, to prevent alerting on partial alert buckets,
-        # this function post-filters the results to be in the range
-        # (start_time - interval, end_time - interval) so only the
-        # buckets that had the entire interval in the query are reported.
-        # This needs to be a post-filter so that all the data for the interval is
-        # included in the original time_bucket aggregation, then we filter
-        # on the post-aggregated results
```
mcgrawia, line 201: these comments about why the time range is adjusted are important and need to be preserved

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
mcgrawia: ah seeing below the time delta comes from the alert rule's interval so this should work for non-daily buckets. What is the whole origin part mod the time delta though?

I think a comment describing how this function works would be helpful

notthattal: added in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1670/changes/a883000768b1b07b114c0553b5fb2d739ab1c278)

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
notthattal: added back in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1670/changes/a883000768b1b07b114c0553b5fb2d739ab1c278)

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
notthattal: updated in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1670/changes/ee0e81613e8b366dd1b80b150e52eaa5cc32ff1c)

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
```diff
@@ -47,6 +51,39 @@
...
+    which ones had no data.
+
+    The origin adjustment exists because TimescaleDB's time_bucket buckets relative
+    to a fixed reference date (TIME_BUCKET_ORIGIN), not to our window's start. So, to
+    account for this, we snap adjusted_start_time to that same grid so our generated
+    timestamps match the ones the query returns, then create a bucket for each td
+    until our end time.
+
+    Timescale's docs on time_bucket:
+    https://www.tigerdata.com/docs/reference/timescaledb/hyperfunctions/time-series-utilities/time_bucket
```
mcgrawia, line 83: from the link above, it looks like the `time_bucket` function's origin is not fixed: 

> midnight on January 3, 2000 (for buckets < month) or midnight on January 1, 2000 (for month/year/century buckets)

mcgrawia: but thanks for the explanation, that makes sense 👍

mcgrawia (approved):

mcgrawia (changes requested):
whoops see comment above, didn't mean to approve yet

ml-engine/src/ml_engine/job_executors/alert_check_executor.py
notthattal: ah whoops didn't notice that. I'll update

notthattal: I added a comment above TIME_BUCKET_ORIGIN explaining why it's correct in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1670/changes/cd07241b19e99c632fad70a64ffb6ad4e69e2702). I was thinking I could also add something like 

```python
if interval in second, min, hour, day:
     origin = jan_3rd_origin
else:
     origin = jan_1st_origin
```
to guard against if we decide to add intervals >= month but wasn't sure if that was necessary. lmk what you think

mcgrawia: I think it might be worth adding the guard / fallback in case we support longer intervals in the future. If a dev adds "month" to the supported intervals, they won't realize they need to come back and update this as well to match. So I think having the backup is helpful to guard against future changes

notthattal: added in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1670/changes/a105a77c692ec8ca90e753086f02523430600b4b)

mcgrawia (approved):

PR #1684, opened 2026-05-29
feat(UP-4451): add reCAPTCHA Enterprise to onboarding signup form

TheLukaszNs-Blazity:
Gate the public tenant signup endpoint with score-based (invisible) reCAPTCHA Enterprise, covering both onboarding form variants (linear and wizard) via their shared submission path.

Backend:
- RecaptchaConfig + RecaptchaEnterpriseVerifier (Assessments REST API)
- verify the token in the tenant signup route (400 on failure)
- add optional recaptcha_token to TenantSignupRequest
- fail open when unconfigured so demo/dev flows are unaffected
- unit tests for the verifier and the signup gate

Frontend:
- lazy-loaded reCAPTCHA Enterprise script + token execution in the onboarding submission mutation
- badge shown only while the form is mounted
- map RECAPTCHA_ENTERPRISE_SITE_KEY to a VITE_-prefixed build var

Also regenerate the API changelog and OpenAPI spec.


deployment/docker-compose/genai-engine/.env.template
```diff
@@ -6,3 +6,14 @@ GENAI_ENGINE_OPENAI_PROVIDER=OpenAI
...
+# reCAPTCHA Enterprise (optional) — gates the public tenant signup endpoint.
+# When all three are set, the backend verifies the client-side token via the
+# reCAPTCHA Enterprise Assessments API. Leave blank to disable (verification is
+# skipped). RECAPTCHA_ENTERPRISE_SITE_KEY must also be exposed to the UI build.
+RECAPTCHA_ENTERPRISE_PROJECT_ID=
+RECAPTCHA_ENTERPRISE_SITE_KEY=
+RECAPTCHA_ENTERPRISE_API_KEY=
+# Optional tuning (defaults: 0.5 / onboarding_signup)
+RECAPTCHA_ENTERPRISE_SCORE_THRESHOLD=0.5
+RECAPTCHA_ENTERPRISE_EXPECTED_ACTION=onboarding_signup
```
mcgrawia, line 19: @TheLukaszNs-Blazity any new ENV vars on the engine also need to be added to the CFTs so we can set them in the deployed versions

CC @ntatsumi

genai-engine/ui/.env.local.example
```diff
@@ -1,3 +1,7 @@
 # Copy this file to .env.local and fill in your recording token
 METICULOUS_RECORDING_TOKEN=your_recording_token_here
 AMPLITUDE_API_KEY=your_amplitude_key_here
+# reCAPTCHA Enterprise score-based site key for the onboarding form (optional).
+# Leave blank to disable reCAPTCHA on the client. Must match the site key the
+# backend assesses against.
+RECAPTCHA_ENTERPRISE_SITE_KEY=your_recaptcha_enterprise_site_key_here
```
mcgrawia, line 7: for local dev should we leave this disabled by default?

mcgrawia: also similar question as the amplitude key, how does this get into the FE for the deployed envs?

mcgrawia (approved):

PR #1687, opened 2026-05-29
fix the demo dataset not matching with the transform variable names

notthattal:
## Description

Adds the ability to filter transform attribute paths for lists using negative indexes to get the last items from a list and updates the demo task resources to have consistent naming as well as updating one of the transforms to use this new negative index functionality.

## Partner MR
- https://gitlab.com/ArthurAI/unify-frontend/-/merge_requests/1355

### DIFF
--- a/genai-engine/src/utils/demo_task_fixtures/demo_task_resources.py
+++ b/genai-engine/src/utils/demo_task_fixtures/demo_task_resources.py
@@ -166,16 +166,16 @@ DEMO_TASK_DATASET_VERSION_REQUEST = NewDatasetVersionRequest(
         NewDatasetVersionRowRequest(
             data=[
                 NewDatasetVersionRowColumnItemRequest(
-                    column_name="query",
-                    column_value=query,
+                    column_name="input_messages",
+                    column_value=input_messages,
                 ),
                 NewDatasetVersionRowColumnItemRequest(
-                    column_name="response",
-                    column_value=response,
+                    column_name="output_messages",
+                    column_value=output_messages,
                 ),
             ],
         )
-        for query, response in DEMO_TASK_DATASET_ROWS
+        for input_messages, output_messages in DEMO_TASK_DATASET_ROWS
     ],
     rows_to_delete=[],
     rows_to_update=[],
--- a/genai-engine/src/utils/demo_task_fixtures/demo_task_resources.py
+++ b/genai-engine/src/utils/demo_task_fixtures/demo_task_resources.py
@@ -58,14 +58,14 @@ DEMO_TASK_SYSTEM_PROMPT = (
 )
 
 DEMO_TASK_ANSWER_RELEVANCE_EVAL_PROMPT = (
-    "Given a list of input and output messages:\n\n"
-    "<input_messages>\n"
-    "{{input_messages}}\n"
-    "</input_messages>\n\n"
-    "<output_messages>\n"
-    "{{output_messages}}\n"
-    "</output_messages>\n\n"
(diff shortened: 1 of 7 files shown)


genai-engine/src/utils/demo_task_fixtures/demo_task_resources.py
```diff
@@ -166,16 +166,16 @@
...
                 NewDatasetVersionRowColumnItemRequest(
-                    column_name="response",
-                    column_value=response,
+                    column_name="output_messages",
+                    column_value=output_messages,
                 ),
             ],
         )
-        for query, response in DEMO_TASK_DATASET_ROWS
+        for input_messages, output_messages in DEMO_TASK_DATASET_ROWS
```
mcgrawia: @notthattal don't the other evals and prompt use query/response? These all need to pick one or the other naming paradigm

notthattal: the other two evals use response but they use the 'DEMO_TASK_RESPONSE_EXTRACTION_TRANSFORM' which only extracts the output which we wouldn't use for the dataset. The eval that uses the transform with both input and output messages is 'DEMO_TASK_ANSWER_RELEVANCE_EVAL_PROMPT'

mcgrawia: Right but the prompt now uses the `query` column to populate the `query` variable in the experiment.

Is there a reason we shouldn't make them all the same?

notthattal: the only reason I didn't make it query was because technically input_messages is more correct because it's not really extracting only the query it's extracting the whole history but I can make it query

mcgrawia: hmm ok I think we need to reconcile this then because the current example rows in the dataset are not the input messages/output messages, they are question answer. If we decide to go this messages route, we'll also need to update the query template to expect input messages as a variable, and update the agent code to set it

notthattal: Another option I was thinking of was adding the ability to use negative indexes in a transform so that way we could extract `.values.-1.content` since the user query should be the last message that's sent we just can't guarantee the number of messages being passed in as input will be consistent. But, the last message would be what the user asked each time. but I don't think that follows the open inference convention

mcgrawia: ha yeah I made a ticket for this awhile back but never got to it https://arthurai.atlassian.net/browse/UP-4348

If we did have that though, what is the outcome? what would the prompt and dataset look like for the experiment?

notthattal: the prompt and dataset would be the same the only difference would be for the answer relevance eval instead of it specifying "given a list of input messages and a response" we could have it say "given a user query and a response" and it would just be the very last message in the history

mcgrawia: ok so if we do that, the dataset would be `query, response` and the evals and prompts would use query/response as inputs? and the transforms would set query/response?

notthattal: yeah exactly

mcgrawia: ok that seems reasonable to me

mcgrawia (approved):

PR #1693, opened 2026-05-29
fix(genai-engine): close three multi-tenant signup/auth gaps

videetparekh:
**HIGH SEVERITY**

1. **reCAPTCHA fails OPEN on HTTP 4xx / JSON / Pydantic errors** — `genai-engine/src/clients/recaptcha/recaptcha_verifier.py:101-106`. The broad `except Exception` after `raise_for_status()` / `.json()` / `model_validate(...)` absorbs attacker-triggered 4xx and returns `success=True`. An attacker can defeat reCAPTCHA on `/api/v2/tenant/signup` by sending a malformed token. Docstring scopes fail-open to transport errors only — narrow the except to `httpx.TransportError` / `TimeoutException`.

   > **Fix:** Errors an attacker can trigger (4xx, bad JSON, bad shape) now reject the request. Real Google outages still let legit signups through.

2. **`create_auto_task` stamps `DEFAULT_ORG_ID` regardless of caller** — `genai-engine/src/repositories/tasks_repository.py:400`. `TRACES_WRITE` (`schemas/enums.py:158-163`) admits `TASK_ADMIN` (a tenant role). A tenant `TASK_ADMIN` posting OTEL traces with an unmapped `service.name` auto-creates a task in `DEFAULT_ORG_ID`; their org-scoped reads (`get_all_tasks`, `enforce_query_org_scope`) never see it back. Either resolve `org_id` from `request.state.org_scope`, or drop `TASK_ADMIN` from `TRACES_WRITE`.

   > **Fix:** No code change. Tenants can't write traces in v1 by design, so the wrong-org stamp is currently unreachable. Flagged as a v2 hazard in the design doc's open questions; deferred to whichever ticket enables tenant trace ingest.

3. **`attach_notebook_to_experiment` does not validate notebook ownership** — `genai-engine/src/repositories/prompt_experiment_repository.py:1030`, `agentic_experiment_repository.py:740`, `rag_experiment_repository.py:942`. Experiment is org-checked via `_get_db_experiment(..., org_scope=)`, but `db_experiment.notebook_id = notebook_id` is written with no `NotebookRepository._get_db_notebook(notebook_id, org_scope=)`. A tenant can attach a foreign-org notebook UUID to their own experiment; downstream code that dereferences `experiment.notebook_id` without re-scoping (e.g., notebook run_count aggregations) leaks/corrupts.

   > **Fix:** Verify the notebook belongs to the same task before linking. Tenants can no longer reference foreign-org notebooks.

4. **`SpanRepository.get_session_traces` / `compute_session_metrics` pagination broken** — `genai-engine/src/repositories/span_repository.py:402-403,467-468`. The new tenant filter is applied AFTER `get_trace_ids_for_session` already returned `(total_count, page_slice)`. The code then sets `count = len(trace_ids)` (page-after-filter size), and the router forwards that as `SessionTracesResponse.count`. Pagination total is wrong; the `if count == 0` check 404s when owned traces happen to fall outside the current page slice; pages can be partially empty while more data exists. Filter must run at the SQL layer before LIMIT/OFFSET.

   > **Fix:** Move the org filter into the SQL query (before LIMIT/OFFSET). Counts, page boundaries, and 404 behavior are now correct for tenants.

5. **`LLMClient.acount_tokens` bypasses `_add_provider_credentials`** — `genai-engine/src/clients/llm/llm_client.py:188-198`. Only `api_key` / `api_base` are forwarded; `acompletion` (line 303) and `completion` (line 206) both call `_add_provider_credentials` first. For Azure (`api_version` required), Vertex (`vertex_project`/`vertex_credentials`), and Bedrock (AWS creds), the call raises. `base_chatbot_service.summarize_and_emit_replace` (line 124) calls this to gate context-overflow summarization, so chatbot summarization is broken on every non-OpenAI provider.

   > **Fix:** Token counter uses the same credential plumbing as `completion` / `acompletion`. Chatbot history summarization now works on Azure, Vertex, and Bedrock too.

**MEDIUM SEVERITY**

6. **`tenant_signup_routes` leaks raw exception text on a public endpoint** — `genai-engine/src/routers/v2/tenant_signup_routes.py:152,159-164`. Both `except ValueError` and `except Exception` raise `HTTPException(detail=f"Failed to signup tenant: {e}")` with no `from e`. `{e}` interpolates SQLAlchemy `IntegrityError` (constraint names, SQL fragments, table schema), `DBAPIError`, etc. back to an unauthenticated caller. Sanitize the response detail and use `raise … from e`.

   > **Fix:** Sanitized error messages — generic detail to the client, full traceback logged server-side. Exception chaining preserved with `from e`.

7. **`stream_demo_chatbot` leaks raw exception text** — `genai-engine/src/routers/v1/demo_task_routes.py:67-77`. Same `detail=f"Failed to stream demo chatbot response: {e}"` pattern on a `TASK_WRITE`-gated route (reachable by `TENANT_USER`).

   > **Fix:** Same treatment as item 6 — generic message to the client, traceback to logs.

8. **`save_feedback` org_id mismatch for task-less inferences** — `genai-engine/src/repositories/feedback_repository.py:35-41`. The JOIN through `DatabaseTask` falls back to `DEFAULT_ORG_ID` when `task_id IS NULL`. `inference_repository._save_prompt:388-396` and `save_response:428` use `SYSTEM_ORG_ID` for the same case. Same logical inference now has feedback in `DEFAULT_ORG_ID` and rule_results in `SYSTEM_ORG_ID` — multi-tenant isolation invariant broken at the schema level. Mirror the explicit `SYSTEM_ORG_ID` branch, or read from the newly-denormalized `DatabaseInference.org_id`.

   > **Fix:** Feedback strictly inherits the inference's org. Dropped the silent `DEFAULT_ORG_ID` fallback; tenant callers must match the derived org (defense-in-depth at the repo layer against bypassed route checks and task-deleted-mid-call races).

9. **`ChatbotRepository.stream_response` calls `chatbot_service.stream(...)` instead of `safe_stream(...)`** — `genai-engine/src/repositories/chatbot_repository.py:130`. `BaseChatbotService.safe_stream` (line 327) is the exception-to-SSE wrapper; `DemoChatbotService` uses it correctly via `demo_task_repository.py:448`. Here the main chatbot bypasses it, so any LLM/tool/prompt exception inside `stream` propagates uncaught through the StreamingResponse generator — the client gets an abrupt EOF with no SSE error frame, and tracing is never flushed.

   > **Fix:** One-identifier swap — call `safe_stream` instead of `stream`. Errors now reach the client as proper SSE error events instead of an abrupt connection close.

10. **`BaseChatbotService.summarize_history` slice-zero + StopIteration** — `genai-engine/src/services/chatbot/base_chatbot_service.py:81,84-86`. (a) `next(m for m in messages if m.role == MessageRole.SYSTEM.value)` raises `StopIteration` if no SYSTEM message exists — subclasses are free to construct prompts without one. (b) `keep_count = len(non_system) // 2` is `0` when `len(non_system) == 1`; `non_system[:-0]` → `[]` and `non_system[-0:]` → full list, so the function emits a bogus summary message plus the original message duplicated. Trigger: a single message large enough to exceed the model context window on its own.

    > **Fix:** Tolerate a missing system prompt, and bail out (return the input unchanged) when there's only one message to compress — that case used to corrupt the emitted history.

11. **`safe_stream` exception path skips tracing flush / span-end** — `genai-engine/src/services/chatbot/base_chatbot_service.py:337-342`. Every terminal branch in `stream` calls `self.tracing.end_span(...)` + `self.tracing.flush()`; the exception path here yields a generic SSE error without ending the agent_span or any open child spans. `InternalTraceService` has no `__enter__`/`__exit__` so Python unwinding does not finalize spans. Failed chatbot calls become invisible to observability (and to the demo chatbot continuous-eval queue).

    > **Fix:** On exception, close the root agent span with an error status and flush tracing before re-raising. Failed chatbot calls now show up in dashboards and the eval queue.

12. **`assert api_key.key is not None` used as runtime narrowing** — `genai-engine/src/routers/v2/tenant_signup_routes.py:168`. `ApiKey.key` is `Optional[str]`; the assert is the only narrowing before the response model is constructed. Under `python -O` the assert is stripped, `api_key.key=None` flows into Pydantic, the response 500s AFTER `db_session.commit()` at line 149 — caller never receives the raw key but the org/task/API-key/onboarding rows are committed (orphan state). Replace with `if api_key.key is None: raise HTTPException(500)`.

   > **Fix:** Replaced the strippable `assert` with an explicit check inside the transaction (before commit). A missing key now triggers rollback instead of leaving orphan rows behind under optimized Python builds.


mcgrawia (approved):
thanks for tracking these down and fixing @videetparekh !

PR #1700, opened 2026-06-01
[no-ref] allow users to read system and chatbot configs

mcgrawia:


notthattal (approved):

mcgrawia:
closing because these shouldn't be enabled for tenant users, will prevent calls on FE

PR #1723, opened 2026-06-03
[BE] Close 14 Pattern C org_scope gaps (UP-4470)

videetparekh:
## Summary

Follow-up to UP-4429. The Pattern C audit (design §7) flagged 14 task-scoped repository fetch methods missing `org_scope`. This PR closes **only the methods whose org isolation is a 1-hop check**, and **defers the 12 methods that would require a 2-hop join** to reach `tasks.org_id` (see "Deferred" below). `org_scope` is optional (`UUID | None = None`), so admin/system callers are unaffected.

Ticket: https://arthurai.atlassian.net/browse/UP-4470

## Closed in this PR (1-hop)

| Method | Isolation path | Hops |
|---|---|---|
| `rules_repository.get_rule_by_id` | rule fetched by PK, then `tasks_to_rules → tasks` existence check; `DEFAULT`-scope rules pass under any `org_scope` | 1 |
| `trace_transform_repository.get_transform_dependents` | gates on the parent transform's ownership (`transform → tasks.org_id`); a tenant that doesn't own the transform sees no dependents | 1 |

Call sites threaded: `task_management_routes.archive_task_rule` → `get_rule_by_id`; `transform_routes` dependents handler → `get_transform_dependents`.

## ⚠️ Deferred — 2-hop joins (not in this PR)

These 12 methods can only reach `tasks.org_id` via a **2-step join**, which adds latency on read paths. Rather than ship that, they're reverted to their original (no `org_scope`) form and deferred. The clean fix is to **denormalize `org_id`** onto these tables (turning the scope check into a 1-hop / 0-join filter) — which the UP-4470 ticket explicitly scopes **out**. Tracking on the ticket.

| Repo | Methods | 2-hop path |
|---|---|---|
| `metrics_repository` | `get_metric_by_id`, `get_metric`, `get_metrics_by_metric_id` | `metric → tasks_to_metrics → tasks` |
| `trace_transform_repository` | `get_latest_definition`, `list_versions`, `get_version_by_id` | `version → transform → tasks` |
| `agentic/prompt/rag_experiment_repository` | `_get_db_test_case`, `_get_db_test_cases` (×3) | `test_case → experiment → tasks` |

Note: all 12 have **zero route exposure today** (metrics + experiment test-case helpers are reached only by system/background paths; the transform-version methods are already preceded by a `get_transform_by_id(org_scope=…)` 404 guard at every route call site), so deferring them does not open a live gap.

## Tests

- `tests/multitenancy/test_pattern_c_repository_gaps.py`: per-method contract for the two closed methods (foreign org → 404/empty, `None` → visible, owner → visible) + `DEFAULT`-rule passthrough.
- Coverage inclusion list (`tests/coverage/test_repository_org_scope_coverage.py`): the two closed methods added; the 12 deferred ones intentionally excluded.


videetparekh:
Closing — on review, **none of the 14 methods audited in UP-4470 closes a reachable cross-org gap**, so there's nothing critical to ship here.

- **The 2 methods this PR touched are already protected by their callers on `dev`:**
  - `TraceTransformRepository.get_transform_dependents` — its route (`GET /traces/transforms/{transform_id}/dependents`) and the internal `delete_transform` caller both pre-validate the transform via the pre-existing `get_transform_by_id(org_scope=...)` / `_get_db_transform_by_id(org_scope=...)` 404 check. A foreign-org caller never reaches the method.
  - `RuleRepository.get_rule_by_id` — its route (`DELETE /tasks/{task_id}/rules/{rule_id}`) is `@enforce_org_scope()`-gated on the path task_id and explicitly rejects rules not linked to the caller's own task. The repo `org_scope` only changed a 400 into a 404.
- **The 12 deferred methods** (metrics, transform versions, experiment test cases) have **zero route exposure**.

So UP-4470 as scoped was hardening / Pattern C contract-consistency, not a vulnerability fix. Per discussion, closing rather than shipping unreachable defense-in-depth code. The durable follow-up (denormalize `org_id` onto child tables → single indexed predicate, no joins, uniform safe-by-default repos) is captured on the ticket.

PR #1725, opened 2026-06-03
chore: add .gitattributes and skip CI jobs when no relevant files changed

iabouhashish:
## Summary

Three fixes for recurring merge conflicts on feature branches, specifically the version line in `staging.openapi.json`.

### Root cause

`generate_changelog` was reading the version from `genai-engine/src/version` and writing it into the generated `staging.openapi.json`. That file was not included in the version bump CI until recently, so it was 98 patches behind (`2.1.503` vs `2.1.601`). Every time a developer ran `generate_changelog`, it injected the stale version into `staging.openapi.json`, causing a conflict when merging `dev`.

### Fix 1 — sync `genai-engine/src/version`

Updated `genai-engine/src/version` from `2.1.503` to `2.1.601` to match the rest of the repo. The version bump CI (`9660de680`) now keeps all three version files in sync going forward.

### Fix 2 — `generate_changelog` preserves the existing version

`changelog.py` now reads the version from the existing `staging.openapi.json` instead of from `genai-engine/src/version`. The version field in `staging.openapi.json` is exclusively owned by the bump CI (`sed` on every version bump commit). `generate_changelog` should only update API spec content, never the version number.

### Fix 3 — clean up `.gitattributes`

The original `.gitattributes` used `merge=ours` which is a **custom driver** requiring `[merge "ours"] driver = true` in git config — without it, the entries are silently ignored (flagged in code review). This was replaced with `-merge` (git's built-in take-ours strategy, no config needed), then further trimmed:

| File | Strategy | Rationale |
|------|----------|-----------|
| `version`, `genai-engine/version`, `genai-engine/src/version` | *(removed)* | Feature branches never touch these — the bump CI owns them on `dev`. No conflict to resolve. |
| `genai-engine/staging.openapi.json` | *(removed)* | Contains real API spec content. Conflicts here should be resolved manually, not silently discarded. Fix 2 eliminates the spurious version-line conflicts. |
| `genai-engine/uv.lock`, `ml-engine/uv.lock`, etc. | `-merge` (built-in) | Lock files conflict frequently across branches and are always regenerated with `uv sync` — safe to take ours. |
| `genai-engine/src/api_changelog.md` | `merge=union` | Append-only dated entries; union preserves additions from both sides. |

### Also on this branch (earlier commits)

- `dc93eae64` — trigger ml-engine CI on `staging.openapi.json` changes (ml-engine generates its client from this file)
- `9660de680` — include `genai-engine/src/version` in the version bump job

## Test plan
- [ ] Run `generate_changelog` on a feature branch behind `dev` — version in `staging.openapi.json` should remain unchanged
- [ ] Merge `dev` into a feature branch — no conflict on `staging.openapi.json` version line
- [ ] Lock file conflicts resolve silently to `HEAD` side (no merge markers)
- [ ] `api_changelog.md` union merge preserves entries from both sides


.github/workflows/arthur-engine-workflow.yml
```diff
@@ -201,6 +188,29 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v5
+      - uses: dorny/paths-filter@v2
+        id: changes
+        with:
+          filters: |
+            changed:
+              - 'ml-engine/**'
+              - 'arthur-observability-sdk/**'
+              - 'genai-engine/staging.openapi.json'
```
fryz, line 213: i don't think either of these are relevant for the ml-engine? should these live in the check-genai-engine-changes?

iabouhashish: arthur-observability-sdk/** and genai-engine/staging.openapi.json both belong in check-ml-engine-changes. Looking at lines 369 and 447 of the workflow, ml-engine generates its genai_client directly from genai-engine/staging.openapi.json as part of both unit tests

jballone (approved):

jdbarthur (approved):

PR #1744, opened 2026-06-08
fix(UP-4505): stop red target-lost hint flashing in task tour Step 3

TheLukaszNs-Blazity:
Clicks inside an active tour step briefly detach/re-attach the highlighted node (list re-renders, drawer swaps), which momentarily resolves the target to null and emits target:lost. The checklist panel rendered the red "target lost" hint on that transient state, so the copy flashed on nearly every click in the evals section (Step 3).

Gate the hint behind useDelayedFlag, a TanStack Pacer-backed hook that debounces the lost->shown edge (~200ms) while clearing instantly when the target re-resolves, so transient blips never surface the hint but a genuinely missing target still does.


xV3rt:
Realistically, do we even need the hint in the first place? Does it provide any value?

xV3rt (approved):

PR #1771, opened 2026-06-10
Emit guardrail trace spans from stateful validate

xV3rt:
Related to UP-4554


genai-engine/src/services/trace/guardrail_span_emitter.py
```diff
@@ -0,0 +1,145 @@
...
+        span = tracing.start_guardrail_span(
+            name=f"guardrail.validate_{stage}",
+            parent_span_id=parent_span_id_bytes,
+            span_id=own_span_id,
+            user_id=user_id,
+            session_id=session_id,
+        )
+        tracing.set_input_json(span, input_payload)
+        tracing.set_output_json(span, result.model_dump(mode="json"))
+        tracing.end_span(span)
```
mcgrawia: @xV3rt I think this overall looks good, but one major concern is that it seems like the span is immediately started/ended. The span should be started before the guardrail executes, then ended after so the user can see how long the guardrail took to complete. Right now it seems like the duration will be nearly zero because the span is created and immediately ended after the guardrail has already run

genai-engine/src/validation/prompt.py
```diff
@@ -30,8 +34,27 @@ def validate_prompt(
...
+            rule_results=result.rule_results,
+            input_payload={"prompt": body.prompt},
+            is_response=False,
+            trace_id=trace_id,
+            parent_span_id=parent_span_id,
+            user_id=body.user_id,
+            session_id=body.conversation_id,
+        )
+
+    return result
```
mcgrawia: more specifically, you should start a guardrail span here before the call to `.evaluate()` above. Then, if the guardrail evaluates successfully, it should store the results on the span. If it errors, the code should catch the error and set a failure status on the span with an error message before ending the span and propagating the error

genai-engine/src/services/trace/guardrail_span_emitter.py
xV3rt: Great catch, thanks for spotting, should be addressed now

genai-engine/src/routers/v2/validate_routes.py
```diff
@@ -131,6 +131,16 @@ def default_validate_response(
 def validate_prompt_endpoint(
     body: PromptValidationRequest,
     task_id: UUID,
+    trace_id: str | None = Query(
+        default=None,
+        description="Optional trace ID (32-hex) to attach the emitted guardrail span "
+        "to an existing trace. If omitted, a trace is derived from the inference ID.",
+    ),
+    parent_span_id: str | None = Query(
```
mcgrawia: apologies @xV3rt missed this in the first round, but is it possible to pass these parameters as headers instead of query args? Typically trace info is stored in request/response headers

mcgrawia: otherwise lgtm!

xV3rt: Addressed!

fryz:
just leaving a quick comment here that this looks like what i expected re: BE changes, so overall I approve of this approach.

genai-engine/src/routers/v2/validate_routes.py
```diff
@@ -131,21 +132,19 @@ def default_validate_response(
 def validate_prompt_endpoint(
     body: PromptValidationRequest,
     task_id: UUID,
-    trace_id: str | None = Query(
+    traceparent: str | None = Header(
```
mcgrawia, line 135: TIL there's a W3C header for this! nice!

mcgrawia (approved):

PR #1772, opened 2026-06-10
create gcs image support tdd

notthattal:
## Description

Creates the GCS image support for datasets TDD

## Jira Ticket

- https://arthurai.atlassian.net/browse/UP-4558


genai-engine/technical_documentation/gcs_image_support_tdd.md
```diff
@@ -0,0 +1,215 @@
+# GCS Image-URI Resolution — Technical Design Document
+
+## Goal
+
+Support image columns whose cell values are GCS bucket URIs (`gs://bucket/path/image.png`) in addition to inline base64. When the raw-data (FetchData) job runs in the engine, resolve each `gs://` value to a base64 string so the platform renders the image. Inline base64 continues to work unchanged.
+
+## Plan
+
+### Step 1: Create the `ImageResolver`
```
fryz, line 9: It would be better to move the batching + parallelism out of the ImageResolver class and put it into the job (or in a new abstraction) - ImageResolver should just take a single row and own the responsibility of resolving which type of image it is (gcs, base64, others later) and return base64.

genai-engine/technical_documentation/gcs_image_support_tdd.md
```diff
@@ -0,0 +1,215 @@
...
+        dataset = get_dataset_or_available_dataset_from_id(datasets_client, dataset_id) # new helper
+        if dataset.dataset_schema:
+            schema = client_to_common_dataset_schema(dataset.dataset_schema)
+            image_resolver = ImageResolver(
+                self.connector_constructor,
+                dataset.project_id,
+                self.logger,
+            )
+            
+            data = image_resolver.resolve_images(data, schema)
```
fryz: The reason to push the orchestration / parallelism here would be to avoid having to copy the entire dataset into memory (already in duckdb) and instead only pass a batch or even single row at a time.

fryz (commented):
Some notes on keeping the abstractions tighter and avoiding moving more data around than we need to

genai-engine/technical_documentation/gcs_image_support_tdd.md
notthattal: Moved outside the resolver class in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1772/changes/8965cce36ec9c522295f84a468ae60bd63dad3eb)

genai-engine/technical_documentation/gcs_image_support_tdd.md
notthattal: I moved the orchestration outside the resolver class, but it won't resolve loading the dataset into memory. Currently that's done in _load_data within the fetch excecutor regardless of these changes. We just use the already loaded data to be resolved

mcgrawia (approved):
LGTM 👍

PR #1775, opened 2026-06-11
UP-4930 token limits for tenant orgs

mcgrawia:
## UP-4390: Tenant Token Credit System

Implements per-org lifetime LLM-token caps with deployment-level opt-in. New tenant signups can be capped at a configured limit; existing tenants are retroactively capped when the migration runs *if* the deployment is opted in. Calls that exceed the limit get a structured `429 TOKEN_LIMIT_EXCEEDED` response, and the FE surfaces a global "out of credits" dialog instead of inline errors.

## Screenshots of Limits Running Out in the UI

Screenshot 2026-06-16 at 10 20 03 PM
Screenshot 2026-06-16 at 10 20 15 PM
Screenshot 2026-06-16 at 10 40 21 PM
Screenshot 2026-06-16 at 10 41 59 PM
image

### Design highlights (resolved during review)

- **Single enforcement point in `LLMClient`.** The quota check fires *inside* `LLMClient.completion`/`.acompletion`, not at route handlers. New endpoints that fire LLM calls can't accidentally bypass the gate. mid-flow exhaustion (chatbot tool loops, multi-step experiments) is also caught at the next LLM call.
- **`org_id` is a required `UUID`, not `Optional`.** Every LLM call bills the org that owns the *resource* (task, dataset, annotation, etc.), not the caller. Admin keys hitting tenant resources still bill the tenant. Unmetered orgs are encoded as `tokens_limit IS NULL`, not via a sentinel arg.
- **Opt-in by deployment.** `GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT` defaults to unset → no limits applied. Setting it to a positive integer (a) caps new tenant signups, and (b) the migration retroactively backfills existing tenant orgs to that value when it runs with the env var present.
- **Independent billing transaction.** `record_org_token_usage` and `enforce_token_quota` open their own DB sessions via `db_session_context()` — token-usage writes commit independently of the caller's transaction, so accounting persists even if the surrounding request rolls back.

### Backend

**Schema** — single Alembic migration `5e84b8d7aa1f`:
- Adds `tokens_limit BIGINT NULL` and `tokens_used BIGINT NOT NULL DEFAULT 0` to `organizations`.
- Conditional backfill: if `GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT` is set to a positive integer at migration time, every existing tenant org (everything except `DEFAULT_ORG_ID` and `SYSTEM_ORG_ID`) is updated to that value. Otherwise nothing is backfilled.

**Config / constants:**
- `GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT` env var, `Config.default_tenant_token_limit() -> Optional[int]` returning `None` for unset / invalid / non-positive.

**Gate (`enforce_token_quota`):**
- Raises `HTTPException(429, detail={"error_code": "TOKEN_LIMIT_EXCEEDED", "message", "tokens_limit", "tokens_used"})`.
- Skips for unmetered orgs (`tokens_limit IS NULL`).
- Falls through silently for unknown org IDs — billing isn't the place to surface 404s.
- Invoked once at the top of `LLMClient.completion`/`.acompletion`.

**Recorder (`record_org_token_usage`):**
- Atomic SQL `tokens_used = tokens_used + :n`; no Python-side read-modify-write.
- Zero / negative no-op (defends against responses missing a usage block).
- Failures are logged but never raised — the LLM call already happened.

**Threading:** `org_id: UUID` (required) added to ~12 production signatures including `ChatCompletionService` (all four methods), `BaseChatbotService.__init__`, `LLMEvaluator.run`, `MLEvaluator.run` (the `BaseEvaluator` contract is now uniform — no `# type: ignore[override]`), `LLMEvalsRepository.run_llm_eval`, `AgenticPromptRepository.run_saved_prompt`/`run_unsaved_prompt`, `SyntheticDataService.generate_initial`/`continue_conversation`. Sources: chatbot/prompt routes use `task.org_id` (via `get_validated_task`), system chatbot uses `SYSTEM_ORG_ID`, experiment background threads query `DatabaseTask.org_id`, continuous-eval worker uses `annotation.org_id`, dataset routes use a `_dataset_org_id(...)` helper, unsaved prompts fall back to `current_user.org_scope or DEFAULT_ORG_ID`.

**Background-job error UX:**
- `continuous_eval_queue_service` does a pre-check at job start so the trace's downstream "missing variables" symptom is replaced by the actual credit-limit message on the annotation. The broad `except` also recognizes `429 TOKEN_LIMIT_EXCEEDED` for mid-eval gate trips.
- `prompt_experiment_executor` writes the credit-limit message to `prompt_result.output_content` (the field the FE already renders) when the gate fires; non-quota failures stay log-only so stack traces don't leak as model "output".
- Shared helper `extract_token_limit_message(exc) -> Optional[str]` returns the user-facing message for our 429, `None` for anything else — used by both workers.

**Tenant signup:** `create_organization(tokens_limit=Config.default_tenant_token_limit())`. When the env var is unset, new orgs get `NULL` (unmetered) — same as the migration's no-backfill default.

### Frontend

- **`src/lib/api-errors.ts`** — `isTokenLimitExceededError(err)` + `getTokenLimitDetail(err)`, handles both axios responses and raw-fetch error objects. Detects `429 + error_code === "TOKEN_LIMIT_EXCEEDED"`.
- **`src/contexts/OutOfCreditsContext.tsx`** — provider + `useOutOfCreditsDialog()` hook.
- **`src/components/common/OutOfCreditsDialog.tsx`** — MUI dialog with theme tokens, mounted at the app shell.
- **9 call sites wired**: `useChatbot` (chatbot + demo chatbot streaming), `streamCompletions` + `useRunPrompt` (prompt playground; `streamCompletions`'s `onError` gained an optional `cause: unknown` so the structured 429 body can route to the dialog instead of being stringified), `usePromptExperiments` (create), `useRagExperiments` (create), `useCreateNewExperiment`, `useTestRun` (create test run), `useRerunContinuousEval`, `useSyntheticDataSession` (start + sendMessage).

### Infrastructure / Deploy

**CloudFormation** — new `GenaiEngineDefaultTenantTokenLimit` parameter (Type: String, Default: '') plumbed end-to-end through six templates:
- `genai-engine/arthur-genai-engine-ecs-task-definition.yml` (+ GPU variant) — parameter + container `Environment` entry as `GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT`.
- `root-arthur-genai-engine-cpu.yml` / `-gpu.yml` — parameter, parameter-group, label, and pass-through to nested stack.
- `root-arthur-engine-cpu.yml` / `-gpu.yml` (umbrella) — same.

**arthur-scope CI** — `scope/deployments/environments/aws-prod-platform/.gitlab-ci.yml` sets `GenaiEngineDefaultTenantTokenLimit="5000000"` alongside the existing `GenaiEngineDemoMode="enabled"`. On the next deploy, the prod-platform stack opts in with a 5M tenant cap.

### Test coverage

- **23 new unit tests** in `tests/unit/repositories/test_organizations_repository.py`:
  - `OrgTokenQuotaStatus.is_exhausted` — 6 parametrized cases covering unmetered, fresh, under, at-limit, over.
  - `enforce_token_quota` — passes for unmetered, passes under limit, raises 429 at-limit and over-limit (with full structured detail assertions), silent pass-through for unknown org.
  - `record_org_token_usage` — atomic increments accumulate, zero/negative are no-op, unknown-org failures stay silent.
  - `extract_token_limit_message` — returns clean message for our 429, returns `None` for non-429, non-our-429 (e.g. provider rate-limit wrappers), and non-HTTPException types.
  - `LLMClient.completion`/`.acompletion` — exhausted org → 429 with `litellm.{a,}completion` never called; under-limit → request flows through.
- **Test infra fix:** `db_session_context()` now reuses the already-initialized `SINGLETON_DB_ENGINE` instead of re-constructing `DatabaseConfig()` on every call, and the autouse fixture sets `TEST_DATABASE=True`. Prod and test session-opening paths converge.
- Full suite: **1900 unit tests passing**, mypy clean across 286 src files, FE `tsc -b` + ESLint + Prettier all green.

### Migration

Single Alembic head (`5e84b8d7aa1f`). The previously-orphan `e1f2a3b4c5d6` (`add_eval_type_support`) head is now serialized into the chain as the parent of this migration — no separate merge revision required.

### Test plan

- [ ] `uv run pytest -m "unit_tests"` — passes (1900 / 7 skipped).
- [ ] `uv run mypy src/` — clean.
- [ ] Manually: set `GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT=10` on a fresh tenant org, run a chatbot turn — confirm second turn opens the out-of-credits dialog with the configured limit/usage values.
- [ ] Verify token usage accrues: `SELECT tokens_used FROM organizations WHERE id = '<tenant>';` increments after each LLM call.
- [ ] Verify a continuous eval annotation on an over-budget org shows the credit-limit message, not "Could not extract variables".
- [ ] Verify a prompt experiment over the limit renders the credit-limit message in `output_content`, not a blank "failed" cell.
- [ ] Unset the env var on a fresh deploy — confirm tenants signup without a limit and the migration doesn't backfill.

### Out of scope / known limitations

- **Scorer LLM calls** (rule engine — hallucination, sensitive-data, etc.) instantiate LangChain `ChatOpenAI` / `AzureChatOpenAI` directly. They're not billed in this MR — automatically covered once they migrate onto `LLMClient` (separate effort).
- **ML Engine polling worker** runs out-of-process; not billed by this MR.
- **No admin endpoint to raise per-org limits** — Ops adjusts via DB UPDATE for now. Self-serve raising is its own ticket.
- **Helm and docker-compose deployment paths** weren't updated — Helm doesn't currently wire any tenant-mode env vars, so adding just this one would be inconsistent; let me know if you want them mirrored.


genai-engine/src/utils/constants.py
```diff
@@ -24,6 +24,10 @@
 GENAI_ENGINE_AGENTIC_UI_ENABLED_ENV_VAR = "GENAI_ENGINE_AGENTIC_UI_ENABLED"
 GENAI_ENGINE_DEMO_MODE_ENV_VAR = "GENAI_ENGINE_DEMO_MODE"
 GENAI_ENGINE_TENANT_MODEL_WHITELIST_ENV_VAR = "GENAI_ENGINE_TENANT_MODEL_WHITELIST"
+GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT_ENV_VAR = (
+    "GENAI_ENGINE_DEFAULT_TENANT_TOKEN_LIMIT"
+)
+DEFAULT_TENANT_TOKEN_LIMIT = 5_000_000
```
fryz: shouldn't the default be null? otherwise, all of our engine customers will start getting rate-limited when they create new organizations (which eventually will be new projects in the platform)

genai-engine/src/repositories/organizations_repository.py
```diff
@@ -1,11 +1,114 @@
...
+    Opens its own DB session so it's safe to call from any context
+    (in-request, background thread, queue worker) without coupling to a
+    caller-owned session.
+    """
+    with db_session_context() as session:
+        status = OrganizationsRepository(session).get_token_quota_status(org_id)
+    if status is None or not status.is_exhausted:
+        return
+    raise HTTPException(
+        status_code=402,
```
fryz: note that i think it's more conventional to use 429 for token limits being hit

genai-engine/alembic/versions/2026_06_04_1000-5e84b8d7aa1f_add_token_limit_to_organizations.py
```diff
@@ -0,0 +1,69 @@
+"""add token limit to organizations
+
+Revision ID: 5e84b8d7aa1f
+Revises: e1f2a3b4c5d6
```
videetparekh, line 4: There seem to be two alembic heads. I expect this will cause problems?

mcgrawia: where are you seeing that? this is what I see locally on this branch

```
uv run alembic history
e1f2a3b4c5d6 -> 5e84b8d7aa1f (head), add token limit to organizations
2966018cec1d -> e1f2a3b4c5d6, add eval_type support to llm_evals and continuous_evals
d4e5f6a7b8c9, 490cd5586c25 -> 2966018cec1d (mergepoint), merge onboarding and demo_certificates heads
7e1c3a4b5d2f -> d4e5f6a7b8c9, create_onboarding_tables
7e1c3a4b5d2f -> 490cd5586c25, add demo_certificates table
3abf881da275 -> 7e1c3a4b5d2f (branchpoint), denormalize org_id to trace_metadata and spans
d894934c8994 -> 3abf881da275, denormalize org_id to task-scoped tables
5de5704b5892 -> d894934c8994, backfill orphan inferences to the __unmapped__ task
697657f9af66 -> 5de5704b5892, add org_id to api_keys
514464b8ca3d -> 697657f9af66, add org_id to tasks
b3f7a2c1d9e0 -> 514464b8ca3d, create organizations table
```

genai-engine/src/repositories/organizations_repository.py
mcgrawia: thanks. addressed in latest commit

genai-engine/src/utils/constants.py
mcgrawia: updated in latest commit:
- env var is null by default, and won't backfill or create tenant orgs with limits
- when set to a positive int, it will backfill (only when the migration runs) and create tenant new orgs set to that limit

mcgrawia: corresponding MR to set the configuration for the engine.platform env: https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1782

genai-engine/alembic/versions/2026_06_04_1000-5e84b8d7aa1f_add_token_limit_to_organizations.py
videetparekh: There's a comment in both the MR description, also raised by my own claude on review. It sounds like you ran the migrations and it was ok though?

mcgrawia: whoops thanks missed that. updated

genai-engine/alembic/versions/2026_06_04_1000-5e84b8d7aa1f_add_token_limit_to_organizations.py
```diff
@@ -0,0 +1,96 @@
...
+            sa.BigInteger(),
+            nullable=False,
+            server_default=sa.text("0"),
+        ),
+    )
+
+    # Conditional retroactive backfill — only fires when the deployment
+    # has opted in via env. Default / system orgs always stay NULL.
+    default_limit = _parse_default_limit_from_env()
+    if default_limit is not None:
```
fryz, line 80: will this fire every time the value is updated? eg: if we change from 5k -> 15k? (for example)

genai-engine/src/clients/llm/llm_client.py
```diff
@@ -101,6 +107,29 @@ def __init__(
         self.vertex_credentials = vertex_credentials
         self.aws_bedrock_credentials = aws_bedrock_credentials
 
+    def record_token_usage(
```
fryz, line 110: im wondering if this would be better served as a decorator over functions that return a Response object (eg: completion, acompletion)

also curious why this exists separate from the helper method `record_org_token_usage`

fryz: ah i see - we use the `record_org_token_usage` outside of the client (eg: in the evals queue)

fryz (approved):

genai-engine/alembic/versions/2026_06_04_1000-5e84b8d7aa1f_add_token_limit_to_organizations.py
mcgrawia: No, the limit is set in the DB when the org is created. If an org is created with a limit, it will need to be upped in the DB if we want to give it more usage, which we could do with some new admin endpoints or similar down the road.

PR #1785, opened 2026-06-12
add gcs image support for datasets

notthattal:
## Description

Adds support for gcs image uris to be converted to base64 for datasets

#### Raw Dataset
Screenshot 2026-06-12 at 12 19 41 PM

#### In Platform
Screenshot 2026-06-12 at 12 04 38 PM

## Jira Ticket

- https://arthurai.atlassian.net/browse/UP-4475

## Partner MR

- https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1776


ml-engine/src/ml_engine/connectors/connector.py
```diff
@@ -38,3 +38,6 @@ def test_connection(self) -> ConnectorCheckResult:
     @abstractmethod
     def list_datasets(self) -> PutAvailableDatasets:
         raise NotImplementedError
+
+    def extract_image(self, image_uri: str) -> str:
+        return image_uri
```
fryz: since this is an Abstract Base Class (ABC), the methods in this class should all be decordated w/ `@abstractmethod` and raise a `NotImplementedError`, and then ALL of the connectors (not just the bucket connector) will need to provide the implementation for `extract_image`.

ml-engine/src/ml_engine/connectors/bucket_based_connector.py
```diff
@@ -592,3 +593,7 @@ def list_datasets(self) -> PutAvailableDatasets:
         raise NotImplementedError(
             "List datasets not implemented for bucket-based connectors.",
         )
+
+    def extract_image(self, image_uri: str) -> str:
```
fryz, line 598: there should be some validation here that the image_uri matches the `gs://` prefix (or whatever prefix you expect it to match based off the prefixes in image_tools)

fryz: also - we probably want to wrap the stuff in this method in a try/catch and raise a specific exception that can be caught by the ImageResolver

ml-engine/src/ml_engine/tools/image_resolver.py
```diff
@@ -0,0 +1,75 @@
...
+            return image_uri
+
+        for connector_spec in matches:
+            try:
+                connector = self.connector_constructor.get_connector_from_spec(
+                    connector_spec.id,
+                )
+                return connector.extract_image(image_uri)
+            except Exception:
+                continue
```
fryz: if we wrap the `extract_image` calls in a try/catch and raise a typed exception, we can handle that here instead of just swallowing all generic exceptions and not logging them 

at the very least, we should be logging the exception in case there's something wrong that we don't expect

notthattal: the reason I chose not to raise the error here is because this tries all the connectors associated with a specific bucket so if the 1st connector failed to fetch the image (for example), the last one might be successful. Also, if there's an api error on google's end or something we would still be able to load the whole dataset. It would just show the text instead of the actual image. 

Either way I think logging makes sense

ml-engine/src/ml_engine/connectors/connector.py
notthattal: Updated in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1785/changes/e1178c95fd8b2739be316146a1ef4f01ba9d3327)

ml-engine/src/ml_engine/connectors/bucket_based_connector.py
notthattal: Updated both in [this commit](https://github.com/arthur-ai/arthur-engine/pull/1785/changes/e1178c95fd8b2739be316146a1ef4f01ba9d3327)

ml-engine/src/ml_engine/tools/image_resolver.py
fryz: yeah - think it's ok to swallow the exception and not re-raise it, but think we should handle exceptions from the `extract_image` calls so we have the ability to log out which exception was thrown

fryz (approved):

PR #1796, opened 2026-06-15
Remove Chat

ntatsumi:
The PR is a net −3,798 lines (4,073 deleted, 275 added) across 59 files. Since the feature is unused, the value is almost entirely in reduction. Concretely, based on what the diff actually touches:

1. Smaller attack surface & fewer secrets to manage
- Deletes chat_routes.py (386 lines) — a whole authenticated API surface (/chat endpoints, permission mappings) you no longer have to secure, pen-test, or patch.
- Removes document upload + file parsing (file_parsing.py), which is classic untrusted-input territory.
- Eliminates S3 / Azure blob clients (S3Client, azure_client, InMemoryS3Client) and their credentials (DEFAULT_S3_BUCKET, DEFAULT_S3_ASSUMABLE_ROLE_ARN, assumable-role ARN). Fewer cloud permissions and stored secrets in the blast radius.
- Drops the embedding-model secret config (GENAI_ENGINE_OPENAI_EMBEDDINGS_NAMES_ENDPOINTS_KEYS) from CloudFormation and .env.

2. Lower operating cost & infra simplification
- No more embedding-model endpoints or S3 buckets/IAM roles to provision — the CloudFormation task definitions (GPU + CPU) each shed ~42 lines of config.
- Stops paying for embedding API calls and object storage for a feature nobody uses.

3. Reduced maintenance burden
- Removes whole subsystems: src/chat/, document & embedding & configuration repositories, document DB models, and chat-related schemas/enums (~150 lines of internal_schemas.py alone).
- Deletes ~1,000+ lines of now-dead tests and the load_test_chat.py load harness — less to keep green in CI.
- This is the kind of dependency surface that drives CVE churn (note the recent LangChain CVE upgrade in commit 94d83167b). The chat code imports langchain_core messages and embeddings; trimming it shrinks future security-upgrade exposure.

4. Cleaner, more honest API
- The OpenAPI spec drops 853 lines (staging.openapi.json) and the generated frontend API client drops 437 lines. Consumers no longer see endpoints that aren't supported.
- A changelog entry is added (api_changelog.md) documenting the removal — good API hygiene.

5. Database simplification
- A migration (...drop_chat_tables.py) drops the chat/document/embedding tables, reducing schema complexity and storage.
- inference_repository.py gets meaningfully simpler (−95/+12), removing chat-specific query paths from a core hot-path repository.

6. Less cognitive load for the team
- New engineers (and onboarding skills) no longer have to understand a dead feature. Removing unused code is the cheapest, lowest-risk complexity reduction available — there's no migration path or deprecation window needed since usage is zero.

Backwards compatibility (very minor):
ConfigurationUpdateRequest/response in system_management_routes.py lose `chat_task_id` and `document_storage_configuration`. This endpoint is general-purpose, not chat-specific so clients reading the config response will find those fields gone.

We could leave the `chat_task_id` and `document_storage_configuration` attributes but I don't think any end user is using them.


genai-engine/alembic/versions/2026_06_15_1200-f0a1b2c3d4e5_drop_chat_tables.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def upgrade() -> None:
+    # Drop in FK-safe order. CASCADE clears dependent indexes/constraints, and
+    # IF EXISTS keeps this idempotent on databases that never created them.
+    op.execute("DROP TABLE IF EXISTS inference_embeddings CASCADE")
+    op.execute("DROP TABLE IF EXISTS embeddings CASCADE")
+    op.execute("DROP TABLE IF EXISTS documents CASCADE")
+    # The vector extension was only ever created when chat was enabled.
+    op.execute("DROP EXTENSION IF EXISTS vector")
```
mcgrawia, line 32: @ntatsumi do you know what the full list of tables is that will be affected by the CASCADE? Mostly just curious if any other tables outside these 3 have foreign keys with these ones that are being deleted

genai-engine/src/schemas/enums.py
```diff
@@ -29,26 +29,9 @@ class RuleScoringMethod(str, Enum):
...
-
 # These are keys in config key : value pairs
 class ApplicationConfigurations(str, Enum):
-    CHAT_TASK_ID = "chat_task_id"
     DEFAULT_CURRENCY = "default_currency"
-    DOCUMENT_STORAGE_ENV = "document_storage_environment"
-    DOCUMENT_STORAGE_BUCKET_NAME = "document_storage_bucket_name"
-    DOCUMENT_STORAGE_ROLE_ARN = "document_storage_assumable_role_arn"
-    DOCUMENT_STORAGE_CONTAINER_NAME = "document_storage_container_name"
-    DOCUMENT_STORAGE_CONNECTION_STRING = "document_storage_connection_string"
```
mcgrawia, line 51: @ntatsumi do we need to clean out the system and application configuration tables that store these values? There might be DBs out there that had these set at one point?

genai-engine/ui/src/components/ml-evaluators/MLEvalFormModal.tsx
```diff
@@ -18,7 +18,12 @@ import Typography from "@mui/material/Typography";
 import { isAxiosError } from "axios";
 import React, { useEffect, useState } from "react";
 
-import type { CreateMLEvalRequest, EvalTypeInput } from "@/lib/api-client/api-client";
+import type { CreateMLEvalRequest } from "@/lib/api-client/api-client";
+
+// Local union for the ML eval type selector. Previously imported as
+// `EvalTypeInput` from the generated client; that alias was dropped from the
+// backend OpenAPI schema, so it's declared locally to preserve prior behavior.
+type EvalTypeInput = "llm_as_a_judge" | "pii" | "pii_v1" | "toxicity" | "prompt_injection";
```
mcgrawia, line 26: where did this type come from? it seems like this was previously in the generated api-client, so the type seems to have gone missing from the API spec?

iabouhashish:
Screenshot 2026-06-17 at 16 07 16
Would Chat Playground need to be deleted as well?

PR #1809, opened 2026-06-17
docs: Add platform-connected multi-tenancy technical plan

fryz:
## Summary

Add a comprehensive technical research and design document outlining the plan to connect the GenAI Engine to the platform's identity and authorization infrastructure, enabling platform-driven SSO and RBAC while maintaining backward compatibility with standalone API-key-only deployments.

## Changes

- **New document:** `PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md` (294 lines)
  - Frames the business problem (unit economics for SaaS land-and-expand)
  - Validates current multi-tenant state against the codebase (org-level tenancy already implemented in v1)
  - Documents the taxonomy mapping between platform (projects/applications/workspaces) and engine (orgs/tasks)
  - Identifies 8 major technical changes required:
    1. Keycloak auth client for platform-minted token validation
    2. Platform-driven authorization (resolve user's accessible orgs via backend call)
    3. Org scope refactor from scalar to set (≈106 touch points)
    4. Unified engine UI (single pane across all user orgs)
    5. Gated org creation when platform-connected
    6. Onboarding flow for existing standalone engines
    7. Cross-org ML engine job support
    8. Engine ↔ platform Keycloak connectivity
  - Tags all claims with validation status (✅ validated in code, ⚠️ partially true, ❓ platform-dependent)
  - Provides file references for validated findings

## Implementation Notes

- Document is explicitly marked as a **research plan, not an implementation spec**
- Approximately half the proposed architecture lives in the platform/control-plane repo (not in this workspace), so those items are tagged as unvalidated dependencies
- Establishes clear scope boundaries: engine UI stays in engine; SSO is free/self-serviceable; RBAC is the gated enterprise lever
- Identifies the key refactor work: converting the scattered org-scope enforcement (60 path endpoints, 11 query endpoints, ~35 repository methods) from single-org to multi-org filtering
- Companion to the existing `MULTI_TENANCY_DESIGN.md` which documents the shipped v1 org model

https://claude.ai/code/session_01UqRkTSSFkVXoUD6dp9cq8u

### DIFF
--- a/genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
+++ b/genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
@@ -1,0 +1,488 @@
+# GenAI Engine — Platform-Connected Multi-Tenancy: Technical Plan
+
+> **Status:** Research / design proposal (for adversarial review)
+> **Date:** 2026-06-16
+> **Source material:** Notion "2026 Engine Architecture Convo" (Convos 1–3), Fireflies transcripts
+> `01KT7C5884KREC51V5NAFAS4R9` (Convo 1), `01KTCHXCADTZXRRCBDQ8B6HVWH` (Convo 2),
+> `01KTS4893H7K9XE852CPRBS05J` (Convo 3), and direct code investigation of
+> `genai-engine/` and `ml-engine/`.
+> **Companion doc:** [`MULTI_TENANCY_DESIGN.md`](MULTI_TENANCY_DESIGN.md) (the shipped v1 org model this plan builds on).
+
+---
+
+## 0. How to read this document
+
+This is a *research plan*, not an implementation spec. Every factual claim about the
+current system has been checked against the code and is tagged:
+
+- ✅ **Validated** — confirmed in this repo, with file references.
+- ⚠️ **Partially true / overstated** — the meeting notes assert something the code only
+  partly supports. These are the dangerous ones; read them carefully.
+- ❓ **Unverifiable here** — depends on the platform / control-plane codebase, which is a
+  **separate repository not available in this workspace**. Roughly half of the proposed
+  architecture lives there and is therefore *unvalidated*.
+
+Section 9 (Assumptions Ledger) and Section 10 (Weak Spots & Open Questions) are the
+adversarial-review payload Ian asked for at the end of Convo 3.
+
+---
+
+## 1. Framing & Scope
+
+### 1.1 What problem are we actually solving?
+
+The business driver (Convo 1) is **unit economics for a SaaS land-and-expand motion**:
+sign up free → get a hosted engine in <60s → convert to Pro (~$200/mo) → Enterprise. The
+single requirement that most constrains *engine architecture* is unit economics; "easy to
+adopt," "data governance," and "land & expand" mostly drive platform/packaging, not the
+engine internals.
+
+Two distinct technical bets came out of that:
+
+1. **Multi-tenant inference / engine** (cost) — pull the expensive models out of the
+   per-tenant engine into a shared, pooled inference service so a lightweight engine can run
+   for **< $200/mo**. *This plan does not cover the inference-service split* — it is a
+   separate, larger effort. Flagging it because the unit-economics math (crossover at ~30–40
+   paying tenants) depends on it, not on the auth work below.
+2. **Platform-connected engine** (adoption + governance) — a user clicks from the platform
+   into the engine with **no separate API-key login (SSO)**, and sees only the resources they
+   are authorized for, governed by the platform's Keycloak (authN) + SpiceDB (authZ). **This
+   is what the plan below covers.**
+
+### 1.2 What is already done (do not re-build)
+
+The engine is **already multi-tenant at the org level** as of the May 2026 v1 work
+(`MULTI_TENANCY_DESIGN.md`, migrations `2026_05_19_*`). Concretely, ✅ validated in code:
+
+- `organizations` table (`src/db_models/organization_models.py:10`), seeded `default`
+  (`…0001`) and `system` (`…0002`) orgs.
+- `tasks.org_id` (NOT NULL) and `api_keys.org_id` (nullable; `NULL` = admin/cross-org)
+  (`src/db_models/task_models.py:29`, `src/db_models/auth_models.py:34`).
+- `org_id` denormalized onto the hot tables (rule results, feedback, spans, trace_metadata,
+  annotations) for single-hop tenant filtering.
+- `TENANT-USER` role + four enforcement patterns (path / resource-id / query-param /
+  admin-only) across **60 path endpoints, 11 query endpoints, ~35 resource-id repo methods**.
+- Public, feature-flagged signup at `POST /api/v2/tenant/signup`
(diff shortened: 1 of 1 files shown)


genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
```diff
@@ -0,0 +1,294 @@
...
+  local installs keep working on admin API keys with **no RBAC**; connecting to the platform is
+  what unlocks org-scoped RBAC.
+
+---
+
+## 2. Current State — Validated Findings
+
+### 2.1 Authentication: we build a new Keycloak-based auth solution
+
+We are **building the auth solution end-to-end**. The engine becomes a **Keycloak client** of
```
mcgrawia, line 95: @fryz one thing I just realized about the KC client aspect...

previously I had been thinking all engines would use a single, public KC client ID to SSO users with the SaaS keycloak, but that won't work because clients need to have a whitelist of specific redirect URLs to allow returning to. We'd have to basically have a wildcard redirect whitelist for this client, since we don't know all the addresses of all possible customer hosted engines in advance. This would be pretty dangerous because any website on the internet could theoretically redirect users to KC, and retrieve their tokens if they have an established session.

That means we probably need to build something in the engine registration flow that creates a dedicated client in KC for each engine that would require the engine's redirect URL as part of the request, so we could set the redirect whitelist for that client.

Maybe there's a way to do this as part of the existing ML engine client? Not sure if we could let that client ID also auth users, so both the ML and GenAI engines would share a single client ID, but the bigger part is now we need some sort of registration flow on the platform side for the GenAI engine in addition to what we have for the ml engine.

genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
```diff
@@ -0,0 +1,294 @@
...
+- Read by `get_org_scope(request) -> Optional[UUID]` (`src/dependencies.py:331`).
+- Enforced by `@enforce_org_scope` (path, 60 endpoints) and `@enforce_query_org_scope`
+  (query, 11 endpoints) in `src/utils/users.py`, plus ~35 repository `get_X_by_id(…,
+  org_scope)` methods (Pattern C) that filter `WHERE org_id == org_scope`.
+- `None` means "admin/cross-org" (filter skipped).
+
+The filters are **scattered, not centralized** — every repo method inlines its own
+`if org_scope is not None: q = q.filter(... == org_scope)`. There is no single choke point.
+
+**Feasibility of scalar → set (`org_id IN (...)`):** mechanically straightforward per call
```
mcgrawia, line 126: @fryz just to confirm.. if we go this approach it seems like the genAI engine won't have a UX that distinguishes between orgs for the user to choose. Just wondering if this will be strange if the user enters the engine UI from a project (with a subset of applications/tasks), then in the engine UI sees all tasks from all orgs they have access to, and is confused? Today there wouldn't be any way to easily see/filter which tasks came from which project.

Additionally, the user would see all tasks across all orgs/projects they have access to, but the API key the ML engine uses for jobs would be limited to just the single org it's permitted to access. So the user would see one unified set of tasks in the UI, but the "list tasks" job in the engine would only report a subset of them as "onboard-able". This feels like it would be odd UX. 

Since the user enters the genAI UI via a project, where tasks across projects are already separate, I would expect the genAI UI to also be segregated by project/org to keep the same UX as the platform (with an org selector to hop between)

mcgrawia: The above was focused more on UX, but I think we could also support the segregated task lists via the set `(org_id IN (...))` paradigm, we'd just make the list tasks API have a filter for a specific org ID, and the UI could provide UX to let users choose what org they want to view tasks for, but the auth would technically let them access any task via API across all orgs they have access to. It would just be purely a FE filtered view that gives the illusion of per-org access

I think this would also solve the URL problem below. The FE could just open the /tasks/id page, and the users token would give them access without having to specify an org. It would only be at the list tasks page the org filtering is applied

genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
```diff
@@ -0,0 +1,294 @@
...
+Each subsection is one change at a high level, followed by what has to happen for it to work.
+These are not ordered or phased — they are the full set of moving parts. Items tagged ❓ live in
+the platform / control-plane repo (not in this workspace) and are dependencies, not engine work.
+
+### 3.1 New Keycloak auth client in the engine
+
+Replace the engine's authentication entry point with a new Keycloak-client implementation so the
+engine validates platform-minted tokens and can drive SSO.
+
+- Register the engine as a Keycloak client of the platform's Keycloak.
```
mcgrawia, line 188: this is what I was referring to above will need some more design

genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
```diff
@@ -0,0 +1,294 @@
...
+Make the engine UI show everything the user is authorized for across all their orgs at once.
+
+- Make the UI a Keycloak client: redirect to platform login when unauthenticated; keep the token
+  in auth-module memory with silent refresh.
+- Extend the existing `GET /users/me` + `AuthContext` plumbing to carry the user's set of orgs
+  and drive the admin/tenant render branch.
+- Add an `org_id` discriminator to list/aggregate response payloads (the v1 work already did this
+  for tasks; extend it to the other list/aggregate endpoints) so the UI can group and label
+  cross-org results.
+- Do not introduce an org-switcher component.
```
mcgrawia, line 242: this was the UX I was mostly debating in the second thread above

genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
```diff
@@ -0,0 +1,294 @@
...
+- ❓ Platform must expose the project-creation endpoint that performs the SpiceDB registration.
+
+### 3.6 Onboarding & ownership proof for existing standalone engines
+
+Let an engine that has been running API-key-only connect to the platform and prove it owns its
+orgs.
+
+- On connect, require the engine's admin key as proof of ownership (an admin key implies a
+  single-tenant engine, which no SaaS tenant could hold), and use it to link the engine to the
+  workspace.
```
mcgrawia, line 262: curious how this will work when each org's access is controlled via an org-scoped API key that will be in a connector in the corresponding project in the platform? Will the bootstrap process create an API key for each org using the admin key, then put it in a connector in each project?

genai-engine/docs/PLATFORM_CONNECTED_MULTI_TENANCY_PLAN.md
ntatsumi: Why not introduce an org-switcher component? I think it's worth the effort. It'll be important for at least a couple of reasons:
* Get a view that doesn't cross dev, qa, and prod for STS customers
* Arthur staff to be able to stay within the demo scope for when presenting demos

PR #1814, opened 2026-06-18
docs: Model Inference Service extraction strategy

videetparekh:
Migrates the **Model Inference Service (MIS) — Extraction Strategy** design doc from Notion into `genai-engine/docs/`, so it is version-controlled and reviewable in-repo. This was requested directly on the original Notion page (to track history, leave comments, and expose it to agents in the codebase).

**Source:** [Notion — Model Inference Service — Extraction Strategy](https://www.notion.so/arthurai/Model-Inference-Service-Extraction-Strategy-3821d37e9a108079aa7acfe96e6c4d69)

The review comments left on the Notion page have been carried over as inline review comments on this PR, anchored to the lines they originally referenced, with attribution to the original author (the comment bodies note who said it, since they can't be posted under their accounts here).


videetparekh (commented):
Review comments carried over from the original Notion page, anchored to the lines they referenced. Attribution to the original author is noted in each comment (they can't be posted under their own accounts here).

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+---
+
+## 1. Model Inference Service API
+
+### Endpoint Design
+
+Single unified inference endpoint. Model name in the payload routes to the correct loaded model.
+
+```plain text
+POST /infer
```
videetparekh: **Ian McGraw** (2026-06-18 19:33 UTC, migrated from Notion):

> is this meant to be “inference”?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+```
+
+**Request:**
```
videetparekh: **Ian McGraw** (2026-06-18 19:35 UTC, migrated from Notion):

> wondering is there a standard ML inference API request we can use instead of rolling our own? I know tensor flow and sage maker all have some sort of inference endpoint, wondering if we can keep the same contract so we can point this at models hosted in AWS/GCP just by switching the endpoint?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+```json
+{
+  "model": "pii",
```
videetparekh: **Ian McGraw** (2026-06-18 19:36 UTC, migrated from Notion):

> Is there a reason this is contained in the body instead of as part of the path?
>
> I would expect AWS/GCP models to be served at some resource path like /models/nameorid/infer
>
> One reason I could think that we may not want it in the path is if the model names are not URL safe strings, but seems like we control them so we can force them to be?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+  "config": {}
+}
+```
+
+**Response envelope:**
+
+```json
+{
+  "model": "pii",
+  "result_type": "entities",
```
videetparekh: **Ian McGraw** (2026-06-18 19:37 UTC, migrated from Notion):

> what is the set of result_types we’ll support?

---

**Ian McGraw** (2026-06-18 19:38 UTC):

> ah see this below.. I think another perk of dedicated paths is we can make strong request/response contracts per model, and type-enforce requests/responses for each rather than relying on one generic endpoint to handle all formats

---

**Ian McGraw** (2026-06-18 19:39 UTC):

> for example, “entities” in this case seems to refer to a text substring, but what happens in the future if we start supporting image entities with bounding boxes?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+GET /models        → list of available model names + metadata
+```
+
+### Parallelism Model
+
+Two levels of parallelism, both already present in the engine today:
+
+**Engine side (unchanged):** `scorer/score.py` fans out guardrail checks via thread pool. Each check becomes a parallel HTTP POST to the MIS instead of a local function call. No changes needed to the parallelism logic — just swap the transport.
+
+**MIS side:** Each model is loaded once into memory at startup. The serving framework handles concurrent inference requests. For GPU: requests to the same model batch automatically. For CPU: thread-pool workers per model.
```
videetparekh, line 123: **Ian McGraw** (2026-06-18 19:41 UTC, migrated from Notion):

> is this something we implemented? I thought with PyTorch you had to batch the requests before calling `.predict()` ? if we implemented cross-request batching, how long do we wait for other requests before submitting the batch?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+Target latency per inference call to MIS: **20–80ms** (local network). Parallel fan-out means wall time is bounded by the slowest single model, not the sum.
+
+### Off-the-Shelf Options (OSS)
```
videetparekh, line 127: **Nori Tatsumi** (2026-06-17 19:49 UTC, migrated from Notion):

> Which ever solution we land on, I’d like to make sure that the models can be downloaded in parallel. One of the reasons Shield takes forever to pull is it’s because one layer contains everything in it.

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+**Triton Inference Server (NVIDIA)**
+
+Most production-grade option. Dynamic batching, multi-GPU, ONNX/TensorRT support, Prometheus metrics built in. Significant operational complexity — models need to be converted to Triton-compatible formats (ONNX or TorchScript). Overkill for six small classifiers but the right long-term answer if GPU utilization optimization becomes critical.
+
+**TorchServe**
+
+PyTorch-native, supports batch inference, has a model store concept. Middle ground between FastAPI and Triton. Less community momentum than BentoML currently.
+
+**Recommendation:** Start with FastAPI (fast to ship, zero new dependencies). Migrate to BentoML when you need proper model versioning and batching. Triton only if GPU throughput becomes a bottleneck.
```
videetparekh: **Ian McGraw** (2026-06-18 19:44 UTC, migrated from Notion; mentioned **@Zach Fry**):

> @Zach Fry curious your thoughts on this one.. IMO I'd rather start with something Off the shelf like BentoML if it has a strong community rather than roll our own. It seems like the optimized serving and batching efficiency would be hard for us to match rolling our own in the short term. I don’t know much about it but sounds like it’s at least worth looking into to see how easily it fits into what we need. We can’t be the first company to need a plain model server

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+## 2. Deployment Strategy
+
+### a. Self-Hosted
+
+Two sub-cases, same underlying mechanism.
+
+**Docker pull/run (OSS, no Arthur connection):**
+
+Users run two containers. The engine image is now CPU-only and ships without model weights (~500MB vs. ~8GB today). The MIS image ships code only; weights are mounted at runtime from the existing `model-upload` distribution (S3/GCS/PVC/EFS), so there is no runtime HuggingFace download and no fat image to ship.
```
videetparekh, line 161: **Ian McGraw** (2026-06-18 19:46 UTC, migrated from Notion):

> if there is no runtime HuggingFace download, how do users running this locally get the models? Does it pull them from S3? if so, how do they get them into their S3 bucket?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+# Private inference: run MIS locally (all data stays on-prem)
+bash <(curl -sSL https://engine.arthur.ai) \
+  --arthur-client-id=<id> \
+  --arthur-client-secret=<secret> \
+  --arthur-api-host=<host> \
+  --private-inference
+```
+
+Without `--private-inference`: script starts only the engine container, sets `MODEL_SERVICE_URL` to Arthur's hosted MIS endpoint. No GPU required on the customer machine.
```
videetparekh, line 200: **Ian McGraw** (2026-06-18 19:48 UTC, migrated from Notion):

> seems like if we do this, we’ll need to make our model service public and open to the internet? is there any worry doing that for security or cost?

---

**Ian McGraw** (2026-06-18 19:56 UTC):

> I imagine we’ll need some sort of auth on this otherwise customers will be getting free GPU guardrails on our dime. We’ll need to think through how we grant self-hosted engines access to this MIS, and how we control how much load each gets. It’s also a question of how do we make getting this authentication easy in the deployment scripts? it seems like the access credential will need to come from our saas platform

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+With `--private-inference`: script pulls and starts both containers locally, wires `MODEL_SERVICE_URL` to localhost automatically. GPU required.
```
videetparekh, line 202: **Ian McGraw** (2026-06-18 19:48 UTC, migrated from Notion):

> does the MIS require GPU? What if users want to run private-inference locally on a machine without one?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+
+Arthur hosts one MIS deployment (or a small number by region). N engine deployments — either per-tenant or shared multi-tenant engines — all route to the same MIS.
+
+MIS is stateless compute only. No tenant context stored. Engine passes inference text, receives scores, handles all persistence and tenancy logic itself.
+
+Auth between engine and MIS: reuse the existing shared-secret pattern already in the codebase. Today `ml-engine` authenticates to `genai-engine` via `GENAI_ENGINE_INTERNAL_API_KEY` (matched against the engine's admin key), wired through docker-compose, CloudFormation, and Helm secrets. MIS adopts the same model — a shared internal secret injected via the existing secrets path. Per-tenant token scoping is explicitly out of scope for now: MIS holds no tenant context, so a single shared secret per MIS deployment is sufficient. (mTLS or a token-issuing service can be revisited later if MIS ever needs to attribute or rate-limit per tenant.)
+
+This is the primary cost driver for the extraction: GPU capacity shared across all tenants rather than provisioned per deployment.
+
+### d. Vertex AI
```
videetparekh, line 246: **Nori Tatsumi** (2026-06-17 19:41 UTC, migrated from Notion):

> This would introduce yet another deployment model that differs significantly from our standard container-based approach. I’m not seeing enough benefit to justify adopting and supporting an additional technology stack, especially one that none of our on-prem customers are likely to use.
>
> I recommend running MIS on Cloud Run.
>
> In terms of the discount, how much would we save on Vertex AI vs. CloudRun? I have a feeling that Vertex AI charges premium as the baseline.

---

**Ian McGraw** (2026-06-18 19:53 UTC):

> I don’t know a ton about vertex AI, but is there a provider that would let us deploy these models in a server less fashion so we’re only paying for the GPU while we’re serving a request? I imagine most of the time this service is going to be idle, it would be nice not to have to pay for constant GPU allocation if we only need it sparingly for infrequent requests

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+### b. On-Prem (CFT / Helm)
+
+**CFT (ECS):**
+
+Two separate ECS services. Engine service uses CPU task definition only — the GPU task definition is eliminated from the engine and moves to MIS.
+
+- MIS: GPU ECS service, internal ALB, min 1 task (never scale to zero — cold start on GPU is 30–60s)
+- Engine: CPU ECS service, scales horizontally via ECS scaling policies
+- Routing: engine resolves MIS via internal ALB DNS, no public exposure
+- Auth: IAM task roles — engine task role gets `execute-api:Invoke` on MIS ALB
```
videetparekh, line 213: **Ian McGraw** (2026-06-18 19:50 UTC, migrated from Notion):

> looks like in this case we’d have to plumb something in the engine container to retrieve and pass AWS Creds when calling the model server? what authentication options are we going to support for the model server?

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+| Self-hosted, connected | Arthur (hosted) | Yes | Arthur | Arthur ↔︎ Customer |
+| On-prem CFT/Helm | Customer (their cloud) | No | Customer | None |
+| SaaS multi-tenant | Arthur | Yes (Arthur infra) | Arthur | Arthur ↔︎ Customer |
+| Vertex AI | Google (via Arthur) | Yes (GCP) | Google/Arthur | Arthur + GCP DPA |
+
+---
+
+## Open Items
+
+- Decide default mode for bash script (Arthur-hosted MIS vs. `--private-inference`). Recommendation: default to private, Arthur-hosted as opt-in with explicit disclosure.
```
videetparekh, line 312: **Tal Erez** (2026-06-17 18:06 UTC, migrated from Notion):

> agreed

---

**Ian McGraw** (2026-06-18 20:00 UTC):

> assuming we figure out the comment above about access/cost control, I would actually say we should make Arthur hosted the default to make it easier to get running and try out the platform. We’ve seen a similar issue not having a saas hosted engine, users turn away because the set up is too difficult. Any sort of install script should be easy-first so users try out the product, then have options to self host when they know they’re serious about buying

---

**Tal Erez** (2026-06-18 20:14 UTC):

> hmm that makes sense, especially if we’re mainly switching this for ease of access

genai-engine/docs/model-inference-service-extraction.md
```diff
@@ -0,0 +1,281 @@
...
+- Define circuit breaker behavior in the engine when MIS is unreachable: fail-open (skip ML checks, pass the inference) vs. fail-closed (block). Likely differs by rule type — fail-open on toxicity, fail-closed on PII.
```
videetparekh, line 313: **Tal Erez** (2026-06-17 18:05 UTC, migrated from Notion; anchored on “pass the inference” / fail-open):

> prob wouldn’t expect the inference to pass if MIS is unreachable

videetparekh, line 313: **Ian McGraw** (2026-06-18 20:01 UTC, migrated from Notion; anchored on “fail-open on toxicity, fail-closed on PII”):

> I think this needs to be configurable. Different apps will have different requirements on tolerance for failure

genai-engine/docs/model-inference-service-extraction.md
videetparekh: @mcgrawia  Typically, the API is supposed to be `infer` not `inference`. There's a couple of ways to lay this out, and no strong preference to either:
```
POST /infer
{"model": "pii", "data": ...}
```

or
```
POST /models/pii/infer
{"data": ...}
```

genai-engine/docs/model-inference-service-extraction.md
videetparekh: @mcgrawia Bento, Torchserve and Triton are all exactly that, and also give us parallelization and batching out of the box.

genai-engine/docs/model-inference-service-extraction.md
videetparekh: See above, there's a couple of standards of doing this. Totally fair that we want model name in the path rather than the body.

genai-engine/docs/model-inference-service-extraction.md
videetparekh: I think result type might be a hallucination.

genai-engine/docs/model-inference-service-extraction.md
videetparekh: You do have to batch requests, but the requests can be a batch of 1. This applies to any service, not just pytorch. We technically batch in the engine as well. 

That doesn't really use the GPU in the best way possible, GPUs can really scale throughput and especially in an MT service, we should absolutely be using a batch size of 128/256 if we have the scale to do so.

genai-engine/docs/model-inference-service-extraction.md
videetparekh: Agreed.

genai-engine/docs/model-inference-service-extraction.md
videetparekh: Updated to path-based

genai-engine/docs/model-inference-service-extraction.md
videetparekh: Removed.

genai-engine/docs/model-inference-service-extraction.md
videetparekh: I actually lean towards using off-the-shelf as well. The "recommendation" is a remnant from a different change. Removing.

PR #1893, opened 2026-07-08
GenAI Engine model PVC: online/offline toggle + chart install fixes for EKS Auto Mode

ntatsumi:
## Summary

Makes the model PVC a clean **online/offline model-loading toggle**, fixes several gaps found cross-referencing the EFS/PVC work against a real EKS install, and documents **GPU deployment on EKS Auto Mode**. Split out from #1786 (which is Terraform-only).

`modelPVC.enabled` is the single online/offline switch:

| `modelPVC.enabled` | Mode | Behavior |
|---|---|---|
| `false` (default) | **Online** | Download models from Hugging Face at startup (or `modelRepositoryURL`); no PVC. |
| `true` | **Offline** | Mount the pre-populated PVC, set `MODEL_STORAGE_PATH` + `HF_HUB_OFFLINE=1`, load with no network. |

## Changes

**Engine**
- `download_models()` (`model_load.py`) early-returns when `HF_HUB_OFFLINE` is set, so offline mode makes no network calls or write attempts into the shared volume. Added unit tests (`tests/unit/utils/test_model_load.py`).

**Model PVC / read-write fix**
- `modelPVC.readOnly` now defaults to **`false`**. A read-only mount crashes with `[Errno 30] Read-only file system` because the HuggingFace loaders write `.lock`/cache files under the mount even offline. Corrected all docs (chart README, model-upload README).

**GPU deployment topology (Auto Mode support)**
- The chart supports two GPU topologies, selected by `genaiEngineDeploymentType`. Every GPU-related value (`genaiEnginePodNodeSelector`, `genaiEnginePodTolerations`, deployment-type block, `arthurGenaiEngineHPA.enabled`) is now commented per **three labeled cases**: CPU deployment / GPU with managed node group + ASG / GPU with EKS Auto Mode + Karpenter.
  - **Managed node group + ASG → DaemonSet** (unchanged default; one pod per GPU node, ASG scales nodes).
  - **EKS Auto Mode / Karpenter → Deployment**, because Karpenter provisions GPU nodes only for unschedulable *workload* pods, not DaemonSets ([Karpenter FAQ](https://karpenter.sh/docs/faq/), [EKS Auto Mode](https://docs.aws.amazon.com/eks/latest/userguide/automode.html)).
- New README section **"How to configure EKS Auto Mode for GPU (Karpenter)"** with a ready-to-apply GPU `NodePool` manifest (GPU instance families, `capability: gpu` label, `nvidia.com/gpu` taint) and the exact `values.yaml` settings.

**Umbrella chart** (`deployment/helm/arthur-engine/values.yaml.template`)
- Surfaced `modelPVC`, `probes`, and `transferEncodingMiddlewareEnabled`, and mirrored the labeled GPU-topology comments.

**Schema fixes**
- `genaiEnginePodTolerations` now allows `operator: Exists` + `effect: NoSchedule` so a tainted GPU NodePool can be tolerated via values (no `kubectl patch`).
- `genaiEngineIngressURL` / `scopeFeIngressURI`: `oneOf` → `anyOf` so an empty string validates (`format` is unenforced, so `""` matched both branches and failed `oneOf`).

**Probes**
- Optional startup + readiness probes on `/health` (default on) to gate the load balancer until workers finish loading models — avoids 502s during cold start.

## Testing

- `helm lint` + `helm template` (subchart and umbrella-merged): default off renders no PVC volume; `modelPVC.enabled=true` renders the read-write mount + `MODEL_STORAGE_PATH`/`HF_HUB_OFFLINE`; the Auto Mode GPU config renders `kind: Deployment`, `nodeSelector: capability: gpu`, and the `Exists`/`NoSchedule` toleration; empty `scopeFeIngressURI` passes schema.
- `pytest -m unit_tests`: new `model_load` tests pass; full genai-engine unit suite + 79% coverage pass.

## Related

- Pairs with #1786 (Terraform module for the EFS-backed `ReadWriteMany` PVC; access point pinned to uid `65532`).


ntatsumi:
Autoscalers add nodes in response to pending pods — pods the scheduler can't place because no existing node has room. That pending pod is the demand signal.

Here's the catch with DaemonSets: a DaemonSet pod is never "pending waiting for a node." The DaemonSet controller only creates a pod for a node after that node already exists and matches the DaemonSet's selectors. If there's no GPU node yet, the DaemonSet simply doesn't create a pod anywhere — so there's no unschedulable pod, so there's no signal telling the autoscaler "I need a GPU node."

This is true of both the classic Cluster Autoscaler and Karpenter. Neither will spin up a node just because a DaemonSet would like to run on it.

Why the two topologies behave differently

Managed node group + ASG (classic EKS):
- The ASG creates GPU nodes. That's driven externally — you set the desired count, or scheduled scaling adjusts it, or Cluster Autoscaler bumps it for some other workload.
- Because something else creates the nodes, the DaemonSet-per-node pattern works great: as the ASG adds a GPU node, the DaemonSet controller drops one genai-engine pod onto it; as it scales down, the pod goes away with the node.
- The node's existence does not depend on the genai-engine pod itself.

EKS Auto Mode:
- There are no node groups or ASGs. Karpenter (managed, built into Auto Mode) is the only thing that creates nodes, and it does so strictly in response to pending pods.
- A DaemonSet produces no pending pod → Karpenter sees no GPU demand → it never launches a GPU node → the DaemonSet has nowhere to land → you get zero GPU genai-engine pods.
- So on Auto Mode you need a Deployment whose pod spec requests nvidia.com/gpu: 1. That pod goes Pending when no GPU node exists → Karpenter sees "a pod needs a GPU" → provisions a GPU node → the pod schedules onto it. The pod's resource request is the demand signal that bootstraps the node.

jdbarthur (approved):

PR #1906, opened 2026-07-10
feat(genai-engine): add model-invocation view to the audit log

ntatsumi:
## Summary

The audit log records **who called which endpoint** (the access trail), but not the engine's **own AI processing**. This PR adds an opt-in extension: when `AUDIT_LOG_INCLUDE_AI_ACTIVITY=true`, each authenticated audit entry is annotated with the model invocations made while serving that request.

- **Hosted LLM judge calls** — hallucination, sensitive-data, relevance, tool-selection
- **On-box ML classifiers** — prompt injection, toxicity, PII (v1 + v2)

Records are **metadata only** (model name, provider, token counts, latency, success/error) — **never** prompt/response content, consistent with the existing audit log's no-payload design.

## Toggle

New env var `AUDIT_LOG_INCLUDE_AI_ACTIVITY` (default **`false`**, opt-in). Only consulted when `AUDIT_LOG_ENABLED` is on.

- **Off** (default): the base `AuditLog` (`ArthurAuditLogEventV1`) is emitted **unchanged** — existing consumers see no difference.
- **On**: an `ExtendedAuditLog` (`ArthurAuditLogEventV2`) with a `model_invocations[]` array is emitted.

## How it works

Model invocations fire deep in the scorer — often inside `TracedThreadPoolExecutor` worker threads — while the audit entry is assembled in `BaseHTTPMiddleware` after the request completes. A request-scoped `ContextVar` bridges the gap:

- **`monitoring/ai_activity.py`** (new): the `ContextVar` accumulator, a `record_model_invocation()` that is a **no-op when disabled or outside a request**, and a `track_ml_model_invocation` decorator for local models.
- **The audit middleware starts the accumulator in its own context *before* `call_next` and reads it back *after*.** Starlette copies that context into the `call_next` task, so the scorer appends to the same list the middleware later reads. (Setting it inside a dependency does **not** work — `ContextVar` mutations made *inside* `call_next` don't propagate back out to the middleware. This was caught by the end-to-end test below.)
- **`TracedThreadPoolExecutor`** now copies the `contextvars` context into worker threads (workers only append to the shared list, never reassign).
- **LLM calls** recorded at the single choke point `LLMExecutor.execute()`; **local models** via a one-line decorator on each classifier's `score()`. Instrumenting at the model-call layer (not the rule layer) avoids double-counting.
- All recording is **best-effort** — wrapped so a failed audit record can never break an evaluation.

`model_name` is resolved for **every** LLM call: directly from the model for hallucination/sensitive-data, and via a defensive `extract_chain_model_name()` helper for the relevance/tool-selection LangChain chains.

## Verification — actually run end-to-end

`tests/unit/monitoring/test_ai_activity_e2e.py` drives the **real app** (real audit middleware → auth → rules engine → real decorated scorer) with a live `POST /validate_prompt` that runs the **DeBERTa prompt-injection model**, then asserts the recorded invocation in the on-disk `audit.log`:

```json
{
  "model_name": "ProtectAI/deberta-v3-base-prompt-injection-v2",
  "provider": "local", "model_type": "ml_classifier",
  "operation": "prompt_injection", "latency_ms": 685, "success": true
}
```

## Coverage notes (inherent to a request-scoped log)

- Model calls made **outside a request** (background continuous-eval runs) are not attached.
- The audit-skipped `/completions` and chatbot `/stream` endpoints are not annotated.

## Testing

- `test_ai_activity.py` — no-op-when-disabled, record/read, reset-on-request-end, **thread-pool propagation**, local-model decorator (success + failure).
- `test_ai_activity_e2e.py` — the real-app proof above.
- `test_llm_client.py` — `extract_chain_model_name` for legacy + structured chains + `None` fallback.
- `helpers.py` — `read_audit_entries` made version-aware (handles mixed V1/V2 logs).
- **Full `unit_tests` suite: 1928 passed, 7 skipped.** `black`/`isort` clean; `mypy` adds zero new errors.

## Docs

`docs/audit-logging.md` updated with the new env var, the V2 event, the `model_invocations` schema + example, and coverage notes.


genai-engine/src/monitoring/ai_activity.py
```diff
@@ -0,0 +1,128 @@
...
+
+def begin_ai_activity() -> Optional[Token[Optional[list[ModelInvocation]]]]:
+    """Start collecting model invocations for the current request, if enabled.
+
+    Must be called from the audit middleware **before** ``call_next`` so the
+    downstream app inherits this context. Returns a reset token to pass to
+    :func:`end_ai_activity`, or None when the feature is disabled.
+    """
+    # Imported lazily to keep this module off the scorer's hot-path import chain.
+    from config.config import Config
```
notthattal, line 48: I think I'm confused by the comment. Why does this needs to be inside the function?

ntatsumi: It's just explaining why the import is inside the function body. The mechanism allow this import to only run when begin_ai_activity is called. When the AI activity logging is not enabled, this never gets imported.

notthattal (approved):

PR #1990, opened 2026-07-24
UP-4695 Fix showing alert when leaving dataset flow with unsaved changes

jan-lewandoski-blazity:
## Summary

PR #1934 added an "unsaved changes" confirmation when leaving a dataset with configured-but-unsaved columns/rows. That guard only covered the dataset's own controls (the in-page Back button and version switching) — leaving via the left sidebar (section links or Back to All Tasks) navigated away immediately and silently discarded the work.

This PR closes that gap so the same confirmation appears for sidebar navigation as well.

## Why the previous approach didn't cover it

The sidebar lives in TaskLayout, outside the dataset's context and its <Outlet/>, so its navigate() calls never went through the dataset's guard. React Router's useBlocker isn't available either, because the app uses BrowserRouter (not a data router).

## Approach

Introduce a small, reusable NavigationGuardContext provided in TaskLayout (wrapping both the sidebar and the routed page):

- The provider owns all the bookkeeping — the deferred navigation and the "is a prompt showing" flag. A page doesn't manage any of that itself.
- A page only needs to (1) register a blocker predicate (registerBlocker(() => hasUnsavedWork)) and (2) render its own confirmation modal from the context's isBlocking flag, wiring the buttons to confirmNavigation / cancelNavigation.
- TaskLayout routes both sidebar section links and "Back to All Tasks" through runGuardedNavigation(...). When a blocker is active, the intended destination is deferred and the page's modal is shown; on confirm, navigation proceeds to the originally intended destination.

The dataset view's in-page Back button now funnels through the same guard too, so all exit paths share one confirmation flow and message.

## Preview

https://github.com/user-attachments/assets/a588644c-4dbc-478c-bd1f-75f5e85c5964


genai-engine/ui/src/components/datasets/DatasetDetailView.tsx
```diff
@@ -490,9 +488,9 @@ const DatasetDetailViewContent: React.FC<DatasetDetailViewContentProps> = ({ dat
       )}
 
       <ConfirmationModal
-        open={state.confirmation.type === "unsavedNavigation"}
-        onClose={() => dispatch({ type: "UI/HIDE_CONFIRMATION" })}
-        onConfirm={handleConfirmNavigation}
+        open={isBlocking}
+        onClose={cancelNavigation}
+        onConfirm={confirmNavigation}
```
xV3rt, line 496: I think after your refactor we don't need `unsavedNavigation` type anymore, can probably remove it completely from code

genai-engine/ui/src/components/datasets/DatasetDetailView.tsx
```diff
@@ -94,17 +96,13 @@ const DatasetDetailViewContent: React.FC<DatasetDetailViewContentProps> = ({ dat
...
+  }, [runGuardedNavigation, navigate, task?.id]);
 
-  const handleConfirmNavigation = useCallback(() => {
-    dispatch({ type: "UI/HIDE_CONFIRMATION" });
-    navigate(`/tasks/${task?.id}/datasets`);
-  }, [navigate, task?.id, dispatch]);
+  useEffect(() => {
+    registerBlocker(() => hasUnsavedWork);
+    return () => registerBlocker(null);
+  }, [hasUnsavedWork, registerBlocker]);
```
xV3rt, line 105: Do we even need the function wrapper here?

I'd imagine the logic would look something like this:
```
registerBlocker(isDirty: boolean)
```

I'm not saying the current solution is wrong, just finding a difficult time understanding the complexity here. Was there any edge case we needed to cover?

xV3rt: So basically the useEffect would look something like this
```
useEffect(() => registerBlocker(hasUnsavedWork), [registerBlocker, hasUnsavedWork]);
```

xV3rt (dismissed):
This looks good so I'm gonna approve, there is one nitpick regarding unused type and one question regarding the way we register blockers inside useEffect

genai-engine/ui/src/components/datasets/DatasetDetailView.tsx
jan-lewandoski-blazity: Hey @xV3rt, are you referring to the `unsavedNavigation` type in `showConfirmation` inside `/contexts/dataset/actions.ts:117`?

It seems like the entire `datasetActions` const is dead code, it's not referenced anywhere. Do you think we can fully remove it then?

genai-engine/ui/src/components/datasets/DatasetDetailView.tsx
jan-lewandoski-blazity: Agree on removing the function wrapper - it added some complexity without providing much value.

However, when it comes to moving the `registerBlocker()` call outside of the `useEffect` hook - what's the benefit in your opinion? We still need the cleanup, so it seems to me like keeping both mounting and unmounting in one place is just cleaner. What do you think?

genai-engine/ui/src/components/datasets/DatasetDetailView.tsx
xV3rt: Yeah I think we can remove all of it, I thought only one of the types there was redundant

genai-engine/ui/src/components/datasets/DatasetDetailView.tsx
xV3rt: Let's keep it contained inside useEffect like you planned and simplify by removing the function wrapper if possible

xV3rt (approved):

PR #2009, opened 2026-07-28
NO_REF Set up sourcemaps upload for meticulous test coverage purposes

xV3rt:


fryz:
@xV3rt - just want to confirm that setting these source maps will only apply to our dev environments (and our prod envs will still use minified w/o source maps)

jan-lewandoski-blazity (approved):

xV3rt:
@fryz it will apply to none of our builds, not even for the dev/staging builds. It will apply exclusively to the build generated during meticulous CI needed for testing purposes (so the build artifacts are only uploaded to them through CI and not shared anywhere publicly). `GENERATE_SOURCEMAPS` env variable controls that, so as long as we don't set it to `true` during any of those build pipelines, nothing will change for them

PR #2011, opened 2026-07-28
UP-4699 Add optional whitelist for models in provider configuration

jan-lewandoski-blazity:
[UP-4699](https://arthurai.atlassian.net/browse/UP-4699)

### Problem

The model dropdowns list every model LiteLLM knows for a provider, not what the customer
uses. This PR adds a whitelist for model names, so that users can select which models will appear in dropdowns when selecting given model provider.

### Approach

Nullable `model_whitelist` JSON column on the existing `secret_storage` provider row.
`NULL` = unfiltered. Filtering lives in `ModelProviderRepository.list_models_for_provider`,
so all six consumers of `available_models` narrow together.

Admins curate under Settings → Model Providers → *Visible models*.

**Upgrade is a no-op** — the column is nullable with no backfill, so existing providers
read `NULL`, show `All models`, and behave exactly as before.

### API

- `GET /api/v1/model_providers/{provider}/model_whitelist` → `{whitelist, catalog}`
  (catalog included so the editor can re-add removed models)
- `PUT` same path → `{models}`; `null` clears; rejects unknown IDs and `[]`

`available_models` keeps its response shape and returns fewer entries.

### Preview:

https://github.com/user-attachments/assets/48053bd1-7a2c-4f00-b2df-46d78ca4d41a


genai-engine/src/repositories/model_provider_repository.py
```diff
@@ -322,6 +322,52 @@ def list_model_providers(self) -> List[ModelProviderResponse]:
 
         return providers
 
-    def list_models_for_provider(self, provider: ModelProvider) -> List[str]:
+    def _get_provider_secret(
```
notthattal: could you refactor other functions in this class to use this new function? For example, I see in get_model_provider_client() they are calling the same db call. Also, they seem to be raising the exact same 400 error you are in set_model_whitelist(). Maybe it makes sense to just raise that error inside this function if it's not found?

genai-engine/src/repositories/model_provider_repository.py
```diff
@@ -322,6 +322,52 @@ def list_model_providers(self) -> List[ModelProviderResponse]:
...
+
+    def _catalog_via_client(self, provider: ModelProvider) -> List[str]:
         client = self.get_model_provider_client(provider)
         return client.get_available_models()
+
+    def list_catalog_models_for_provider(self, provider: ModelProvider) -> List[str]:
+        static_catalog = get_static_catalog(provider)
+        if static_catalog is not None:
+            return static_catalog
+        return self._catalog_via_client(provider)
```
notthattal, line 365: wouldn't static_catalog never be none for a supported provider meaning we won't ever reach this line?

genai-engine/tests/unit/routes/model_providers_and_secrets/test_model_provider_and_secrets_endpoints.py
```diff
@@ -558,3 +558,142 @@ def test_setting_azure_provider_credentials(client: GenaiEngineTestClientBase):
     assert response.status_code == 204
 
     db_session.close()
+
```
notthattal, line 561: For all of these tests I would follow convention and create functions inside GenaiEngineTestClientBase like we do for other unit tests instead of repeatedly calling the raw calls in this file.

genai-engine/ui/src/components/model-providers/hooks/useSaveModelWhitelist.tsx
```diff
@@ -0,0 +1,20 @@
...
+  const queryClient = useQueryClient();
+  const { api } = useApi()!;
+
+  return useMutation({
+    mutationFn: async ({ provider, models }: { provider: ModelProvider; models: string[] | null }) => {
+      await api.setModelProviderWhitelistApiV1ModelProvidersProviderModelWhitelistPut(provider, { models });
+    },
+    onSuccess: async (_data, { provider }) => {
+      await queryClient.invalidateQueries({ queryKey: queryKeys.providers.modelWhitelist(provider) });
+      await queryClient.invalidateQueries({ queryKey: ["availableModels"] });
```
xV3rt: Could we use `queryKeys` here as well, just like in line above? We are trying to keep a single source of truth dictionary here to have an easier time managing the keys. Particularly useful in case of invalidation.

genai-engine/ui/src/components/model-providers/index.tsx
```diff
@@ -70,13 +76,34 @@ export const ModelProviders: React.FC = () => {
...
 
     await saveProviderMutation.mutateAsync({ provider: editModal.provider, data });
+
+    if (whitelistDirty) {
+      await saveWhitelistMutation.mutateAsync({
+        provider: editModal.provider.provider,
+        models: whitelist,
+      });
+    }
   };
```
xV3rt, line 114: The problem here is that we close the modal by useSaveProvider's onSuccess, which runs before the whitelist PUT even starts. If somehow the second request fails, the dialog will already be gone and the error vanishes with no feedback for the user. I think we can adjust handleEditSave a little bit and use snackbar to communicate potential error.

```
const handleEditSave = async (data: PutModelProviderCredentials) => {
    if (!editModal.provider) return;

    try {
      await saveProviderMutation.mutateAsync({ provider: editModal.provider, data });
    } catch (err) {
      const detail = isAxiosError(err) ? err.response?.data?.detail : undefined;
      enqueueSnackbar(detail || "Failed to save credentials", { variant: "error" });
      return;
    }

    if (whitelistDirty) {
      try {
        await saveWhitelistMutation.mutateAsync({
          provider: editModal.provider.provider,
          models: whitelist,
        });
      } catch (err) {
        const detail = isAxiosError(err) ? err.response?.data?.detail : undefined;
        enqueueSnackbar(detail || "Credentials saved, but updating visible models failed", { variant: "error" });
        return;
      }
    }

    setEditModal({ isOpen: false, provider: null });
  };
```
We have a hook for the snackbar:
```
const { enqueueSnackbar } = useSnackbar();
```

genai-engine/ui/src/components/model-providers/components/edit-form/components/model-whitelist/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+import { ModelProvider } from "@/lib/api-client/api-client";
+
+type Props = {
+  provider: ModelProvider;
+  providerDisplayName: string;
+  providerEnabled: boolean;
+  value: string[] | null;
+  onChange: (models: string[] | null) => void;
+  onInitialValue: (models: string[] | null) => void;
```
xV3rt: Not a big fan of the concept of `onInitialValue` which sets a state of another component under the hood now.

genai-engine/ui/src/components/model-providers/components/edit-form/components/model-whitelist/index.tsx
```diff
@@ -0,0 +1,139 @@
...
+
+  const { data, isLoading, error } = useModelWhitelist(provider, !catalogUnavailable);
+
+  const seeded = useRef(false);
+  if (data && !seeded.current) {
+    seeded.current = true;
+    onInitialValue(data.whitelist ?? null);
+  }
+
+  const restricted = value !== null;
```
xV3rt: Could we maybe solve this problem by deriving value from a `dirty` prop?
```
  const models = dirty ? value : (data?.whitelist ?? null);
  const restricted = models !== null;
```

xV3rt: This means we could replace `onInitialValue` prop with a `dirty` prop passed as such from model-providers
```
            whitelistDirty={whitelistDirty}
```

xV3rt: And that would allow us to simplify in model-providers
```
const saveProviderMutation = useSaveProvider();
```
(no more state setting in onSuccess inside)

genai-engine/ui/src/components/model-providers/index.tsx
```diff
@@ -70,13 +76,34 @@ export const ModelProviders: React.FC = () => {
...
+
+    setWhitelistDirty(false);
     setEditModal({ isOpen: true, provider });
   };
 
+  // Seeding from the fetch must not mark the state dirty
+  const handleWhitelistInitial = (models: string[] | null) => {
+    setWhitelist(models);
+    setWhitelistDirty(false);
+  };
```
xV3rt: This should become redundant if we derive from `dirty` (described in another comment)

genai-engine/ui/src/components/model-providers/components/edit-form/components/model-whitelist/index.tsx
xV3rt: This will also require some changes to the tests but I think it's worth it overall, much easier to debug/maintain if we are simply passing a boolean prop instead of a function that will manipulate state of another component

genai-engine/src/api_changelog.md
```diff
@@ -4,6 +4,9 @@ The intention of this changelog is to document API changes as they happen to eff
 
 # 07/24/2026
 - **CHANGE** for **URL**: /api/v2/datasets/{dataset_id}/versions/{version_number}/restore  endpoint added
+# 07/27/2026
+- **CHANGE** for **URL**: /api/v1/model_providers/{provider}/model_whitelist  endpoint added
+- **CHANGE** for **URL**: /api/v1/model_providers/{provider}/model_whitelist  endpoint added
```
xV3rt: Probably an issue during rebase, but we want those ordered properly, 27th should end up above 24th

xV3rt: Also linting seems to be failing, you might want to run `black --check` for BE files, and linters for FE changes if needed

genai-engine/src/repositories/model_provider_repository.py
jan-lewandoski-blazity: This condition only serves `hosted_vllm` which does not expose a static list of model providers. Instead, It fetches the list only after providing the base url. Added a comment explaining this case - let me know if that's fine 👍

genai-engine/src/repositories/model_provider_repository.py
jan-lewandoski-blazity: Good idea, all three now go through the shared lookup: `set_model_provider_credentials`, `delete_model_provider_credentials`, and `get_model_provider_client. 

I left `list_model_providers` alone because it has no name filter and selects only .name.

I didn't move the `400` into `_get_provider_secret` itself, though. Two callers need it to return None. `set_model_provider_credentials` uses None to decide insert vs. update, so raising there would make it impossible to configure a provider the first time, and `get_model_whitelist` returns None because the whitelist editor opens during first-time setup, before any secret row exists.

So the raise now lives in a `_require_provider_secret` wrapper, and the two methods that did raise the identical error (`get_model_provider_client` and `set_model_whitelist`) both get it from there 👍

genai-engine/tests/unit/routes/model_providers_and_secrets/test_model_provider_and_secrets_endpoints.py
jan-lewandoski-blazity: Done ✅

genai-engine/ui/src/components/model-providers/components/edit-form/components/model-whitelist/index.tsx
jan-lewandoski-blazity: Good idea! ✅

genai-engine/ui/src/components/model-providers/index.tsx
jan-lewandoski-blazity: Good catch, fixed ✅

jan-lewandoski-blazity:
[Claude's comment](https://github.com/arthur-ai/arthur-engine/pull/2011#issuecomment-5114952454) resolved in [101889d](https://github.com/arthur-ai/arthur-engine/pull/2011/commits/101889dcc3d43dd1f29226d4c625286f6010f265)

genai-engine/src/repositories/model_provider_repository.py
```diff
@@ -116,6 +116,34 @@ def _retrieve_aws_bedrock_credentials_from_secret(
             "aws_session_name": creds_dict.get("aws_session_name", ""),
         }
 
+    def _provider_secret_query(self, provider: ModelProvider) -> Query[Any]:
```
xV3rt: Can't we type it stricter here? Something like `Query[DatabaseSecretStorage]`

genai-engine/src/api_changelog.md
```diff
@@ -97,6 +97,12 @@ The intention of this changelog is to document API changes as they happen to eff
 - **CHANGE** for **URL**: /api/v1/tasks/{task_id}/rag_notebooks  removed '#/components/schemas/SavedRagConfig-Output' from the 'experiments/items/rag_configs/items/' response property 'oneOf' list for the response status '201'
 - **CHANGE** for **URL**: /api/v1/tasks/{task_id}/rag_notebooks  removed '#/components/schemas/SavedRagConfig-Output' from the 'state/rag_configs/anyOf[subschema #1]/items/' response property 'oneOf' list for the response status '201'
 
+# 07/24/2026
+- **CHANGE** for **URL**: /api/v2/datasets/{dataset_id}/versions/{version_number}/restore  endpoint added
+# 07/27/2026
+- **CHANGE** for **URL**: /api/v1/model_providers/{provider}/model_whitelist  endpoint added
+- **CHANGE** for **URL**: /api/v1/model_providers/{provider}/model_whitelist  endpoint added
+
```
xV3rt: Still an issue here with changelog order. We can maybe fold the 24th entry into 27th? Since this wasn't merged yet I feel like just 1 entry in the changelog is correct here, we don't need more than one, even if the PR is actively being worked on for a while

xV3rt: + this entry duplicates what is directly below so I think it can actually be safely deleted instead

jan-lewandoski-blazity: Order fixed - regenerated the changelog from scratch.
 
As for the duplicated entry - there are two endpoints, GET and PUT. Both render identically and the file already has ~30 of these pairs from other multi-method endpoints. I could delete one by hand, but the next person to regenerate would just get it back, so I'd leave the generated output as is 👍

notthattal (dismissed):
BE lgtm

genai-engine/alembic/versions/2026_07_27_1200-b3f2a91c4d17_add_model_whitelist_to_secret_storage.py
```diff
@@ -0,0 +1,31 @@
...
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "b3f2a91c4d17"
+down_revision = "ea519bf96e9c"
+branch_labels = None
+depends_on = None
```
xV3rt, line 20: LGTM, let's just make sure that those migration details (revision, down revision, ids etc) are still correct considering how much time has passed since this file was generated

xV3rt (approved):

genai-engine/alembic/versions/2026_07_27_1200-b3f2a91c4d17_add_model_whitelist_to_secret_storage.py
jan-lewandoski-blazity: Reviewed, it's correct ✅

PR #2012, opened 2026-07-28
chore(renovate): make transitive + uv updates actually flow

ntatsumi:
## Why

We expected Renovate to auto-address the HIGH-severity findings in **Security → Code scanning**, but it wasn't. Those alerts are all from **Trivy image scanning** (advisory-only, per `security/README.md`) — a different surface than Renovate manages. They split into four buckets:

| Bucket | ~Count | Example | Renovate can fix? |
|---|---|---|---|
| Debian OS packages in base image | ~55 | krb5, `libexpat1`, `util-linux` | ❌ floating distroless/slim tags; ~half unfixed → VEX only |
| `quinn-proto` (Rust crate in `uv` binary) | 2 | `/usr/local/bin/uv` | ⚠️ only once the `uv==` pin is tracked |
| Transitive Python deps **with** a fix | ~16 | `pillow` 12.2→12.3, `pyasn1` 0.6.3→0.6.4 | ✅ but only via `lockFileMaintenance` |
| No upstream fix | 1+ | `joserfc` | ❌ VEX only |

The config on `dev` already had the docker managers, `osvVulnerabilityAlerts`, and `lockFileMaintenance` — but it **wasn't producing PRs** for the fixable buckets. This PR closes that automation gap (current VEX triage of the unfixable OS/no-fix CVEs is intentionally left to the team).

## What changed (`renovate.json` only)

1. **Unblock `lockFileMaintenance`** — it was pinned to a once-weekly window (`* 0-4 * * 1`) and sat permanently in "Awaiting Schedule" on Dependency Dashboard #128 with **0 PRs ever**. It's the *only* lever that bumps transitive `uv.lock` deps (`pillow`, `pyasn1`). Now inherits the daily schedule + `prCreation: immediate`.
2. **Track the `uv==` pin** — added a `custom.regex` manager (pypi datasource) covering all Dockerfiles (`genai-engine/dockerfile`, `Dockerfile_changelog`, `ml-engine/Dockerfile`, `deployment/model-upload/Dockerfile`). Also added `"custom.regex"` to `enabledManagers` — required for custom managers to run when `enabledManagers` is set. This is the path for the `quinn-proto`-in-`uv` finding.
3. **Raise throughput** — `prConcurrentLimit: 20` / `prHourlyLimit: 6` so the 60+ update backlog (throttled by `config:recommended` defaults of 2/hr, 10 concurrent) drains.
4. **Label vulnerability PRs** — `vulnerabilityAlerts.labels: ["security"]`.

Validated with `renovate-config-validator` against current Renovate.

## Caveats

- **`quinn-proto` won't clear immediately**: `uv` 0.11.25 still vendors `quinn-proto` 0.11.14; the fix needs a `uv` release bundling ≥0.11.15. The custom manager just makes Renovate propose newer `uv` automatically.
- **OS-package + no-fix alerts remain** (floating base tags / unfixed CVEs) — those stay on the team's VEX path in `security/vex/openvex.json`, out of scope here.

## Verify after merge

1. Tick `renovate/lock-file-maintenance` under "Awaiting Schedule" on Dependency Dashboard #128 → a **"Lock file maintenance"** PR should open bumping `pillow 12.2.0 → 12.3.0` and `pyasn1 0.6.3 → 0.6.4` in the `uv.lock` files.
2. Dashboard "Detected Dependencies" should now list `uv` (pypi).
3. After merge + image rebuild, the next daily `image-vuln-scan.yml` run should drop the `pillow`/`pyasn1` alerts from Security → Code scanning.


ntatsumi:
Good catch — this is correct and I've fixed it in b7920216c.

I confirmed the built-in default independently via `getOptions()`:

```
lockFileMaintenance default schedule: [ "before 4am on monday" ]
```

So removing the key left the weekly Monday default in place (as you described: mergeable object → default merges in first → child `schedule` array overwrites the parent at flatten time, no inheritance). The job would have stayed in **Awaiting Schedule** and the `pillow`/`pyasn1` bumps still wouldn't flow.

Fix applied:

```json
"lockFileMaintenance": {
  "enabled": true,
  "schedule": ["* 0-4 * * *"]
}
```

Also dropped `prCreation: "immediate"` per your note — it's already the default and only governs post-branch PR creation, not schedule gating, so it was misleading noise. Thanks for the precise write-up.

PR #2019, opened 2026-07-28
ci: rebuild model-upload images when their deps/Dockerfile change

ntatsumi:
## Problem

The model-upload image build jobs (`build-genai-engine-models-docker-image` → `build-ecs-model-upload.yml`, and the `-fs` equivalent) only run when `check-models-updates` detects a **new HuggingFace model revision** (or a missing image, for bootstrap). On every other version bump, the `tag-*` jobs simply **re-tag the previous image**.

That means a change to the model-upload **Dockerfile, Python dependencies, or entrypoint scripts** is silently re-tagged onto the *old* image instead of rebuilt — so it never ships to `genai-engine-models-s3` / `-fs`. We hit exactly this: the `uv 0.11.25 → 0.11.26` security bump (quinn-proto GHSA-4w2j-m93h-cj5j) and `pyasn1 0.6.4` landed in source but the routine version bump re-tagged the stale `:latest-dev`, leaving the vulnerable images published. (GCS is unaffected — `build-gcp-model-upload` rebuilds on every bump.)

## Fix

`check-models-updates` now also sets `has_updates=true` when any file **baked into** the model-upload image changed since the previous release build. The build only runs on `Increment arthur-engine version` commits, so the previous such commit is the last built baseline; the job diffs it against `HEAD` over:

- `Dockerfile`
- `pyproject.toml`, `uv.lock` (dependencies)
- `download_models.py`, `upload_models.py` (entrypoint scripts)

Any change there forces a rebuild instead of a re-tag. Requires `fetch-depth: 0` on the checkout for the diff. Reuses the existing "append `has_updates=true` to `$GITHUB_OUTPUT` to override" pattern already used by the missing-image bootstrap check, so the `build` vs `tag` gating is unchanged otherwise.

## Verification

- Workflow YAML validated (parses; pre-commit `check yaml` passed).
- Simulated the new logic against **real history**: the diff over the `2.1.741 → 2.1.742` window flags `deployment/model-upload/Dockerfile` (the uv bump) → would have auto-rebuilt `s3`/`fs` instead of re-tagging. The `2.1.742 → 2.1.743` window (no build-input change) correctly yields a re-tag.

Also updates `security/README.md`, whose "rebuilt only when models change" note was now stale.


ntatsumi:
Good catch — fixed in 98cefd955.

You're right that the baseline grep walked HEAD's whole ancestry, which breaks on `main`. I confirmed the topology against real history: release PRs land on `main` as **merge commits** (2 parents), e.g. `#1980` = `[666f7bb5d (prev release #1923), 30d021b00 (dev tip 2.1.725)]`, and the merge's tree equals the dev tip's tree. So on that release my old `--skip=1 -n 1` picked `30d021b00` and `git diff` was **empty every time** — a total no-op on production, exactly as you described.

Fix is your suggested `--first-parent`, which I verified holds on both branches (merge-commit releases are the current, confirmed strategy):

- **main** HEAD `f0863bcb2` (#1980): baseline moves from `30d021b00` (swept-in dev bump → empty diff) to **`666f7bb5d` (#1923, the previous release)** → correct release-wide window.
- **dev** HEAD `fb2b08d39` (2.1.743): baseline stays **`d81c72b19` (2.1.742)** — unchanged. Confirmed dev bumps sit on dev's first-parent line (`fb2b08d39 → d81c72b19 → 15dbf24eb → f8ca57051`), so `--first-parent` doesn't skip them.

I expanded the inline comment to spell out the dev-vs-main reasoning and flag `--first-parent` as load-bearing. Re your rebase caveat: releases are merge commits today, so this is correct; if the release strategy ever switches to rebase/squash, the baseline would need to anchor to the previous `version` value (or an image-embedded build SHA) instead — noted for follow-up.

Thanks for the detailed review — the merge-commit-tree-equals-HEAD detail was the key insight.

PR #2073, opened 2026-07-30
security: patch base /lib libraries and report true OS package versions

ntatsumi:
## Problem

A customer's Wiz gate fails our released images (reported against `genai-engine-gpu-2.1.748`) with **19 HIGH/CRITICAL OS-package CVEs** — 5 on `libc6`, 6 on `libssl3` (incl. CRITICAL CVE-2026-34182), 2 each across the four `krb5` packages — plus a `liblzma5` MEDIUM.

Investigating turned up **two distinct defects**. My first read was that this was purely stale metadata; that was wrong, and the more serious half is real.

## Defect 1 — the base's `/lib` libraries were never patched (real)

**The distroless base is not merged-`/usr`.** It keeps *real files* in `/lib/<triplet>/`:

- glibc — `libc.so.6`, `libm`, `libpthread`, `libnss_*`, **and the ELF interpreter that `/lib64` symlinks to**
- `liblzma`, `libexpat`, `libgcc_s`, `libbz2`, `libcom_err`, `libcrypt`, `libkeyutils`, `libncursesw`, `libreadline`

`COPY --from=install /usr/lib /usr/lib` never reached any of them. The upgraded glibc landed at `/usr/lib/.../libc.so.6` as a *new* file while the base's `/lib/.../libc.so.6` stayed at **2.36-9+deb12u13**, and `/lib64/ld-linux-x86-64.so.2` → that same unpatched loader.

**So every image shipped so far carried unpatched glibc and liblzma5.** The 5 libc6 HIGHs and the liblzma5 MEDIUM are real findings, and `security/vex/openvex.json` wrongly recorded libc6 as `fixed`.

**Fix:** mirror the base's `/lib` layout using the build stage's apt-upgraded libraries. Only paths the base already has are emitted, so files are *replaced*, not added, and the layout is preserved.

## Defect 2 — stale dpkg metadata (cosmetic, but blocks gates)

`libssl3` and the four krb5 packages *do* live under `/usr/lib` and were genuinely being patched — their shipped `.so` files are **byte-identical (md5)** to the upgraded build stage. But nothing ever touched `/var/lib/dpkg/status.d`, which the base ships declaring `libssl3 3.0.19` and krb5 `u4`. `detectionMethod: PACKAGE` reads exactly there, so those **14 findings are stale metadata**.

VEX already documented this, but VEX only silences **our** Trivy run — it does nothing for a customer's Wiz policy gate, which is why these return every release.

**Fix:** rewrite `status.d` to the versions the build stage actually installed. Only packages the stage installed that ship a `.so` under `/lib|/usr/lib/<triplet>/` are synced — exactly the files the two copies overwrite. Packages the base has but the stage lacks keep their version, since their files survive untouched. `md5sums` sidecars are refreshed from the stage's dpkg database via `dpkg-query --control-path` (dpkg stores them arch-qualified — `libc6:amd64.md5sums` — for `Multi-Arch: same` packages, which all of these are). Every shipped shared object matches byte-for-byte; nine `/usr/share/doc` changelog entries stay stale because these images never copy `/usr/share`, which is inert since no scanner resolves CVEs from checksums.

Applied as identical blocks to `genai-engine/dockerfile`, `ml-engine/Dockerfile`, and both distroless `model-upload` runtimes. `runtime-gcs` is slim-based and already correct.

## Verification

Assembled the final image with both blocks **extracted verbatim from these files**:

| Check | Result |
| --- | --- |
| `GLIBC 2.36-9+deb12u13` anywhere in the filesystem | **0 occurrences** (was the shipped glibc) |
| `/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2` | byte-identical (md5) to the patched build stage |
| All 7 flagged packages in `status.d` | declare the exact `fixedVersion` Wiz requires |
| **Image still runs** | python 3.12.13 imports `ssl`/`lzma`/`sqlite3`, reports **OpenSSL 3.0.20**; bash works |
| python3.11 cleanup | unaffected (sync `COPY` ordered before the `rm`) |
| `docker build --check` | no errors; warning counts unchanged from `dev` |

Replacing the ELF interpreter is the risky part of this change, hence the explicit runtime check.

**Not built end-to-end:** the genai-engine UI stage needs `GITLAB_UNIFY_FRONTEND_TOKEN`, so CI is the first full build.

## Reviewer notes

- **Images build on the next `Increment arthur-engine version` commit** — this PR alone produces no rescannable image.
- **VEX document: the 13 now-patched `fixed` statements are removed** (libssl3 ×6, libc6 ×5, krb5 ×2 across four packages), leaving 28 `not_affected` statements. This restores the convention the doc itself states — *fix, don't justify*: it now lists only findings that are unexploitable in our usage, so a CVE absent from it was remediated, not quietly dropped. The scope note records that the libc6 entries were wrong for prior releases, for auditors.
- **`aiohttp` 3.13.5 → 3.14.1 not bumped**: 7 medium / 4 low, no HIGH, transitive via langchain — fails neither the `severity: HIGH` threshold nor `ignoreTransitiveVulnerabilities: true`. Happy to add it (needs a `uv.lock` regen).
- **Pre-existing, out of scope:** `ml-engine` copies a much larger `/usr/lib` than the base declares (libpq, unixODBC, MS ODBC, Oracle Instant Client), so its SBOM *under*-reports those libraries entirely. Worth a follow-up.

### DIFF
--- a/deployment/model-upload/Dockerfile
+++ b/deployment/model-upload/Dockerfile
@@ -15,6 +15,12 @@
 
 ARG BACKEND=s3
 
+# The distroless base is referenced here so the builder stage can read the dpkg
+# metadata it ships (see the status.d sync at the end of that stage), in addition to
+# the runtime-s3/runtime-fs stages below. The same tag resolves to the same digest
+# within one build, so this adds a metadata read, not a second base image.
+FROM gcr.io/distroless/python3-debian12:nonroot AS distroless_meta
+
 # =============================================================================
 # Stage 1: Download models (shared across all backends)
 # =============================================================================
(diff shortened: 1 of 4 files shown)


jdbarthur (dismissed):

ntatsumi:
## Resolution — confirmed and fixed in ed455dae5

The finding is correct. Fixed by resolving the path through dpkg instead of composing it, as suggested:

```sh
if src="$(dpkg-query --control-path "$pkg" md5sums 2>/dev/null)" && [ -f "$src" ]; then
  cp "$src" "${f}.md5sums"; fi
```

Reproduced the dead branch in a bookworm container first — all seven packages the `.so`-under-`<triplet>` filter selects are `Multi-Arch: same`, and the guard was false on every one:

```
$ ls /var/lib/dpkg/info/ | grep libc6.*md5sums
libc6:amd64.md5sums
$ [ -f /var/lib/dpkg/info/libc6.md5sums ] && echo TRUE || echo FALSE
FALSE
$ dpkg-query --control-path libc6 md5sums
/var/lib/dpkg/info/libc6:amd64.md5sums          # Multi-Arch=same
```

### One correction: the path-normalization caveat doesn't apply

The review states that fixing the path is insufficient because "the build stage *is* merged-`/usr`, so its md5sums list `usr/lib/x86_64-linux-gnu/libc.so.6`, whereas the base is not, and its `status.d` md5sums list `lib/x86_64-linux-gnu/libc.so.6`," concluding that "a straight copy would then produce checksums describing paths that do not exist in the final image."

That premise doesn't hold. Merged-`/usr` is a filesystem-level symlink; it does not change the paths dpkg records in its control files. Both sides use the same `lib/<triplet>/` convention for libc6's `/lib`-resident files:

```
build stage  (dpkg-query --control-path libc6 md5sums):
  108877c29ef6a482337ddc42ea8b0985  lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
  33f1cb6525f26614bd3a334d63632484  lib/x86_64-linux-gnu/libc.so.6

distroless base (status.d/libc6.md5sums):
  a4104e139c611a4e5ae22fc480fdf76c  lib/<triplet>/ld-linux-…
  b7bacb77f7b131d15fbb93ce7eff7d31  lib/<triplet>/libc.so.6
```

So the copy is path-compatible and needs no normalization. That's why I fixed the block rather than taking the suggested alternative of deleting it — deleting would leave every shared-object checksum describing the u13 files this PR replaces, which is the less consistent of the two end states.

### Verification

Built ml-engine for real and checked every entry in the refreshed sidecars against the shipped file:

| Package | entries | match | mismatch |
| --- | --- | --- | --- |
| libc6 | 284 | 282 | 2 |
| libssl3 | 9 | 7 | 2 |
| libkrb5-3 | 7 | 6 | 1 |
| libk5crypto3 | 3 | 2 | 1 |
| libkrb5support0 | 3 | 2 | 1 |
| libgssapi-krb5-2 | 4 | 3 | 1 |
| liblzma5 | 7 | 6 | 1 |

**Zero mismatches on any shared object.** All nine remaining are `/usr/share/doc/*/changelog.Debian.gz` (plus `NEWS.gz`), which stay at the base's copies because these images copy `/lib`, `/usr/lib` and `/usr/local` but never `/usr/share`.

That residue is inert — nothing resolves CVEs from checksums — but it does mean the original comment's "the checksums describe the files that actually ship" was still overclaiming even with the path fixed. Rather than trim the claim to fit, the comment blocks in all three Dockerfiles now state exactly what matches and what doesn't, and why. The PR description had the same overclaim and has been corrected.

`Version:` sync re-verified unchanged after this edit — all six packages still report the `fixedVersion` the scans require, and the three sync blocks remain byte-identical across the Dockerfiles.

PR #2076, opened 2026-07-30
ci: add Claude auto-fix for Renovate PRs that break CI

ntatsumi:
## Why

Grouped Renovate PRs sometimes fail CI because a bumped dependency changed its API — e.g. `zod` type changes and stricter `@tanstack/query` eslint rules broke `npm minor and patch` (#2066). These need real **code adaptation**, not a formatter, so they sit red and block the queue until a human fixes them. This workflow has Claude do that adaptation automatically, safely, and unattended.

Scope: **code-adaptation** breakages only (TS/type errors, eslint, mypy, formatting). Dependency-resolution/lockfile conflicts (e.g. #2064's `litellm` vs `arthur-common` pin) are explicitly **out of scope** — the fixer comments and stops.

## How it works

1. **Trigger:** `workflow_run` on **"Arthur Engine CI" → completed**, filtered to `conclusion == failure` on a `renovate/*` branch. The listener runs from `dev` (default branch), so a PR can't tamper with the fixer that judges it.
2. **Fix:** checks out the PR branch and runs `anthropics/claude-code-action@v1` with a tightly-scoped prompt. Claude reads the failed logs, adapts the source, and **re-runs the real required checks itself**.
3. **Gate:** Claude writes a `.claude-autofix-success` marker only if the full suite passes; otherwise it discards edits and comments. A later step commits/pushes **only if that marker exists** — nothing broken is committed.
4. **Merge path:** pushes as the **Arthur GitHub Automator App**, then re-approves as **github-actions[bot]** (two distinct identities, required by `require_last_push_approval`) → CI re-runs → merge queue merges it.

## "Don't break anything" — three layers
- **Self-verify:** no commit unless Claude confirmed every required check passes locally.
- **Hard backstop:** the pushed fix still must pass all 6 required checks on the PR *and* the merge-queue branch — this workflow can't bypass that.
- **Scope/tool guards:** prompt forbids editing `.github/**`, tests, version pins, lockfiles, and forbids silencing checks (`@ts-ignore`, `# type: ignore`, skips, coverage drops). Loop guard: one attempt per branch head (`[claude-autofix]` commit marker).

## ⚠️ Prerequisites to confirm (please verify before relying on it)
1. **`claude-code-action` under `workflow_run`** — run in agent mode with an explicit prompt. First live run is the real test; if it balks, a tiny `workflow_run → workflow_dispatch` shim fixes it.
2. **Actions approvals allowed** — Settings → Actions → "Allow GitHub Actions to create and approve pull requests" must be ON (github-actions[bot] is the approver). If disallowed, switch the approver to a second app identity.
3. **`shared-protected-branch-secrets` environment** must not require manual reviewer approval (else the job pauses). `version-workflow.yml` uses it unattended, so likely fine.

## Testing plan
After merge, smoke-test on **#2066** (zod/tanstack/prettier): trigger a fresh CI failure (`@renovatebot rebase`) and watch the fixer run. #2064 should be correctly **skipped/commented** (dependency conflict, out of scope).

## Follow-up (not in this PR)
Add a small `renovate.json` `packageRules` constraint so the `litellm`/`arthur-common` conflict (#2064) never goes red in the first place.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -12,7 +12,12 @@ on:
 
 jobs:
   claude-review:
-    # Skip only for Renovate PRs so the job is skipped, not failed.
+    # Review human PRs (skip pure Renovate syncs), PLUS review Renovate PRs the
+    # instant our Automator App pushes a Claude auto-fix to them — so code Claude
+    # wrote on a dependency PR always gets a review pass. The auto-fixer also does an
+    # inline review-and-address gate before it approves; this is the independent,
+    # fresh-context second look.
+    #
     # GitHub App bots use login 'renovate[bot]'; legacy/self-hosted may use 'renovate'.
     # Check both the PR author AND the triggering actor: Renovate can rebase/synchronize
     # a PR opened by another bot (e.g. the version-bump app), which makes github.actor
@@ -20,10 +25,13 @@ jobs:
     # claude-code-action gates on the triggering actor, so guarding only on the author
     # lets the job run and then fail with "Workflow initiated by non-human actor: renovate".
     if: >-
-      github.event.pull_request.user.login != 'renovate' &&
-      github.event.pull_request.user.login != 'renovate[bot]' &&
-      github.actor != 'renovate' &&
-      github.actor != 'renovate[bot]'
+      (
+        github.event.pull_request.user.login != 'renovate' &&
+        github.event.pull_request.user.login != 'renovate[bot]' &&
+        github.actor != 'renovate' &&
+        github.actor != 'renovate[bot]'
+      ) ||
+      github.actor == 'arthur-github-automator-app[bot]'
 
     # Optional: Filter by PR author
     # if: |
--- a/.github/workflows/renovate-autofix.yml
+++ b/.github/workflows/renovate-autofix.yml
@@ -1,8 +1,11 @@
 name: Renovate Auto-Fix
 
 # When "Arthur Engine CI" fails on a Renovate PR, let Claude adapt the source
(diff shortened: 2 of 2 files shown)


jdbarthur (dismissed):
noice

ntatsumi:
Thanks — all three are valid; fixed in cd77d015d.

**1. Privileged job gated on branch name only (security).** Correct and the most important one. Fixed with defense in depth:
- Job `if:` now also requires `head_repository.full_name == github.repository` (rejects fork-originated runs, which is the real danger since this job holds write creds + an approver token) **and** `workflow_run.actor.login == 'renovate[bot]'` (a human pushing a `renovate/*` branch no longer qualifies).
- The resolved PR's author is re-verified as `renovate[bot]` before any push/approve, and the fallback `gh pr list` is constrained to `--author app/renovate`.
- The `.claude-autofix-success` marker is moved **out of the worktree** into `$RUNNER_TEMP`, so it can't be committed to the branch to forge a "verified" signal — closing the `rm -f` → dirty-tree → approve path you described.

**2. `git diff` vs `git add -A` disagree on untracked files.** Correct — a fix that only adds a new file (e.g. an ambient `.d.ts`) would be dropped as "no changes" while the loop guard stayed open. Switched the gate to `git status --porcelain`.

**3. `autoflake` missing from reproduce/fix/verify.** Correct — it's a real job-failing gate. Added its `--check` (step 3) and `--in-place` (step 4) invocations.

The two "not flagged" notes (Actions-approval setting, load-bearing `persist-credentials`) match my understanding and are called out as prerequisites in the PR description.

PR #2083, opened 2026-07-31
fix(aibom): stop stamping the engine version into the committed AI-BOM

ntatsumi:
Fixes the release-PR blockage at its source, replacing the downstream guard in #2082 (now closed).

## The churn

**#1909** added `aibom.cdx.json` to the files the model-upload workflows stage, and `generate_aibom.py` stamps `metadata.component.version` with the engine version. That field changes on every release, so `git diff --staged --quiet` is never true and the workflows commit `Update models manifest [skip ci]` after **every** release build:

| commit | aibom version |
| --- | --- |
| a0796ef98 | 2.1.761-dev |
| 362633792 | 2.1.758-dev |
| 4e206b11e | 2.1.757-dev |
| edc47c3af | 2.1.755-dev |
| 1d4a0fa8f | 2.1.754-dev |
| 8e434e942 | **unknown** ← #1909 itself |

GitHub honours CI-skip directives on a PR's **head commit**, so any release PR cut while that commit is dev's tip opens with zero workflow runs, and `main-branch-ruleset`'s five required checks can never report — permanently `BLOCKED`, not merely pending. That's #2078 and #2081. Before #1909 the commit only landed on real HuggingFace model updates, which is why every earlier release PR was fine.

## The fix

The committed copy is documented as a *"diffable model inventory"*. It describes the bundled models, which don't change when the engine version does — so the stamp doesn't belong there. It stays in the two copies that genuinely are release-specific:

| Copy | Generated by | Version stamp |
| --- | --- | --- |
| S3 `aibom/<version>/` | `push-aibom` (regenerates independently) | **kept** |
| Baked into the image `/models` | Dockerfile `downloader` | **kept** |
| Committed `aibom.cdx.json` | `build-*-model-upload` | **dropped** |

## Second commit: drop the unused `aibom-<version>` CI artifact

Nothing downloads it — there is no `download-artifact` for `aibom` anywhere, and every machine consumer regenerates its own copy. The bytes are identical to the committed file, which lives in git permanently and is easier to reach.

It was also uploaded 1–3 times per run, non-deterministically. The three model-upload workflows are `workflow_call` reusable workflows invoked from one parent run, so whenever more than one builds rather than retags, each uploads the same file under the same name:

```
run 30573975963:
  name=aibom-2.1.754-dev  id=8772249119  size=2330
  name=aibom-2.1.754-dev  id=8772230878  size=2330
  name=aibom-2.1.754-dev  id=8772205216  size=2330
```

That makes `gh run download aibom-2.1.754-dev` ambiguous, so it was unhelpful for the one thing it existed to do. 21 lines removed, nothing lost.

The **committed** copy is deliberately kept: it is the only version-controlled record of model hashes, licenses and tasks (`models-manifest.json` holds commit SHAs only), and it no longer churns.

### Dropping the CLI flag would not have been enough

`--engine-version` defaults to `os.environ.get("VERSION", "unknown")`, and CI sets `VERSION` in `$GITHUB_ENV`. Simply removing the flag would have left the stamp leaking in through the environment and fixed nothing. Hence an explicit `--no-engine-version`, mirroring the existing `--no-timestamp`, which exists for exactly the same diff-stability reason.

## Verification

`aibom.cdx.json` is **regenerated in this commit**, so the first post-merge run matches it and produces no commit at all. Against real HuggingFace metadata:

```
diff <committed> <freshly generated with --no-engine-version>
9,10c9
<       "name": "arthur-genai-engine-models",
<       "version": "2.1.761-dev"
---
>       "name": "arthur-genai-engine-models"
```

Exactly the version line, nothing else — and regenerating a second time is a no-op, which is the property that actually matters.

Three tests cover what could silently regress:

- the field is absent when omitted
- omitting it perturbs **nothing else** in the document (`stamped.pop("version") == bare`)
- `--no-engine-version` overrides the `$VERSION` environment default — the trap above, encoded so nobody reintroduces it

9 passed. black + isort clean.

## Why this instead of #2082

#2082 guarded the release-PR boundary by landing an empty commit when dev's tip carried a skip directive. It needed a `chore(release)` prefix coupled to an allow-list in `version-workflow.yml` — without which the empty commit triggered a version bump, a rebuild, a fresh `[skip ci]` manifest commit, and the PR was blocked again. Two files that had to agree forever, guarding a symptom. This removes the symptom instead.

## Note

Merging this also unblocks **#2081**: it advances dev's tip to a merge commit with no skip directive, firing `pull_request: synchronize` so the required checks finally run.


ntatsumi:
Good catch, and correct — fixed in 618667275.

Removing the `Upload AIBOM artifact` step in the second commit made both halves of that bullet false: the title `Repo + CI artifact` and the clause "and also uploaded as a workflow artifact". I added five lines directly beneath that sentence and didn't re-read the two above them. Took the suggested wording essentially as-is:

```
- **Repo** — the committed `aibom.cdx.json` (diffable model inventory) is regenerated by the
  model-upload build workflows and committed back to the branch.
```

Also grepped for any other stale artifact wording about the AI-BOM across the repo's markdown — nothing else refers to it.

On the verification notes: the `metadata.tools[].version` check is the useful one, since it's the only other version-ish field in the document and a constant `GENERATOR_VERSION` rather than anything per-release. That agrees with the empirical result the PR relies on — regenerating twice against live HuggingFace metadata is a byte-level no-op, which is what actually guarantees the manifest commit stops being produced.

PR #2099, opened 2026-07-31
ci(renovate): route major bumps to domain experts for review

ntatsumi:
## What

Routes **major** dependency-bump PRs to the right domain experts as requested reviewers, deterministically.

| Category | Manager(s) | Major-bump reviewers |
|---|---|---|
| Frontend | `npm` | `jan-lewandoski-blazity`, `xV3rt` |
| Backend | `pep621` | `ntatsumi`, `notthattal` |
| Infra | `dockerfile`, `docker-compose`, `custom.regex` (the `uv` pin) | `ntatsumi`, `jdbarthur` |

## Why

Major bumps are intentionally held for human review (labelled `major`, **not** auto-approved — see `renovate-autofix.yml`). But reviewer assignment didn't respect domain: the manager-wide npm/pep621 rules add **one randomly-sampled** reviewer (`reviewersSampleSize: 1`) to *every* update including majors, so a frontend major could land on a backend reviewer and vice-versa.

## How

- **Scoped** the two existing reviewer rules to `matchUpdateTypes: ["minor", "patch"]` — unchanged behavior for auto-merged updates (still one random reviewer for visibility).
- **Added** three `matchUpdateTypes: ["major"]` rules that request the right experts. No `reviewersSampleSize`, so **both** experts are always requested (not sampled). Scoping the existing rules to non-major means majors are matched *only* by the new rules, so there's no dependency on packageRule array-merge/sample-size precedence.
- The `major` label + autofix "flag for human review, don't self-approve" flow is **unchanged**; these rules only add reviewers.

## Notes

- Infra constraints still gate *which* majors appear: Python base-image bumps are disabled and postgres is capped `<17`, so those never produce a major PR. Real infra majors routed by the new rule: **Node base image** and **`uv`** majors.
- All handles verified as repo collaborators with write+ access (`jan-lewandoski-blazity`: write, `ntatsumi`: admin, `xV3rt`: write, `notthattal`: write, `jdbarthur`: admin). The frontend handle is `jan-lewandoski-blazity` (no "w" — the "w" spelling doesn't exist).
- Validated with `renovate-config-validator` (the new rules pass cleanly; the pre-existing `managerFilePatterns` warnings are a validator-version artifact and out of scope).

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -131,10 +131,10 @@
       "reviewers": ["ntatsumi", "notthattal"]
     },
     {
-      "description": "Route major infra bumps (Docker base images, docker-compose services, and the uv pin tracked by the custom regex manager) to the backend experts for human review.",
+      "description": "Route major infra bumps (Docker base images, docker-compose services, and the uv pin tracked by the custom regex manager) to the infra experts for human review.",
       "matchManagers": ["dockerfile", "docker-compose", "custom.regex"],
       "matchUpdateTypes": ["major"],
-      "reviewers": ["ntatsumi", "notthattal"]
+      "reviewers": ["ntatsumi", "jdbarthur"]
     }
   ]
 }
--- a/renovate.json
+++ b/renovate.json
@@ -105,14 +105,36 @@
       "allowedVersions": "<17"
     },
     {
+      "description": "Non-major npm updates get one randomly-sampled reviewer for visibility (they are auto-merged). Major npm bumps are routed to the frontend experts by the rule below.",
       "matchManagers": ["npm"],
+      "matchUpdateTypes": ["minor", "patch"],
       "reviewers": ["TheLukaszNs-Blazity", "xV3rt", "martin-arthur-ai"],
       "reviewersSampleSize": 1
     },
     {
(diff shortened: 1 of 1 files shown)


ntatsumi:
Addressed in `b71a5c8`, but with a different fix than suggested.

The finding is correct: scoping the reviewer rules to `["minor", "patch"]` did drop the auto-assigned reviewer for `lockFileMaintenance` and `replacement`. Rather than assign reviewers to those, they're now **auto-merged** (added to the `automerge`/`autoApprove` rule) — both are low-risk (nightly lockfile refresh; curated drop-in swaps from `replacements:all`) and gated by CI + the auto-fixer, so they shouldn't sit waiting on a human at all.

On the broader list — `pin`, `pinDigest`, `digest`, `bump`, `rollback` don't occur under this config: `config:recommended` doesn't enable pinDigests/rollbackPrs/bumpVersion, and dependencies are exact-pinned (`==`) with a committed `yarn.lock`, so there's nothing to pin/bump. `lockFileMaintenance` and `replacement` were the only real cases, and both are now handled.

One accepted trade-off: `replacement` swaps auto-merge on green CI without a human glance — acceptable since they're curated drop-ins and CI (type-check/lint/build/tests) gates them, same as any other auto-merged update.

PR #2109, opened 2026-08-06
chore(ui): enforce the 3-day dependency cooldown at the Yarn install layer

ntatsumi:
Closes the gap that `renovate.json` alone leaves open: `"minimumReleaseAge": "3 days"` has been set for a while, but it only gates **Renovate PRs**. A plain `yarn add` in `genai-engine/ui` — or any transitive resolution during a lockfile-changing install — was governed only by Yarn's default gate.

Ported from [unify-frontend!1450](https://gitlab.com/ArthurAI/unify-frontend/-/merge_requests/1450), but much smaller here — see below.

## Changes

1. **`npmMinimalAgeGate: 3d`** in `.yarnrc.yml` — refuses any version published less than 3 days ago, matching the Renovate window. Compromised releases are typically yanked within hours, so the cooldown filters them at the install layer.

2. **`npmPreapprovedPackages: ["@arthur/*"]`** — exempts first-party packages.

3. **`@arthur/**` added to `renovate.json`'s first-party soak exemption**, so both layers agree. Mirrors `tool.uv exclude-newer-package` for `arthur-common` / `arthur-client`.

4. **`meticulous.yaml`: `corepack prepare --activate`** (was `yarn@4.9.2`).

5. **Documented in `genai-engine/ui/CLAUDE.md`**, not `.yarnrc.yml` — Yarn rewrites that file and strips comments, so a comment there would not survive the next `yarn install`.

## No Yarn bump needed (unlike !1450)

That MR had to bump 4.9.2 → 4.18.0 because 4.9.2 predates the feature. We are already on **4.17.1**, which supports the gate (landed in 4.10) *and* already parses the day suffix correctly — `yarn config get npmMinimalAgeGate` resolves `3d` → **4320** minutes, so [yarnpkg/berry#6991](https://github.com/yarnpkg/berry/issues/6991) is already fixed at 4.17.1.

So none of !1450's side effects apply: no `enableScripts` / `approvedGitRepositories` migration churn, no `__metadata.version` bump (already 10), no `resolve` builtin-patch hash reshuffle. **`yarn.lock` is untouched.**

> [!IMPORTANT]
> The gate is **on by default at `1d`** in Yarn ≥4.10. This repo has therefore had a silent **1-day** install-layer cooldown since moving to 4.17.1. This widens it to 3 days; it does not add a gate from nothing.

## Why `@arthur/*` needs an exemption here but not in unify-frontend

There, `@arthur/*` is `workspace:*` and the gate never sees it. Here `@arthur/shared-components@npm:^3.23.3` resolves from the GitLab registry as a real packument, so it *is* gated. Two independent reasons to exempt it:

- First-party code carries no third-party supply-chain risk — the same rationale as the existing `renovate.json` rule.
- **Yarn's gate is fail-closed on missing publish timestamps.** From the bundled release:
  ```js
  let n = s?.[r];
  if (typeof n > "u" || (new Date().getTime() - new Date(n).getTime())/60/1e3 < a) return true  // quarantined
  ```
  A version absent from the packument's `time` map is treated as quarantined. If GitLab's npm registry omits per-version times (I could not confirm either way without a registry token), an unexempted bump would be blocked **indefinitely**, not merely for 3 days.

## Verified against the live registry, not just read from docs

Using `@aws-sdk/client-s3` (publishes daily) in a throwaway project on 4.17.1:

| Scenario | Result |
| --- | --- |
| Pin `3.1104.0`, published **0.96 days** ago | ❌ `YN0016: All versions satisfying "3.1104.0" are quarantined` |
| Unpinned `yarn add` | ✅ degrades to `3.1101.0` (5.96d), skipping the 3 fresher releases |
| Same 0.96d pin, but scope preapproved | ✅ installs — escape hatch works |
| `yarn install --immutable`, that 0.96d version **locked** | ✅ passes — gate not consulted |

**That last row is the compatibility argument.** The gate runs only in `getCandidates` (resolution), so `arthur-engine-workflow.yml`'s `yarn install --immutable` and every existing lockfile entry are unaffected. Only `yarn add`, `yarn up`, and Renovate bumps change behavior.

## On the `meticulous.yaml` change

The step pinned `corepack prepare yarn@4.9.2 --activate`, stale against `package.json`'s 4.17.1. It was **inert** — `packageManager` wins for in-project invocations, which is why CI passed despite the mismatch — and it runs *after* `setup-node`, so it never fed the `cache: yarn` detection either. Dropping the version argument makes `package.json` the single source of truth so it cannot drift again.

Worth noting it was latent rather than harmless: had it taken effect, 4.9.2 cannot read a version-10 lockfile under `--immutable`, and it predates `npmMinimalAgeGate` entirely — which would have made the Meticulous job the one CI path with no cooldown at all.

## Test plan

- [x] `yarn config get npmMinimalAgeGate` → `4320`; `npmPreapprovedPackages` → `["@arthur/*"]`
- [x] Four live-registry scenarios above
- [x] `renovate.json` valid JSON; `.yarnrc.yml` passes `prettier --check`
- [x] `meticulous.yaml` parses; `grep -rn "yarn@[0-9]" .github/` → no matches
- [x] Pre-commit hooks pass
- [ ] CI green — the real check that `--immutable` is unaffected


Encoura-Lewandoski (approved):

.github/workflows/meticulous.yaml
```diff
@@ -44,7 +44,7 @@ jobs:
 
       - name: Prepare Yarn
         working-directory: genai-engine/ui
-        run: corepack prepare yarn@4.9.2 --activate
+        run: corepack prepare --activate
```
jan-lewandoski-blazity, line 47: We removed `yarn@4.9.2` from here but it's still mentioned in `@genai-engine/ui/README.md` as recommended version. Just flagging this in case we want to remove it there as well 👍

jan-lewandoski-blazity (dismissed):

.github/workflows/meticulous.yaml
ntatsumi: Good catch — fixed in 4afff02. The README now points at the `packageManager` field in `package.json` instead of naming a version, and the `corepack prepare` command there is argless too, so both spots read the pin rather than duplicating it.

PR #2138, opened 2026-08-17
ci: make renovate auto-fix goal-oriented and stop it losing work to backgrounding

ntatsumi:
## What

Ports two defects found and fixed in arthur-scope's GitLab parity script ([!1841](https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1841), [!1844](https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1844)), where this same prompt design cost a complete, correct fix. Plus a job timeout bound.

The auto-fixer here **does work** — `7c030378a` and `ae336c500` are real fixes it landed. This is about two failure modes it is still exposed to.

## 1. Session lifecycle — it can throw away a finished fix

The prompt never said the run is a single non-interactive turn.

In arthur-scope, Claude produced a correct fix, verified the constraint change, then started the test suite **in the background** and ended its turn waiting for the result:

> `I'll wait for the background test run to notify completion.`

Nothing wakes a non-interactive run. The session ended at 17 minutes of a 90-minute budget, no marker was written, no comment was posted, and the workspace was torn down with the finished fix inside it. `claude-code-action` is equally non-interactive, so the exposure here is identical — and the Python unit suite and UI checks are exactly the kind of slow command that invites backgrounding.

Added a `HOW THIS SESSION WORKS` preamble, stated before the goal: nothing will wake you, the moment you stop producing output the run ends and the tree is discarded, so **run every check in the foreground and block on it**. The only two valid endings are marker written, or edits discarded plus a cannot-fix comment.

## 2. Scope — it cannot fix the most common way Renovate breaks a build

The prompt listed dependency-resolution conflicts as out of scope and forbade touching any pin or lockfile. So when Renovate bumps half of a lockstep family, the fixer must stand down.

**This repo has been absorbing that cost by hand.** The `transformers` cap (gliner's ceiling), the `importlib-metadata` cap (the OpenTelemetry stack's ceiling), and the `fsspec` and `langchain + openai` groups in `renovate.json` are all humans doing what the fixer was forbidden from doing. The `fsspec` rule even states the general case: *"Splitting them into separate PRs makes each individually unsatisfiable."*

Replaced the file-scoped allowlist with a **goal plus invariants** that must hold of the final diff:

1. **The update still lands** — never downgrade or cap the package being updated to force a solve. That is the tempting shortcut and it no-ops the PR while making it look fixed. Conflicts resolve by moving the **lagging** package forward.
2. **No check is weakened** — and see below.
3. No behaviour change beyond the adaptation.
4. **Lockfiles are generated, never authored** — `uv lock --directory genai-engine`, `yarn install`; never hand-edited.
5. The 3-day release-age soak is not bypassed.
6. `.github/` and `/renovate.json` stay off limits.
7. Any constraint edit is declared in the review comment, with the resolver output that forced it.

Verification is now explicit and has to be real: reproduce the failure first, show the failing command passes, then run the **other** required checks covering what was touched — not just the one that was red — and confirm the invariants against the actual diff.

### Invariant 2 covers a real case from this repo

`7c030378a` adapted `ui/eslint.config.js` for react-hooks v7 — a legitimate fix, since v7 moved its flat configs and the old `configs["recommended-latest"]` reference broke. But it kept only `rules-of-hooks` and `exhaustive-deps` and **quietly dropped the React Compiler rules** that the preset it replaced would have enabled, under a rule that said not to weaken checks.

Invariant 2 now draws that line explicitly: adapting a linter config to a plugin's moved API is allowed; shrinking the set of rules actually enforced without saying so is not — and if a fix must drop rules, it has to be called out in the review comment rather than passing silently.

## 3. `timeout-minutes: 90`

The job inherited GitHub's 6-hour default, so a hung session could hold a runner all day. 90 minutes covers reproduce + fix + two full verification passes.

## Validation

No CI exercises this workflow directly (it is a `workflow_run` listener that only activates from the default branch), so validation is static:

- Parses as YAML; `timeout-minutes` reads back as 90.
- All 37 `${{ }}` expressions balanced and well-formed.
- The set of expressions **inside the prompt** is identical to before — none dropped, none added — so no `steps.pr.outputs.number` / `workflow_run.id` reference was mangled in the rewrite.
- Block-scalar indentation uniform across all 141 prompt lines.
- No remnants of the old `IN SCOPE` / `OUT OF SCOPE` / `HARD RULES` text.

## Note on recent failures

The last four real auto-fix runs (Aug 13–14) all failed identically with `is_error:true`, `num_turns: 1`, `total_cost_usd: 0`, `duration_ms: 396`. That fingerprint is an API-level rejection, not a code problem — the same shape as an `ANTHROPIC_API_KEY` spend limit that was blocking arthur-scope until it was raised. **This PR does not address that**; if the two repos share a key or org it may already be resolved, and re-running one failed PR would confirm it.

## Follow-up not included here

The `autofix: dependency-realignment` label (invariant 7) does not exist in this repo. The prompt tolerates that and falls back to the comment, matching how the existing soak label is handled — but creating the label would make constraint edits greppable. Happy to add it if wanted.

### DIFF
--- a/.github/workflows/renovate-autofix.yml
+++ b/.github/workflows/renovate-autofix.yml
@@ -16,8 +16,13 @@ name: Renovate Auto-Fix
 #      before it can merge. This workflow cannot bypass that.
 #   3. Scope/tool guards: runs only on renovate/* branches, only on CI failure,
 #      a bounded retry budget (default 3 attempts, then it hands off to a human),
-#      and the prompt forbids weakening checks or touching CI/test/dependency-pin
-#      files.
+#      and a prompt bounded by invariants rather than a file allowlist. It forbids
+#      weakening checks (the enforced rule set may never shrink), editing .github/
+#      or /renovate.json, and bypassing the 3-day release-age soak. Dependency
+#      constraints MAY be realigned — that is how a half-bumped lockstep family is
+#      repaired — but never by downgrading or capping the package the PR is
+#      updating, lockfiles are regenerated via uv/yarn rather than hand-edited, and
+#      every constraint change must be declared in the self-review comment.
 #
 # Identity split (required by the ruleset's require_last_push_approval +
 # dismiss_stale_reviews_on_push): the fix is PUSHED as the Arthur GitHub
--- a/.github/workflows/renovate-autofix.yml
+++ b/.github/workflows/renovate-autofix.yml
@@ -63,6 +63,11 @@ jobs:
       ) &&
       startsWith(github.event.workflow_run.head_branch, 'renovate/')
     runs-on: ubuntu-latest
+    # Bound the run. Without this the job inherits GitHub's 6h default, so a Claude
+    # session that hangs burns a runner for hours. The budget has to cover reproducing
+    # the failure, fixing, and re-running the full required suite (all four Python
+    # linters + unit tests, or the three UI checks) twice.
+    timeout-minutes: 90
     # Needed to read the Arthur GitHub Automator App credentials.
     environment: shared-protected-branch-secrets
     permissions:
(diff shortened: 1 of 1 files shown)


ntatsumi:
Good catch — fixed in efda127f8.

You were right that this was the only description of the safety model in the file, and that the `dependency-pin` clause had become flatly false: invariant 4 makes editing `pyproject.toml` / `package.json` constraints and regenerating the lockfiles the *prescribed* repair, and invariant 7 presupposes it.

Rewrote layer 3 around what the invariants actually enforce rather than a file list:

```
#   3. Scope/tool guards: runs only on renovate/* branches, only on CI failure,
#      a bounded retry budget (default 3 attempts, then it hands off to a human),
#      and a prompt bounded by invariants rather than a file allowlist. It forbids
#      weakening checks (the enforced rule set may never shrink), editing .github/
#      or /renovate.json, and bypassing the 3-day release-age soak. Dependency
#      constraints MAY be realigned — that is how a half-bumped lockstep family is
#      repaired — but never by downgrading or capping the package the PR is
#      updating, lockfiles are regenerated via uv/yarn rather than hand-edited, and
#      every constraint change must be declared in the self-review comment.
```

Two notes on the edit beyond your suggestion:

- I also dropped **`test`** from the clause. Test adaptation was never actually forbidden — the pre-PR `IN SCOPE` block explicitly allowed "unit-test adaptations that are a legitimate consequence of the dependency's changed behavior" — so that word was misleading before this PR too.
- I kept the "never downgrading the package being updated" qualifier prominent, since that is the one shortcut that would make a PR go green while silently no-opping the update it exists to deliver.

Verified after the edit: parses as YAML, `timeout-minutes` still reads back as 90, all 37 `${{ }}` expressions balanced, and `dependency-pin` no longer appears anywhere in the repo.

PR #2140, opened 2026-08-18
feat(ci): emit Arthur Engine traces from the Renovate auto-fixer

kc961163:
## Why

The Renovate auto-fixer runs Claude Code to repair dependency-update failures, but its prompts, tool calls, failures, and completion are not currently observable in Arthur.

## What

- Install the OpenTelemetry dependencies without blocking the auto-fix if installation fails.
- Copy the existing Claude Code tracer to `$RUNNER_TEMP`, outside Claude's editable checkout.
- Register prompt, tool success, tool failure, and stop hooks using Claude Code's event-specific default timeouts.
- Keep missing files and tracer errors non-blocking so tracing cannot interrupt the auto-fix.
- Route traces to a dedicated Arthur development task using configuration from the protected GitHub environment.
- Attribute each trace to `renovate-autofix-pr-<PR number>`.

Tracing is auxiliary: a missing tracer, dependency, or configuration value cannot prevent Claude from attempting or publishing an autofix.

## Validation

- `235 passed` in `integrations/claude-code-observability/test_tracer.py`.
- Workflow YAML and the embedded Claude settings JSON parse successfully.
- All five hooks reference the runner-temporary tracer and omit explicit timeouts, matching the existing Arthur Claude tracing workflow examples.
- Simulated missing and crashing tracer paths both return success and cannot block the auto-fix.
- `git diff --check` passes.

The tracing destination and credentials are configured outside the repository through `shared-protected-branch-secrets`; no endpoint, task identifier, or credential is present in this PR.


notthattal (dismissed):
lgtm

ntatsumi:
**Make traces attributable to a PR**

`arthur.user` comes from `USER` (`claude_code_tracer.py:1293`, `:1382`), which is always `runner` on a GitHub-hosted runner, and `arthur.session` is the opaque Claude session UUID. Neither `steps.pr.outputs.number` nor `github.event.workflow_run.id` reaches the tracer, so in the Arthur task every auto-fix trace is the same user with an unmappable ID — you can't get from a Renovate PR to its trace, or from a trace back to its PR. That's the correlation this PR is for.

The tracer's entire env surface is five variables — the three config vars (`:65-67`), `CLAUDE_PROJECT_DIR`, and `USER`/`USERNAME` — so `USER` is the only identity channel injectable from the workflow, and one line in the `env:` block this PR already adds is enough:

```yaml
        env:
          AUTOFIX_MARKER: ${{ runner.temp }}/claude-autofix-success
          USER: renovate-autofix-pr-${{ steps.pr.outputs.number }}
          GENAI_ENGINE_API_KEY: ${{ secrets.GENAI_ENGINE_API_KEY }}
          GENAI_ENGINE_TASK_ID: ${{ vars.GENAI_ENGINE_TASK_ID }}
          GENAI_ENGINE_TRACE_ENDPOINT: ${{ vars.GENAI_ENGINE_TRACE_ENDPOINT }}
```

Step env propagates to the action, to the CLI, to the hook subprocess; the tracer captures it once at session init and emits it as the `arthur.user` resource attribute (`:979`). Blast radius is contained to this step, and nothing else in the session reads `USER`.

Overloading `USER` is a workaround for a tracer gap rather than the right shape — there's no first-class way to attach CI identity. A dedicated `ARTHUR_TRACE_USER`, or PR number and run id as span attributes, is a change to `integrations/claude-code-observability/` and worth its own PR, since the other three workflows using `claude-code-action` would want the same thing. Suggest shipping the `USER:` line here and filing the tracer field separately.

ntatsumi:
**Move the tracer out of the agent's workspace**

The tracer is copied to `.claude/hooks/claude_code_tracer.py`, which `.gitignore:56` ignores, inside the same tree Claude has `Write` and `Bash` on. Two problems fall out of that, one latent and one unconditional.

**The latent one: a deleted tracer blocks every tool call.** I verified the mechanics in this repo — `git clean -fdxn -- .claude` reports `Would remove .claude/hooks/`, while plain `git clean -fd` leaves it alone; `python3 <missing-file>` exits 2, including with stdin piped; and exit 2 from `PreToolUse` is the documented blocking code. So if the tracer disappears mid-session, every subsequent tool call is refused, and Claude can no longer discard its edits, write `$AUTOFIX_MARKER`, or post the cannot-fix comment — the job runs to the 90-minute timeout having produced no fix and no signal.

To be clear about likelihood: the prompt instructs `git checkout -- . && git clean -fd`, which is safe. Firing this needs an escalation to `-fdx`, which isn't instructed but is exactly what you reach for against a dirty `node_modules` or `.venv`. So: not certain, but plausible, and the failure is both silent and unrecoverable when it happens. The guard below also neutralizes any *other* cause of a nonzero tracer exit, which is worth having on its own.

**The unconditional one: the audit trail sits inside the sandbox being audited.** `git status --porcelain -- .claude` returns nothing for this path, so any edit to the tracer is invisible to the commit gate at line 362 and to both the human and Claude reviewers of the pushed fix. Invariant 6 fences off `.github/` and `/renovate.json` but says nothing about `.claude/`. This is the same reasoning the file already applies to the success marker — *"the marker lives OUTSIDE the repo so it cannot be committed to forge a 'verified' signal"* — and it applies to the tracer for the same reason.

Both are fixed by the mechanism this file already uses for the marker. In the setup step:

```yaml
          cp integrations/claude-code-observability/claude_code_tracer.py "$RUNNER_TEMP/claude_code_tracer.py"
```

and per hook (shown for `PreToolUse`; same shape for the other four):

```json
"PreToolUse": [{"matcher": "", "hooks": [{"type": "command", "timeout": 5, "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" pre_tool; true"}]}]
```

`$RUNNER_TEMP` puts the tracer where the agent's own git commands can't reach it; `[ -f ... ]` plus the trailing `; true` guarantees exit 0 whatever happens, so tracing can never block a dependency fix. As a bonus this also removes the need for `cp` to succeed, so the `set -euo pipefail` hard-failure on a missing source file — reachable because the workflow body always comes from the default branch while the tracer comes from the PR branch — stops being a way for tracing setup to kill the job it's meant to observe.

kc961163:
Implemented in `824c365`:

- Traces are attributed with `USER=renovate-autofix-pr-<PR number>` through the Claude action's `settings.env`.
- The tracer now runs from `$RUNNER_TEMP`, outside Claude's checkout.
- Setup, copy, missing-file, tracer-error, and timeout paths are fail-open; each hook has a five-second limit and unconditional success.
- The tracer suite passes (`235 passed`), and simulated missing, crashing, and hanging tracer paths cannot block the autofix.

ntatsumi:
Is 5 second limit too aggressive? Maybe we can start with something more lenient. If our tracing instrumentation is not tolerable we should optimize rather than giving up using it.

kc961163:
Updated in `e666483`:

- Removed both explicit five-second limits from all five hooks, matching the existing Arthur Claude tracing workflows and using Claude Code’s event-specific defaults.
- Kept the `$RUNNER_TEMP` file guard and unconditional success, so missing or failing tracing remains non-blocking.
- Revalidated the workflow YAML and embedded settings JSON; the tracer suite passes (`235 passed`), and missing/crashing tracer simulations exit `0`.

ntatsumi (approved):

PR #2162, opened 2026-08-21
consolidate datasets

notthattal:


genai-engine/src/repositories/tasks_repository.py
```diff
@@ -630,13 +630,8 @@ def check_llm_rule_count(self, enabled_rules: list[DatabaseRule]) -> None:
...
     def link_metric_to_task(self, task_id: str, metric_id: str) -> None:
-        # Check if task is agentic before allowing metric linkage
-        db_task = self.get_db_task_by_id(task_id)
-        if not db_task.is_agentic:
-            raise HTTPException(
-                status_code=400,
-                detail=constants.ERROR_NON_AGENTIC_TASK_METRIC,
-            )
+        # Validate the task exists (404) before creating the link.
+        self.get_db_task_by_id(task_id)
```
TomLisankie, line 634: I think we can be more clear with the intent here:
1. if someone doesn't read the comment, it wouldn't be obvious that the whole point of this is to raise an exception if the task doesn't exist.
2. even if the task does exist, we just fetched and hydrated a whole task object for an existence check.

What if we instead did:
```python
def task_exists(self, task_id: str) -> bool:
    exists_query = self.db_session.query(DatabaseTask).filter(DatabaseTask.id == task_id).exists()
    return bool(self.db_session.query(exists_query).scalar())

# ...

# and instead of self.get_db_task_by_id(task_id) we do
if not self.task_exists(task_id):
    raise HTTPException(
                status_code=404,
                detail="Task %s not found." % task_id,
            )
```

TomLisankie (changes requested):

genai-engine/src/repositories/tasks_repository.py
notthattal: I get the idea, but there is already existing precedent for using this helper to 404 in link_metric_to_task and this new helper would add very similar functionality. Plus, we're only grabbing the db item not creating a Task object.

kc961163 (dismissed):

kc961163 (dismissed):

TomLisankie (dismissed):

kc961163 (dismissed):

TomLisankie (approved):

PR #2167, opened 2026-08-24
Update ML Engine dev install instructions in README

TomLisankie:


ntatsumi:
**Recommendation: drop the JDK prerequisite and use the generator CI already uses.**

`ml-engine/scripts/openapi_client_utils.sh` drives the npm `openapi-generator-cli`, which needs a system JVM. CI doesn't — `arthur-engine-workflow.yml:97` (also :480, :586) uses the PyPI build, which bundles one via `jdk4py`.

The script already self-installs its generator, so put it there: replace the body of `install_or_update_openapi_gen` (line 15) with

```bash
pip install 'openapi-generator-cli[jdk4py]==7.12'
```

and drop the `version-manager` check at lines 32–39 — the pin now lives in the install line.

- **README prereqs stay as-is**, still just uv — no `brew install openjdk`, no `sudo` symlink, no `/opt/homebrew` path that's wrong on Intel Macs. (Documenting it in the README instead also works; it's one `pip install` line in place of three, but then it's on each developer to run it.)
- **Full coverage.** Java is needed by *every* run of the script today, not just `generate`: the npm wrapper shells out to `java -jar … help` at bootstrap, so the line 36 version check fails even for `install python`. It's also needed by the Docker section's step 1, which this PR doesn't cover.
- **CI alignment.** Same generator, same 7.12 pin. And `generate -c ml-engine/scripts/openapitools.json` would let the script use the config file whose settings its inline flags currently duplicate.

TomLisankie:
Thanks — went with this. Pushed 1cdca30a: the README JDK prerequisite is reverted, and `openapi_client_utils.sh` now installs the PyPI generator instead of the npm one, so no system JVM is needed. Two departures from your sketch:

**`python3 -m pip` plus a PATH prepend, not a bare `pip install`.** Earlier versions of the script did `npm install -g @openapitools/openapi-generator-cli`, so anyone who has run it already has that shim on PATH — on my machine it shadowed the pip install, and the script kept invoking the npm wrapper and kept needing system Java. Prepending the pip scripts dir makes the pinned install win, and deriving both pip and that dir from the same `python3` keeps them from disagreeing.

**Kept the inline `generate` flags rather than `-c ml-engine/scripts/openapitools.json`.** That config's paths are relative to the repo root while the script runs from `ml-engine/scripts`, and it doesn't set `packageVersion`, which the script pulls from the spec's `info.version` — switching as-is would silently drop the client's version stamp. Happy to do it as a follow-up if we have the script `cd` to the repo root and add `packageVersion` to the config.

Verified with a failing `java` stub shadowing system Java and `JAVA_HOME` unset: `generate python` writes `.openapi-generator/VERSION` = 7.12.0 and a client stamped 2.1.811 (same as before), and `install python` succeeds — which confirms your point that the old version check broke that path even though it needs no generator at all. `openapi_generator_cli` prefers `jdk4py.JAVA` over `shutil.which("java")`, so system Java is never consulted. ml-engine unit tests: 267 passed, 1 skipped.

PR #2215, opened 2026-09-01
fix(ci): finalize Renovate Claude traces from Agent SDK output

kc961163:
## Problem

A Renovate autofix can finish successfully while its Arthur trace contains tool activity but no model-response spans or final answer. In [run 35328776743](https://github.com/arthur-ai/arthur-engine/actions/runs/35328776743), the Action reported a completed SDK execution and an execution-log file; the [matching development trace](https://engine.development.arthur.ai/tasks/29e239eb-85ce-418c-a509-b1c0c490234c/traces?id=0c9eb6b006f51ee43be10e0ca904418f) lacked the expected model output. The runner's original hook transcript was not retained, so its exact contents remain unverified.

The reproduced failure is a mismatch between finalization and its input: a hook transcript can contain only the prompt while the Action's completed execution log contains the assistant response. Given “Fix the failed Renovate PR” and an SDK assistant response “Autofix complete”, the old tracer misses that response; this change emits an LLM span and sets the parent CHAIN output to “Autofix complete”. CHAIN represents the overall turn; TOOL and LLM spans describe tool operations and model responses inside it.

## Solution

1. Finalize after the Claude Action, using its `session_id` and `execution_file` outputs. Keep the current `claude` step ID so the newer attempt-outcome ledger references the same Action. Finalization is bounded to 120 seconds and tracing errors do not fail the auto-fix. Preserve the heavy-label guard, attempt guard, apply step, and outcome recording from current `dev`.
2. Remove only the Renovate workflow's early Stop hook. Finalization deletes the tracer's saved state; keeping both finalizers could consume it before the completed execution log is available. Keep prompt and tool hooks. Local interactive Claude still uses Stop.
3. Accept SDK JSON arrays alongside local JSONL transcripts, prefer a usable execution file, recover a missing session ID from `system.init`, retain the captured prompt when the SDK file omits a user event, and ensure the parent encloses its emitted model spans.
4. Install the tracer's existing OpenTelemetry requirements in the test workflow so the regression can exercise span construction.

The relevant arthur-engine workflow change is [`.github/workflows/renovate-autofix.yml:384-400`](https://github.com/arthur-ai/arthur-engine/blob/7617404a8694f1b2ef934e82175f88e4790eadaf/.github/workflows/renovate-autofix.yml#L384-L400):

```yaml
CLAUDE_SESSION_ID: ${{ steps.claude.outputs.session_id }}
CLAUDE_EXECUTION_FILE: ${{ steps.claude.outputs.execution_file }}
```

These values connect the post-Action finalizer to the same Claude execution used by the attempt ledger. This is code-level evidence of the intended wiring, not proof of a patched GitHub execution.

In arthur-engine, [`integrations/claude-code-observability/claude_code_tracer.py:1777-1781`](https://github.com/arthur-ai/arthur-engine/blob/7617404a8694f1b2ef934e82175f88e4790eadaf/integrations/claude-code-observability/claude_code_tracer.py#L1777-L1781) finalizes and consumes the working state:

```python
_emit_pending_llm_spans(state, transcript_path, config)
_complete_turn(state, config, transcript_path, end_ns)
_delete_state(session_id)
```

The first call constructs remaining model spans, the second finishes the parent, and the third removes state. This explains moving finalization instead of invoking it twice.

Anthropic documents [`execution_file` and `session_id`](https://github.com/anthropics/claude-code-action/blob/main/base-action/README.md#outputs). The [Action version used by the observed run](https://github.com/anthropics/claude-code-action/blob/a4f54ef2c58884867281bd8e2f8d63352ad019a9/base-action/src/run-claude-sdk.ts#L186-L233) collects SDK messages and writes the execution file after the query loop. Post-Action finalization is this PR's integration choice based on that contract, not a claim that Anthropic prescribes this exact arrangement.

## Validation

- **Focused test, current merged branch:** `236 passed` for `integrations/claude-code-observability/test_tracer.py`. The SDK-file regression fails against the old tracer (zero model spans instead of one) and passes with the fix. It covers stale cached input, missing explicit session ID, no initial SDK user event, retained prompt, assistant output, and parent/child time ordering. Its exporter is mocked; ordering is not measured model latency.
- **Focused workflow checks:** both changed YAML files and embedded settings JSON parse; finalization follows the Claude step and references its outputs; guard/apply/outcome steps match `dev`. Five shell cases pass: session plus file, file only, missing outputs, missing tracer, and tracer process failure. These use a stub tracer and timeout shim, not a GitHub runner.
- **Local runtime / stored API result, September 19:** a separate synthetic comparison through the existing local Claude task stored one CHAIN with no output for the old tracer, versus one CHAIN plus one LLM with `Synthetic finalization proof output` for the proposed tracer. This was not an interactive Claude or Renovate invocation and was not rerun during the base-branch merge.
- **Static check:** `git diff --check` passes. Four files differ from `dev`; no GenAI API, database, UI, migration, or Codex changes.

## Remaining E2E acceptance

The observed GitHub run demonstrates the original symptom, not the patched workflow. After approval and merge, inspect the next eligible Renovate execution that actually runs Claude. Match its session to the existing development task and verify TOOL and LLM children, retained prompt, final assistant output, and parent timing. A green workflow or skipped Claude step is not tracing proof. Hard runner termination can prevent finalization.

Follow-up to #2140. Codex remains a separate control; this PR does not change its hook contract or reinstall local hooks.

### DIFF
--- a/.github/workflows/renovate-autofix.yml
+++ b/.github/workflows/renovate-autofix.yml
@@ -162,6 +162,7 @@ jobs:
             echo "::warning::Arthur tracer could not be copied; continuing without tracing."
 
       - name: Claude auto-fix (edits + self-verify only; no commit)
+        id: claude_autofix
         if: steps.guard.outputs.proceed == 'true'
         # The success marker lives OUTSIDE the repo (in RUNNER_TEMP) so it cannot be
         # committed to the PR branch to forge a "verified" signal.
@@ -181,8 +182,7 @@ jobs:
                 "UserPromptSubmit": [{"matcher": "", "hooks": [{"type": "command", "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" user_prompt_submit; true"}]}],
                 "PreToolUse": [{"matcher": "", "hooks": [{"type": "command", "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" pre_tool; true"}]}],
                 "PostToolUse": [{"matcher": "", "hooks": [{"type": "command", "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" post_tool; true"}]}],
-                "PostToolUseFailure": [{"matcher": "", "hooks": [{"type": "command", "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" post_tool_failure; true"}]}],
-                "Stop": [{"matcher": "", "hooks": [{"type": "command", "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" stop; true"}]}]
+                "PostToolUseFailure": [{"matcher": "", "hooks": [{"type": "command", "command": "T=\"${{ runner.temp }}/claude_code_tracer.py\"; [ -f \"$T\" ] && python3 \"$T\" post_tool_failure; true"}]}]
               }
             }
           # Allow the action to run even though the PR was authored by a bot.
@@ -361,6 +361,25 @@ jobs:
             invariants hold, verified, self-reviewed, and review findings addressed". The change still goes
             through the full CI + merge queue (and an independent Claude review) before merging. Never leave
             broken, unreviewed, or unverified edits.
+
+      - name: Finalize Arthur Engine trace
+        if: always() && steps.guard.outputs.proceed == 'true'
+        env:
+          USER: renovate-autofix-pr-${{ steps.pr.outputs.number }}
+          GENAI_ENGINE_API_KEY: ${{ secrets.GENAI_ENGINE_API_KEY }}
+          GENAI_ENGINE_TASK_ID: ${{ vars.GENAI_ENGINE_TASK_ID }}
+          GENAI_ENGINE_TRACE_ENDPOINT: ${{ vars.GENAI_ENGINE_TRACE_ENDPOINT }}
+          CLAUDE_SESSION_ID: ${{ steps.claude_autofix.outputs.session_id }}
+          CLAUDE_EXECUTION_FILE: ${{ steps.claude_autofix.outputs.execution_file }}
+        run: |
+          T="$RUNNER_TEMP/claude_code_tracer.py"
+          if { [ -n "$CLAUDE_SESSION_ID" ] || [ -f "$CLAUDE_EXECUTION_FILE" ]; } && [ -f "$T" ]; then
+            jq -cn \
+              --arg session_id "$CLAUDE_SESSION_ID" \
+              --arg execution_file "$CLAUDE_EXECUTION_FILE" \
+              '{session_id: $session_id, execution_file: $execution_file}' | \
+              timeout 120s python3 "$T" stop || true
+          fi
 
       - name: Commit, push, then approve (minor/patch) or flag (major)
         if: steps.guard.outputs.proceed == 'true'
--- a/integrations/claude-code-observability/claude_code_tracer.py
+++ b/integrations/claude-code-observability/claude_code_tracer.py
@@ -271,6 +271,46 @@ def _session_lock(session_id: str):
 # ---------------------------------------------------------------------------
 
 
+def _read_transcript_entries(path: str) -> list[dict]:
+    """Read Claude Code JSONL or Claude Agent SDK JSON-array output."""
+    try:
+        text = Path(path).read_text()
+    except OSError:
+        return []
+
+    try:
+        parsed = json.loads(text)
+    except json.JSONDecodeError:
+        parsed = None
+
+    if isinstance(parsed, list):
+        return [entry for entry in parsed if isinstance(entry, dict)]
+    if isinstance(parsed, dict):
+        return [parsed]
+
+    entries = []
+    for line in text.splitlines():
+        if not line.strip():
+            continue
+        try:
+            entry = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if isinstance(entry, dict):
+            entries.append(entry)
+    return entries
+
+
+def _execution_session_id(path: str) -> str:
+    """Return the Agent SDK session ID from its system.init event."""
+    for entry in _read_transcript_entries(path):
+        if entry.get("type") == "system" and entry.get("subtype") == "init":
+            session_id = entry.get("session_id", "")
+            if isinstance(session_id, str):
+                return session_id
+    return ""
+
+
 def _is_real_transcript(path: str) -> bool:
     """Return True if the transcript has at least one human or assistant entry.
 
@@ -279,21 +319,10 @@ def _is_real_transcript(path: str) -> bool:
     one of these as the authoritative transcript would leave the tracer with
     nothing to extract LLM spans from.
     """
-    try:
-        with open(path) as fh:
-            for line in fh:
-                line = line.strip()
-                if not line:
-                    continue
-                try:
-                    entry = json.loads(line)
-                except json.JSONDecodeError:
-                    continue
-                if entry.get("type") in ("user", "assistant"):
-                    return True
-        return False
-    except OSError:
-        return False
+    return any(
+        entry.get("type") in ("user", "assistant")
+        for entry in _read_transcript_entries(path)
+    )
 
 
 def _find_transcript_path(data: dict, session_id: str) -> Optional[str]:
@@ -377,19 +406,11 @@ def _is_human_message(entry: dict) -> bool:
 
 def _count_human_messages(transcript_path: str) -> int:
     """Count human (non-tool-result) user messages to detect turn boundaries."""
-    try:
-        count = 0
-        for line in Path(transcript_path).read_text().splitlines():
-            if not line.strip():
-                continue
-            try:
-                if _is_human_message(json.loads(line)):
-                    count += 1
-            except json.JSONDecodeError:
-                pass
-        return count
-    except Exception:
-        return 0
+    return sum(
+        1
+        for entry in _read_transcript_entries(transcript_path)
+        if _is_human_message(entry)
+    )
 
 
 def _iso_to_ns(ts: str) -> int:
@@ -423,6 +444,7 @@ def _extract_llm_spans_for_turn(
     human_count_at_start: int,
     trace_id_hex: str,
     root_span_id_hex: str,
+    fallback_human_prompt: str = "",
 ) -> list[dict]:
     """
     Extract LLM spans from transcript entries that belong to this turn.
@@ -465,7 +487,15 @@ def _extract_llm_spans_for_turn(
     """
     spans = []
     try:
-        lines = Path(transcript_path).read_text().splitlines()
+        entries = _read_transcript_entries(transcript_path)
+        if fallback_human_prompt and not any(_is_human_message(e) for e in entries):
+            entries.insert(
+                0,
+                {
+                    "type": "user",
+                    "message": {"role": "user", "content": fallback_human_prompt},
+                },
+            )
         human_count = 0
         in_turn = False
 
@@ -593,14 +623,7 @@ def _extract_llm_spans_for_turn(
             group_stop_reason = ""
             group_input_snapshot = {}
 
-        for line in lines:
-            if not line.strip():
-                continue
-            try:
-                entry = json.loads(line)
-            except json.JSONDecodeError:
-                continue
-
+        for entry in entries:
             entry_type = entry.get("type", "")
 
             # ── Human message: marks the start of this turn ──────────────────
@@ -1074,6 +1097,7 @@ def _emit_pending_llm_spans(
         current_trace.get("human_count_at_start", 0),
         current_trace["trace_id"],
         current_trace["root_span_id"],
+        current_trace.get("prompt_preview", ""),
     )
 
     emitted = current_trace.get("emitted_llm_span_count", 0)
@@ -1086,6 +1110,10 @@ def _emit_pending_llm_spans(
         current_trace["last_llm_output"] = all_llm_spans[-1]["attributes"].get(
             "output.value",
             "",
+        )
+        current_trace["last_llm_end_ns"] = max(
+            current_trace.get("last_llm_end_ns", 0),
+            max(span["end_ns"] for span in all_llm_spans),
         )
 
     if not new_spans:
@@ -1126,6 +1154,7 @@ def _complete_turn(
     trace_id = current_trace["trace_id"]
     root_span_id = current_trace["root_span_id"]
     turn_start_ns = current_trace["turn_start_ns"]
+    end_ns = max(end_ns, current_trace.get("last_llm_end_ns", 0))
     turn_number = current_trace.get("turn_number", 1)
     prompt_preview = current_trace.get("prompt_preview", "")
     final_output = current_trace.get("last_llm_output", "")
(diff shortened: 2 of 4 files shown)


ntatsumi (commented):
## Code review — trace lifecycle changes

Reviewed the full diff and ran the suite on the head commit (**236 passed**). I also re-ran each finding as an executable probe against both this branch and `dev` so the regressions are separated from the new-path gaps; the numbers in the inline comments are from those runs.

The core idea is right — a `Stop` hook genuinely cannot see the Agent SDK execution log, and moving finalization after the action is the correct shape. Three things I'd want addressed before merge:

**Regressions on paths this PR wasn't targeting:**
1. `end_ns = max(end_ns, last_llm_end_ns)` (`claude_code_tracer.py:1157`) is unconditional, so the synthetic 10 ms/token end time now inflates the root CHAIN span on interactive sessions too — measured 30 s reported for a ~0 s turn, overlapping the next turn's root span by 30 s. `dev` reports the true 5 s.
2. `_is_real_transcript` (`:322`) lost its streaming early-return: 0.2 ms / 0 MB → 619 ms / 198 MB on a 37 MB transcript, inside a blocking hook, once per candidate per turn.
3. The dropped `except Exception` wrappers (`:409`, `:1519`) let a `UnicodeDecodeError` escape `handle_stop` before `_complete_turn`, losing the root span and orphaning the state file. `dev` degrades gracefully and still closes the turn.

**The fix itself has gaps:**
4. `fallback_human_prompt` (`:491`) emits nothing at all once `human_count_at_start >= 1` — silently reproducing the bug being fixed.
5. `emitted_llm_span_count` (JSONL-derived) now indexes a span list derived from the execution file (`:1100`), which can drop the final response or duplicate earlier ones.
6. The continuation check at `:1750` compares a count read from the execution file against one computed from the JSONL; with >= 2 user entries in the execution file it emits **two root CHAIN spans for one turn** (verified).

**Smaller:** empty-string `session_id` short-circuit (`:304`), SDK spans stamped at finalize wall clock (`:1736`), finalization now silently skipped if the action writes no outputs (`renovate-autofix.yml:395`), and a dead `USER` env var (`:387`).

Also 👍 on `claude-code-observability-workflow.yml` installing `requirements.txt` — the suite was previously running without the tracer's own deps.

Details and repro numbers inline.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -1126,6 +1154,7 @@ def _complete_turn(
     trace_id = current_trace["trace_id"]
     root_span_id = current_trace["root_span_id"]
     turn_start_ns = current_trace["turn_start_ns"]
+    end_ns = max(end_ns, current_trace.get("last_llm_end_ns", 0))
```
ntatsumi, line 1157: **Regression (confirmed): this stretches the CHAIN span on _every_ path, not just the SDK one.**

`last_llm_end_ns` is not a measured timestamp — `_flush_group` computes `end_ns = start_ns + max(output_tokens * 10_000_000, 1_000_000)`, i.e. a flat 10 ms/token guess. Applying it unconditionally in `_complete_turn` pushes that guess into the root span for interactive sessions too, and `_emit_pending_llm_spans` persists the value to the state file so it survives across hook invocations.

Measured on a 3000-output-token final reply, real turn duration ~0s:

```
after post_tool: emitted_llm_span_count=2, last_llm_end_ns persisted=YES (now +30.0s)
turn 1 CHAIN: duration=30.0s (real ~0.0s), end vs turn-2 start: +30.0s -> OVERLAP
```

Same probe on `dev`: `CHAIN duration = 5.0s (real 5.0s)`.

So turn 1's root span reports 30 s it never took and *overlaps turn 2's root span by 30 s* — two sibling CHAIN spans with inverted wall clocks, and root-span latency becomes fiction for all tracer users, not just Renovate.

The `max()` is only needed when the spans carry no real timestamps (the SDK path). Suggest gating it — e.g. set `last_llm_end_ns` only when `group_ts` was empty, or pass a flag from `handle_stop` — so timestamped transcripts keep using the true `end_ns`.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -465,7 +487,15 @@ def _extract_llm_spans_for_turn(
     """
     spans = []
     try:
-        lines = Path(transcript_path).read_text().splitlines()
+        entries = _read_transcript_entries(transcript_path)
+        if fallback_human_prompt and not any(_is_human_message(e) for e in entries):
```
ntatsumi, line 491: **The fallback silently does nothing once `human_count_at_start >= 1`.**

The synthetic entry is inserted at index 0, so `human_count` reaches exactly 1 — and the turn gate at line 637 is `if human_count > human_count_at_start`. Verified against the execution-file shape from the new test:

```
human_count_at_start=0 -> 1 LLM spans
human_count_at_start=1 -> 0 LLM spans
human_count_at_start=2 -> 0 LLM spans
```

`human_count_at_start` is `>= 1` whenever the trace was created by the `handle_pre_tool` fallback with >= 2 human entries (`max(0, current_human_count - 1)`, line 1475), by the `handle_stop` continuation branch (line 1769), or if the Agent SDK writes the prompt into the hook JSONL before `UserPromptSubmit` fires — the comment at line 1361 assumes it does not, which is unverified in SDK mode.

In those cases you get zero LLM spans, reproducing exactly the bug this PR fixes, with no log line to show it happened. Resetting `human_count = human_count_at_start` (or forcing `human_count_at_start = 0`) when the fallback entry is used would make it robust.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -1074,6 +1097,7 @@ def _emit_pending_llm_spans(
         current_trace.get("human_count_at_start", 0),
         current_trace["trace_id"],
         current_trace["root_span_id"],
+        current_trace.get("prompt_preview", ""),
```
ntatsumi, line 1100: **`emitted_llm_span_count` and `all_llm_spans` now come from different files.**

`emitted` (line 1103) is accumulated during the run from the **hook JSONL** transcript, but `all_llm_spans` is derived from whatever `transcript_path` is — which `handle_stop` now sets to the **Agent SDK execution file**. `new_spans = all_llm_spans[emitted:]` assumes the two sources agree exactly in group count and order, and nothing keys the counter to the transcript it was computed from.

```
execution file groups=1, emitted_from_jsonl=0 -> 1 new spans exported
execution file groups=1, emitted_from_jsonl=1 -> 0 new spans exported
execution file groups=1, emitted_from_jsonl=2 -> 0 new spans exported
```

If the JSONL produced an extra assistant group the execution file does not have (auto-compact summary, an API-error/retry entry), the slice is empty and the final response is never emitted — the reported symptom returns. If the execution file yields *more* groups, the tail re-exports already-sent responses under fresh span IDs, duplicating LLM spans. Consider recording which transcript the counter was computed against and resetting it to 0 when the source switches.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -1709,7 +1733,13 @@ def handle_stop(data: dict, config: dict) -> None:
             log.debug("No active trace for session %s at stop", session_id)
             return
 
-        transcript_path = _get_cached_transcript_path(data, state, session_id)
+        transcript_path = (
```
ntatsumi, line 1736: **Agent SDK entries carry no `timestamp`, so every span from this file is stamped at finalize-step wall clock.**

`group_ts = entry.get("timestamp", "")` (line 686) yields `""`, and `_iso_to_ns("")` returns `time.time_ns()`. The new test deliberately does `assistant.pop("timestamp")`, so this is the expected shape. Measured:

```
span start within [now-, now+]: True, end - now = 0.50s in the future
```

Combined with the `max()` at line 1157, the CHAIN end lands at `now + output_tokens * 10 ms` — i.e. in the future relative to the tracer process — and the model spans render *after* the TOOL spans they caused, since the finalize step runs at a later step boundary than the tool hooks.

The PR body notes "ordering is not measured model latency", but the effect is a timeline that is actively wrong rather than merely imprecise. The SDK `result` entry carries `duration_ms`, and `system/init` is written at start — anchoring the synthetic spans between those two would at least keep them inside the real window.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -1709,7 +1733,13 @@ def handle_stop(data: dict, config: dict) -> None:
...
+            execution_file
+            if execution_file
+            and Path(execution_file).exists()
+            and _is_real_transcript(execution_file)
+            else _get_cached_transcript_path(data, state, session_id)
+        )
```
ntatsumi, line 1742: **New hazard: the continuation check below now compares counts from two different files, and can emit two root spans for one turn.**

Line 1750 runs `_count_human_messages(transcript_path)` — now possibly the execution file — against `ct["human_count_at_start"]`, which was computed from the hook JSONL. If the execution file contains >= 2 string-content user entries, `_cur_human > _start + 1` trips spuriously: the block completes the current turn, creates a *new* trace, and then falls through to `_complete_turn` again.

Verified end-to-end through `handle_stop`:

```
EXEC 0 human entries-> CHAIN spans=1, LLM spans=1   (the PR's happy path)
EXEC 1 human entry  -> CHAIN spans=1, LLM spans=1
EXEC 2 human entries-> CHAIN spans=2, LLM spans=2
   CHAIN turn=1 out='first'
   CHAIN turn=1 out='second'
```

Two root CHAIN spans, both labelled `turn_number=1`, for a single turn. Whether the Action's execution file ever echoes two string-content user messages I could not confirm, so this is a latent risk rather than a live bug — but the fix is cheap: skip the continuation check entirely when `transcript_path is execution_file`, since the count it is compared against did not come from that file.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -271,6 +271,46 @@ def _session_lock(session_id: str):
...
+        try:
+            entry = json.loads(line)
+        except json.JSONDecodeError:
+            continue
+        if isinstance(entry, dict):
+            entries.append(entry)
+    return entries
+
+
+def _execution_session_id(path: str) -> str:
```
ntatsumi, line 304: `""` is a `str`, so an `init` event that lacks `session_id` returns early with an empty string instead of continuing the scan:

```
_execution_session_id -> ''   (expected 'real')
```
(input: a first `{"type":"system","subtype":"init"}` with no `session_id`, followed by a real one)

`handle_stop` then falls back to `"unknown"`, finds no state, and returns without finalizing anything. Suggest `if isinstance(session_id, str) and session_id:` so falsy values keep scanning.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -377,19 +406,11 @@ def _is_human_message(entry: dict) -> bool:
...
-                continue
-            try:
-                if _is_human_message(json.loads(line)):
-                    count += 1
-            except json.JSONDecodeError:
-                pass
-        return count
-    except Exception:
-        return 0
+    return sum(
```
ntatsumi, line 409: **Regression (confirmed): dropping the `except Exception` wrapper can now abort turn finalization.**

This function and `_get_latest_human_message` (line 1519) both lost their `try/except Exception`, and `_read_transcript_entries` only catches `OSError` around `read_text()`. A `UnicodeDecodeError` is a `ValueError`, not an `OSError`, so it propagates out of `handle_stop` at line 1750 — *before* `_emit_pending_llm_spans`, `_complete_turn` and `_delete_state` run.

Same transcript (one line containing a byte invalid in UTF-8), both branches:

```
PR : raised=UnicodeDecodeError, chain spans exported=0, state file left=True
dev: raised=None,               chain spans exported=1, state file left=False
     WARNING Failed to extract LLM spans: 'utf-8' codec can't decode byte 0xff ...
```

On `dev` these degraded to `0` / `""`, the broad handler inside `_extract_llm_spans_for_turn` absorbed it, and the turn still closed. Now the root CHAIN span is lost and the state file is orphaned until the 48 h sweep. Catching `ValueError` alongside `OSError` in `_read_transcript_entries` (or `errors="replace"` on `read_text`) restores the old behaviour.

integrations/claude-code-observability/claude_code_tracer.py
```diff
@@ -279,21 +319,10 @@ def _is_real_transcript(path: str) -> bool:
...
-                try:
-                    entry = json.loads(line)
-                except json.JSONDecodeError:
-                    continue
-                if entry.get("type") in ("user", "assistant"):
-                    return True
-        return False
-    except OSError:
-        return False
+    return any(
```
ntatsumi, line 322: **Performance regression: this went from a streaming early-return to a full read + parse of the whole transcript.**

`_read_transcript_entries` materialises the entire file and JSON-parses every line before `any()` gets a chance to short-circuit, whereas the old `open()` loop returned at the first `user`/`assistant` line. Measured on a 37 MB transcript:

```
dev: _is_real_transcript -> True in   0.2 ms, peak mem   0.0 MB
PR : _is_real_transcript -> True in 619.3 ms, peak mem 198.3 MB
```

`_find_transcript_path` (line 358) calls this once per candidate, and candidate 3 is a glob over all project dirs sorted largest-first. `handle_user_prompt_submit` clears the cached path each turn (line 1357), so on a long session this is ~0.6 s and ~200 MB inside a blocking hook, per candidate, per turn — to answer a boolean.

A generator would keep the new JSON-array support and restore the short-circuit: have `_read_transcript_entries` `yield` instead of building a list (`_count_human_messages`/`_get_latest_human_message` iterate it once anyway; only `_extract_llm_spans_for_turn` needs `list()` for its `insert(0, ...)`).

.github/workflows/renovate-autofix.yml
```diff
@@ -380,6 +379,27 @@ jobs:
...
+        env:
+          USER: renovate-autofix-pr-${{ steps.pr.outputs.number }}
+          GENAI_ENGINE_API_KEY: ${{ secrets.GENAI_ENGINE_API_KEY }}
+          GENAI_ENGINE_TASK_ID: ${{ vars.GENAI_ENGINE_TASK_ID }}
+          GENAI_ENGINE_TRACE_ENDPOINT: ${{ vars.GENAI_ENGINE_TRACE_ENDPOINT }}
+          CLAUDE_SESSION_ID: ${{ steps.claude.outputs.session_id }}
+          CLAUDE_EXECUTION_FILE: ${{ steps.claude.outputs.execution_file }}
+        run: |
+          T="$RUNNER_TEMP/claude_code_tracer.py"
+          if { [ -n "$CLAUDE_SESSION_ID" ] || [ -f "$CLAUDE_EXECUTION_FILE" ]; } && [ -f "$T" ]; then
```
ntatsumi, line 395: Worth noting that finalization is now entirely dependent on this guard. `[ -f "" ]` is false, so if the action dies before writing either output the step no-ops — and `if: always()` does not survive a hard runner termination (cancellation, timeout, OOM) either.

The result is a trace with TOOL spans, no root CHAIN span, and an orphaned state file until the 48 h `_cleanup_stale_states` sweep — the same failure class this PR fixes. The old `Stop` hook closed the turn on any normal session end regardless of action outputs.

Keeping the `Stop` hook as a *last-resort* closer (and having the finalize step re-open/extend rather than assume it owns the trace) would cover both, but that needs the state not to be deleted at stop. At minimum, an `else echo "::warning::trace not finalized"` would make the silent case visible.

(The wiring itself is correct — I confirmed `session_id` and `execution_file` both exist as outputs on `claude-code-action@v1`.)

.github/workflows/renovate-autofix.yml
```diff
@@ -380,6 +379,27 @@ jobs:
             through the full CI + merge queue (and an independent Claude review) before merging. Never leave
             broken, unreviewed, or unverified edits.
 
+      # Finalize once the Action has written its Agent SDK execution log. A Stop
+      # hook would close and delete the trace state before this file is available.
+      - name: Finalize Arthur Engine trace
+        if: always() && steps.guard.outputs.proceed == 'true'
+        env:
+          USER: renovate-autofix-pr-${{ steps.pr.outputs.number }}
```
ntatsumi, line 387: Minor: this is dead. `_complete_turn` reads `username` from `state.get("username")`, which is only ever populated at session init inside `handle_user_prompt_submit` (line 1322) / `handle_pre_tool` (line 1411) — `handle_stop` never consults the environment. The value already in the state file comes from the `USER` set on the `claude` step at line 193, so this line changes nothing.

PR #2251, opened 2026-09-06
Make ML Engine health port configurable

kc961163:
## Summary

- add an `ML_ENGINE_PORT` setting with the existing `7492` default
- pass the configured port to the ML Engine health server
- cover integer and environment-style string values

## Why

ML Engine currently exposes its health server on a fixed port. Making that listener configurable allows multiple local ML Engine processes and deployment environments to choose an available health port while preserving the existing default behavior.

## Verification

- 9 focused configuration/job-agent tests passed, 1 skipped
- strict mypy: 49 source files, no issues
- black and isort checks passed
- `git diff --check` passed

The existing job-agent suite emits one pre-existing thread warning in `test_handle_process_failure`; it does not fail the suite.

### DIFF
--- a/ml-engine/src/ml_engine/config/config.py
+++ b/ml-engine/src/ml_engine/config/config.py
@@ -54,3 +54,10 @@ class Config:
             settings.GENAI_ENGINE_MAX_PAGE_SIZE,
             "GENAI_ENGINE_MAX_PAGE_SIZE",
         )
+
+    @staticmethod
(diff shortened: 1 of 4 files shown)


kc961163:
Closing this experimental approach after architecture review. The local port mismatch is real, but we will investigate solving Stack-specific isolation in Arthur Stack orchestration while keeping ML Engine's standard runtime port unchanged.

PR #2314, opened 2026-09-18
chore(renovate): turn the heavy label into a defer list

ntatsumi:
## What

Labelling a Renovate PR `heavy` already takes it out of the auto-fixer's scope, but does nothing to Renovate itself — the branch keeps getting rebased, CI keeps running on it, and the next release brings a fresh PR. [#2307](https://github.com/arthur-ai/arthur-engine/pull/2307) (material-ui monorepo v9) is the case that prompted this.

`renovate-defer.yml` now listens for the label and opens a PR adding the update's packages to a defer list in `renovate.json` — one `enabled: false` rule per deferred PR, so each entry names where it came from and deleting the rule is how the update comes back.

The #2307 deferral is included here as the worked example:

```json
{
  "description": "Deferred by PR #2307 (Update material-ui monorepo to v9), labelled heavy: ...",
  "matchPackageNames": ["@mui/icons-material", "@mui/material", "@mui/x-charts", "@mui/x-date-pickers"],
  "enabled": false
}
```

## Two implementation notes

- **Package names come from the manifest diff, not the PR body.** Renovate truncates that table on large groups — #2307 lost most of its rows to the platform limit. Lockfiles are excluded from the scan, since a `yarn.lock` hunk names every transitive package.
- **`jq` alone reflows the whole config.** `renovate.json` is prettier-formatted at width 100, so the script normalizes afterwards; the diff is the added rule and nothing else.

## One change beyond the ask

`vulnerabilityAlerts` gains `"enabled": true`. An `enabled: false` packageRule also silences **security** updates for the matched packages unless that flag is set explicitly ([renovatebot#40800](https://github.com/renovatebot/renovate/discussions/40800), [renovatebot#15490](https://github.com/renovatebot/renovate/discussions/15490)). Without it every deferred package quietly stops receiving CVE fixes — as the existing `python`/`torch`/`numpy` rule already did.

## Verification

- Script run against #2307 → exactly the 4 MUI packages; against #2308 → `sqlalchemy`, `arthur-client` (pep621 path). Re-running either is a no-op.
- `renovate-config-validator` passes on the updated config; re-running prettier is a no-op.
- Workflow YAML parses; both `run` blocks pass `bash -n`.
- Not exercised end-to-end: the workflow only becomes live once merged to `dev`, and `pull_request_target` always runs the default-branch copy. First real run will be the next `heavy` label.


ntatsumi:
All three addressed in 8a1f822 — both findings confirmed against the repo first.

**1. Extras.** Confirmed: `"psycopg[binary]==3.3.5"` (ml-engine) and `"boto3-stubs[s3]>=1.28.0,<2"` (deployment/model-upload) extract nothing today. Took the suggested shape, including `[[]`/`[]]` over a backslash-escaped `[` — CI runs mawk, so the POSIX spelling matters.

**2. Docker/compose.** Confirmed: compose paths fail the file test outright, and the only Dockerfile lines either regex matches are the `custom.regex` `uv==` ones. Extended the extractor rather than narrowing the job `if:` — the workflow can't know the manager before reading the diff, so gating there would put the manager list in a second place that drifts from `renovate.json`.

The branch stayed small by keying on the tag colon instead of enumerating stage-ref shapes: every internal ref in `genai-engine/dockerfile` (`FROM preinstall AS cpu-install`, `FROM ${TORCH_DEVICE}-install AS install`) carries no tag, variable-interpolated tags carry `$`, and image-minus-tag is already Renovate's `depName`. Against a synthetic diff spanning all four manifest kinds plus `yarn.lock`, the shipped awk emits exactly:

```
@mui/material  boto3-stubs  gcr.io/distroless/python3-debian12  node
postgres  psycopg  python  sqlalchemy  uv
```

Registry-qualified name kept whole, stage refs and `${VAR}` tags skipped, lockfile still excluded. Re-ran #2307 and #2308 against the updated extractor: unchanged.

**3. The silent skip — agreed this is the one that mattered.** Any extractor can fall behind what Renovate proposes, so the workflow now comments on both no-op exits. Exit 4 says the defer list was *not* updated and Renovate will keep proposing the update; exit 3 says everything was already deferred. "The label had no effect" is no longer distinguishable only by absence.

One correction for the record: there are two dedicated `matchManagers` rules covering dockerfile/docker-compose (python disabled, postgres `<17`), not three — the third is the reviewer-routing rule listing five managers. Doesn't change the conclusion.

PR #2321, opened 2026-09-19
UP-4979 · D-07: scan one discovery source per job, resiliently

TomLisankie:
[UP-4979](https://arthurai.atlassian.net/browse/UP-4979) · D-07 · ML Engine half of Agentic Discovery v1.

## Scope, after D-06 absorbed the spec extension

Per the [comment on UP-4978](https://arthurai.atlassian.net/browse/UP-4978) and the updated D-07 description, the spec extension and the `lookback_hours` change landed in D-06 ([arthur-scope !1921](https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1921)) — they had to, because without the config ID in the spec the scheduler cannot give each config its own job. What was left for D-07 is the ML Engine side: wire the handler to execute, per-source resiliency, and persist the result as a Discovery Run.

## What this does

D-06 dispatches one `DISCOVER_AGENTS` job per (source config × engine) and materializes the config into the spec. The handler ignored all of that and ran the old GCP sweep for every job, so a source-scoped job scanned the wrong thing.

`DiscoverAgentsExecutor.execute` now branches on whether the job carries a config:

- **Carries one** → scanned against that config alone, over the lookback the config asked for rather than the 720-hour default, with each batch of records published *as it arrives*. A source that throws on its third page keeps the first two; the exception is then re-raised so the job fails and the Platform reads the leg as failed rather than as a partial success.
- **Carries none** → the existing GCP data-plane sweep, untouched. Both shapes stay live until D-14 migrates GCP onto a source config.

A job that names a config but carries no materialized config (or vice versa) is rejected rather than half-scanned — that is a dispatch bug, and guessing the missing half would scan something nobody asked for.

## Acceptance criteria

| Criterion | Where |
|---|---|
| A scan job carries exactly one source config | `_require_source_config`, plus `test_scan_publishes_every_batch_from_its_one_config` |
| A source that throws mid-scan does not discard results already collected | `run_source_scan` publishes inside the loop; `test_mid_scan_failure_keeps_already_published_records` |
| The run records the failure | `DiscoveryScanOutcome`; `test_failing_source_reports_its_contribution_and_its_failure` |
| One source's failure cannot affect a sibling | Structural — two configs are two jobs; `test_a_sources_failure_does_not_touch_a_sibling_config` pins that a failed scan leaves nothing behind |
| Bindings regenerated | `arthur-client` is consumed here as a PyPI pin, not generated into this repo — bumped instead, see below |
| Credentials never reach the job log | `redact_secrets`; `test_neither_kind_of_credential_reaches_the_job_log` |

## Seams, deliberately empty

Making the vendor call and validating output columns against the D-02 contract are **D-13's** (in progress, Nori). Resolving a discovery record onto a task keyed on `external_id` is **D-08's** (To Do). Both sit behind Protocols in `discovery_scan.py`, so:

- `SOURCE_SCANNERS` is empty. A config naming an unregistered vendor fails **its own job** with that reason — correct behaviour for an unsupported vendor either way, and reported per job rather than per engine.
- There is no record sink yet, so nothing publishes in production until D-08/D-13 land. The resiliency semantics are fully exercised by unit tests with fakes.

The seams are intentionally minimal so D-13 can replace them rather than work around them. **Worth a look from @nori** before this lands.

Credentials are *not* a seam, which is a change from the original plan. D-05 has landed, so `DiscoverAgentsExecutor` reads the config's sensitive fields via `retrieve_discovery_source_credentials` at the point of the scan — never in the job spec, per that route's own warning — and passes them to `DiscoverySourceScanner.scan`, whose signature is widened to take them. D-13's scanner therefore receives its credentials rather than fetching them.

## Credentials do not reach the job log

Per [Krishna's review](https://github.com/arthur-ai/arthur-engine/pull/2321#discussion_r4063609421): the outcome travels to the Platform as log *text*, so an exception that stringifies the request it failed on would publish the credential with it.

Redaction is exact wherever it can be. The values D-05 just returned are removed from the message by exact match — no guess about how a vendor formats its errors, and nothing else can be caught by accident. Scrub sets are passed as arguments rather than held on `DiscoveryScanOutcome`, since that object exists to be serialized into the log.

Pattern matching survives only as a backstop for credentials nobody handed us — an OAuth access token minted from a `client_secret`, or a Splunk session key minted from a password, exists only inside the scan and was never in the returned set. Three rules cover that shape: any `Authorization` value whatever the scheme, a bare scheme token, and credential-*named* query parameters.

A broader `key: value` rule was written first and then removed. Measured against shapes it was not designed around it leaked 7 of 10 and over-redacted 4 of 6: everything it caught was already covered exactly, while `token:`, `auth:` and `cookie:` in ordinary error prose made it eat the diagnosis, and an unrecognised scheme (`Authorization: Splunk <key>`) had it redact the scheme word and leave the credential. The tests now probe derived-vs-configured credentials and assert ordinary error prose comes through byte-identical.

Two things this does **not** cover, both worth their own ticket: `connector_test_executor.py` and `fetch_data_executor.py` log raw `str(e)` from credentialed connectors on paths live in production today, and holding sensitive fields as Pydantic `SecretStr` (the `databricks_connector.py` idiom) would cover accidental interpolation for free.

## Discovery Run persistence is blocked

The ticket asks for the job result to be persisted as a Discovery Run. D-11 is To Do — there is no run store, no run routes, and no run models in the bindings. Nothing to persist to.

`records_published` counts every record the sink accepted, created and updated alike — a source reporting the same twenty agents each scan contributes twenty each time. It is not a count of new agents, and the `DiscoveryRecordSink` docstring now says so, since D-08 and D-11 are both unwritten and that is when the ambiguity would get baked in backwards.

Instead the per-source outcome is emitted as JSON in the job log, in the shape D-11 will persist (config id/name, source id, vendor, job id, scan id, lookback, timings, batches, records published, error count, error). It is in the log *message* rather than in `extra` because `ScopeJobLogExporter` ships only `record.getMessage()` to the Platform. **A follow-up is needed once D-11 lands** to write this through the run API instead.

## Dependencies

`arthur-client` 1.4.2604 → 1.4.2669 for the D-06 spec fields (`discovery_source_config_id`, `discovery_source_config`, `scan_id`), and `arthur-common` 2.4.73 → 2.4.79 for `DiscoveryOutputRecord`. The `litellm` 1.98.0 → 1.100.1 change in the lock is arthur-common's transitive pin, not drift. `check-lockfile-drift.sh` passes on all four projects.

## Verification

- `uv run pytest tests/unit` — 317 passed, 1 skipped (the `PytestUnhandledThreadExceptionWarning` in `job_agent` is pre-existing on `dev`; verified by stashing)
- `uv run mypy src/ml_engine` — clean, 51 files
- `./scripts/lint.sh` — clean
- `./.github/scripts/check-lockfile-drift.sh` — all ok

No changelog entry: this is ML Engine, not the GenAI Engine API.

## Summary by CodeRabbit

- **New Features**
  - Added source-specific discovery scans with incremental publishing of discovered records.
  - Added configurable scan lookback periods, including automatic fallback handling.
  - Added structured scan results showing completion status, published records, and failures.

- **Bug Fixes**
  - Preserves successfully published results when later scanning or publishing fails.
  - Invalid configurations, unsupported vendors, and missing publishing destinations now fail clearly with finalized error results.
  - Existing GCP-wide discovery jobs continue to run as before.


ml-engine/src/ml_engine/job_executors/discover_agents_executor.py
```diff
@@ -32,18 +47,164 @@ def __init__(
...
+        job: Job,
+        job_spec: DiscoverAgentsJobSpec,
+    ) -> None:
+        """Scan the single source config this job was dispatched for.
+
+        The config is read from the job rather than fetched by ID: D-06 snapshots it at
+        dispatch so the run stays reproducible and the query that ran is the query
+        recorded, even if the config is edited or deleted afterwards.
+        """
+        config = self._require_source_config(job_spec)
```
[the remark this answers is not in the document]
TomLisankie: Addressing this

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,189 @@
...
+    )
+    finished_at: Optional[datetime] = None
+    batches_published: int = 0
+    records_published: int = 0
+    error_count: int = 0
+    error: Optional[str] = None
+
+    def record_failure(self, exc: BaseException) -> None:
+        self.error_count += 1
+        self.error = f"{type(exc).__name__}: {exc}"
```
kc961163: A provider library could raise an error containing sensitive request details. For example, a hypothetical error containing token=FAKE_SECRET would be copied unchanged here.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,189 @@
...
+            "vendor": self.vendor,
+            "job_id": self.job_id,
+            "scan_id": self.scan_id,
+            "lookback_hours": self.lookback_hours,
+            "started_at": self.started_at.isoformat(),
+            "finished_at": (self.finished_at.isoformat() if self.finished_at else None),
+            "batches_published": self.batches_published,
+            "records_published": self.records_published,
+            "error_count": self.error_count,
+            "error": self.error,
```
kc961163, line 159: Could we make sure the error stored here is safe to include in job logs? A connector exception could contain a token or sensitive request details, and this currently copies its full message into the outcome. Is it sanitized before reaching this method? If not, could we use a safe error message and add a test confirming a fake token is absent from the outcome and emitted logs?

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,189 @@
...
+        self,
+        config: DiscoverySourceConfigSpec,
+        lookback_hours: int,
+    ) -> Iterator[Sequence[DiscoveryOutputRecord]]: ...
+
+
+class DiscoveryRecordSink(Protocol):
+    """Where a batch of records goes once a scanner has produced it.
+
+    Returns the number of records it accepted, which is what a run reports as this
```
kc961163, line 63: For the Run Store integration, can we confirm that records_published counts all accepted records, including updates to existing Tasks, rather than only newly created Tasks? For example, if a scan accepts 20 records and 15 already exist, I would store the contribution as 20 but avoid describing that as 20 new agents. Is that the intended meaning?

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Good point. Will resolve.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Good point. Will resolve.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Yeah. This doesn't require code changes but I'll clarify it in the docstring.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,290 @@
...
+        logger.error(
+            f"Discovery scan of source config '{config.name}' failed after "
+            f"{outcome.records_published} record(s): "
+            f"{redact_secrets(str(e), known_secrets)}",
+            extra={
+                "discovery_source_config_id": outcome.discovery_source_config_id,
+                "vendor": config.vendor,
+                "records_published": outcome.records_published,
+            },
+            exc_info=True,
```
ntatsumi, line 207: **This does not hold — both credentials reach the Platform.**

`exc_info=True` here makes `ScopeJobLogExporter.emit` take its second branch (`job_log_exporter.py:52`), which posts two more payloads that never touch `redact_secrets`:

- `post_job_logs` with `"".join(traceback.format_exception(...))` — whose last line is the raw exception string
- `post_job_errors` with `JobError(error=str(exc_value))` — also raw

The re-raise then reaches `job_executor.py:424`'s `self.logger.error(..., exc_info=e)` and ships it a second time.

Reproduced on this branch, driving the real exporter with a fake `JobsV1Api`, a configured `password` and a session key minted inside the scan:

```
[log 0] ... failed after 0 record(s): 401 for user svc ([redacted]); retried with Authorization: [redacted]
[log 1] RuntimeError: 401 for user svc (sup3rs3cr3t-pw); retried with Authorization: Splunk SESSIONKEY9f8e7d6c5b4a
[err 0] 401 for user svc (sup3rs3cr3t-pw); retried with Authorization: Splunk SESSIONKEY9f8e7d6c5b4a

configured password leaked: True
derived session key leaked: True
```

`test_neither_kind_of_credential_reaches_the_job_log` asserts only on `record.getMessage()`, so it stays green while both values ship.

The comment at line 288 is what made this invisible: "ScopeJobLogExporter ships getMessage() only" is true of the first payload and false of the other two.

Worth fixing at the choke point rather than the call site — a `logging.Filter` installed alongside the handler in `ExportContextedLogger`, or the scrub inside `emit` where both `getMessage()` and `format_exception(...)` are formatted. That closes this path and also covers `connector_test_executor.py` and `fetch_data_executor.py`, which the description already names as leaking raw `str(e)` today. As written, the same bug has to be re-fixed in every executor.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,290 @@
...
+    ),
+    # Credential-named query parameters only, so a failing search still reports the
+    # parameters that explain it (output_mode, count, earliest).
+    re.compile(
+        r"(?P<keep>[?&](?:[\w-]*"
+        r"(?:token|key|sig|signature|secret|password|auth|credential)[\w-]*)=)"
+        r"[^&\s#]+",
+        re.IGNORECASE,
+    ),
+)
```
ntatsumi: **All three surviving rules over-redact on shapes the tests do not probe.** Ran them against this branch:

| input | output |
|---|---|
| `HTTPError: 401 Authorization: Bearer abc12345 while GET https://splunk.example.com/services/search returned 401` | `HTTPError: 401 Authorization: [redacted]` |
| `ConnectionError: splunk enterprise host unreachable` | `ConnectionError: splunk [redacted] host unreachable` |
| `ValueError: Basic configuration missing for connector` | `ValueError: Basic [redacted] missing for connector` |
| `TimeoutError: token refresh_endpoint timed out` | `TimeoutError: token [redacted] timed out` |
| `GET /x?design=modern&assignee=alice&monkey=1&keyword=agents&author=bob` | all five values redacted |

Three separate causes:

**1. Line 116** — the value class `[^,;}\]\n"']+` does not stop at whitespace, so the rule eats the rest of the line rather than the credential. The URL, the operation and the status code all go. Every case in `test_derived_credentials_are_caught_by_the_backstop` happens to put the credential at end-of-string or inside quotes, so none of them can see it — and "an SDK stringifying the request it failed on" is the shape the comment says this rule exists for. Adding `\s` to the excluded set covers it.

**2. Line 122** — `\b(?:bearer|basic|splunk|token)\s+[A-Za-z0-9\-._~+/=]{8,}` matches `<scheme> <any 8+ character word>`. `splunk` is the one vendor this targets, so an ordinary connection error loses its diagnosis. `test_the_backstop_leaves_ordinary_error_prose_alone` only probes the colon form (`auth:`, `token:`, `cookie:`), never `<scheme> <word>` — so the over-redaction the description reports measuring away is live in a different rule than the one that was removed.

**3. Line 128** — `[\w-]*(?:token|key|sig|...)[\w-]*` matches the credential word as a *substring* of the parameter name: `design` and `assignee` contain `sig`, `monkey` and `keyword` contain `key`, `author` contains `auth`. Also `?signature_version=v4`. Anchoring the alternation to whole `_`/`-` separated segments keeps the intent without the collateral.

The comment above the block says the rules "cover that shape and nothing else", which is the claim worth re-measuring — the `key: value` rule was removed on exactly this evidence.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,290 @@
...
+            extra={
+                "discovery_source_config_id": outcome.discovery_source_config_id,
+                "vendor": config.vendor,
+                "records_published": outcome.records_published,
+            },
+            exc_info=True,
+        )
+        raise
+    finally:
+        finalize_outcome(outcome, logger)
```
ntatsumi: **`except Exception` with `finally: finalize_outcome` reports an aborted scan as a success.**

A `BaseException` — `KeyboardInterrupt`/`SystemExit` on job_agent shutdown, or `GeneratorExit` on the scanner — skips `record_failure` but still runs the `finally`. Ran it on this branch with a scanner raising `KeyboardInterrupt` after two pages had published:

```
propagated: KeyboardInterrupt
succeeded   : True
error       : None
error_count : 0
records     : 3
```

A scan that was killed part-way is then indistinguishable in D-11's run store from a source that genuinely had three records — the one thing the outcome record exists to prevent. `record_failure` already takes `BaseException`, so the narrower catch here reads as unintentional.

ml-engine/src/ml_engine/job_executors/discover_agents_executor.py
```diff
@@ -32,18 +48,244 @@ def __init__(
...
+        """The window to scan, in whole hours.
+
+        D-06 already rounds the config's window up into `lookback_hours`; recomputing
+        it from the config is the fallback for a job dispatched before that landed, and
+        keeps the two from disagreeing about which window actually ran.
+        """
+        if job_spec.lookback_hours is not None:
+            return int(job_spec.lookback_hours)
+        # Rounded up, like D-06 does: a 90-minute window scans 2 hours rather than 1.
+        return ceil(int(config.lookback_window_seconds) / 3600)
```
ntatsumi: **This fallback cannot fire, and the test covering it exercises a state the wire format cannot produce.**

`lookback_hours` is `Field(default=720)` in the binding, and `from_dict` is `obj.get("lookback_hours") if obj.get("lookback_hours") is not None else 720`. Against arthur-client 1.4.2681:

```
key omitted      -> 720
explicit null    -> 720
ctor, no kwarg   -> 720
ctor, None kwarg -> None   <- only reachable from Python
```

The spec reaching `execute` is `job.job_spec.actual_instance`, i.e. the `from_dict` path, so `job_spec.lookback_hours is not None` is always true and `config.lookback_window_seconds` is never read. `test_lookback_falls_back_to_the_configs_window_rounded_up` passes only because `_spec()` builds the model through the pydantic constructor with `lookback_hours=None`.

The consequence is the exact case the docstring names: a job dispatched by a path that does not set the field scans 720 hours against the vendor instead of the config's window, silently.

Either key the fallback on something the wire can express — treat `config.lookback_window_seconds` as the source of truth whenever the spec carries a config — or, if D-06 is now the only dispatcher, drop the branch and its test rather than keep one nothing can reach.

ml-engine/src/ml_engine/job_executors/discover_agents_executor.py
```diff
@@ -32,18 +48,244 @@ def __init__(
...
+        not depend on guessing how a vendor SDK formats its errors.
+        """
+        if self.discovery_sources_client is None:
+            self._fail_before_scan(
+                outcome,
+                RuntimeError(
+                    "No discovery sources client is configured; source credentials "
+                    "cannot be read.",
+                ),
+            )
```
ntatsumi, line 179: This guard has no coverage. `test_missing_record_sink_still_reports_an_outcome` is the only test that omits `discovery_sources_client`, and it also sets `record_sink=None`, so `_execute_source_scan` raises at the sink check on line 125 and never reaches `_source_credentials`.

Verified by mutation on this branch: replacing the condition with `if False:` leaves the suite green at 34 passed. In production that path would `AttributeError` on `None.retrieve_discovery_source_credentials` rather than producing the named outcome this block exists to produce.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,290 @@
...
+        workspace_id: str,
+        data_plane_id: str,
+        config: DiscoverySourceConfigSpec,
+        records: Sequence[DiscoveryOutputRecord],
+    ) -> int: ...
+
+
+# Vendor -> scanner, populated by D-13 as connectors land. Keyed on
+# DiscoverySourceVendor values, e.g. "splunk_enterprise".
+SOURCE_SCANNERS: dict[str, DiscoverySourceScanner] = {}
```
ntatsumi: Seam note for D-13, since the description asks for one: this maps a vendor to a scanner **instance**, and `discover_agents_executor.py:61` does `self.scanners = SOURCE_SCANNERS` — the module dict itself, not a copy.

Two consequences once D-13 fills it:

- `job_agent.py` runs low-memory jobs under `ThreadJobRunner`, so two configs for the same vendor are two jobs in one interpreter calling `.scan()` on the same object. Any per-scan state a real scanner holds — session, paging cursor, the credentials it was just handed — is shared between two sources, which is the cross-source isolation the acceptance criteria call structural.
- Because `self.scanners` aliases the global, a later `self.scanners[v] = s` would mutate it for every other executor in the process.

Mapping vendor -> a factory or class, and `dict(SOURCE_SCANNERS)` on assignment, costs nothing while the registry is empty and makes the seam hard to misuse later.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
```diff
@@ -0,0 +1,290 @@
...
+    something first.
+    """
+    known_secrets = secret_values(credentials)
+    try:
+        for batch in scanner.scan(config, lookback_hours, credentials):
+            if not batch:
+                continue
+            accepted = sink.publish(workspace_id, data_plane_id, config, batch)
+            outcome.batches_published += 1
+            outcome.records_published += accepted
```
ntatsumi, line 195: The other half of the resiliency contract is unexercised: `FakeScanner` can raise, but `RecordingSink` always succeeds, so no test covers `sink.publish` throwing — and that is what the longest docstring in the module (the atomicity paragraph on `DiscoveryRecordSink`) is about.

It also matters for the counters. `batches_published` and `records_published` advance only after `publish` returns, so a sink that commits part of a batch and then raises reports zero for work the Platform accepted, and the partially-landed batch is not counted either. The docstring says such an implementation "owes this module an exception carrying the accepted count" — but nothing here reads one, so D-08 cannot satisfy that contract without a change on this side. A test with a raising sink would pin whichever way you want it to go.

ntatsumi (commented):
Went through this against the branch and verified each finding by running it rather than reading — repros are in the comments.

One blocker: the credential redaction does not hold. `exc_info=True` in `run_source_scan` routes the raw traceback and `str(exc_value)` through `ScopeJobLogExporter`'s second branch, which never sees `redact_secrets`, so both a configured field and a derived session key reach the Platform. The test that covers this asserts only on `record.getMessage()` and stays green. Details and a reproduction inline.

Three more that are real but smaller: the surviving redaction patterns over-redact on shapes their tests do not probe (including `splunk <word>`, which is the one vendor this targets); the `lookback_hours` fallback cannot fire because the binding defaults it to 720 on every wire path; and `except Exception` alongside `finally: finalize_outcome` reports a killed scan as `succeeded: true`.

The rest are a coverage gap I confirmed by mutation, and two notes on the seams since you asked for a look at those.

The shape of this — publish inside the loop, one config per job, an outcome record owed even by the failures that never reach the vendor — reads right to me, and the `_unscannable_outcome` reasoning about null-rather-than-placeholder is the correct call for D-11.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Fixed in 333ab237a, at the choke point as you suggested. Redaction moved to `log_redaction.py`. `ExportContextedLogger` now installs a `SecretRedactingFilter` on the job logger, which scrubs the message and pre-formats a scrubbed `exc_text` before any handler runs, so stdout is covered as well. `ScopeJobLogExporter` also scrubs all three payloads it posts, including the traceback and the `str(exc_value)` job error. The executor registers the credentials it fetches with the filter via `register_secrets(logger, ...)`, and executors that register nothing still get the pattern backstop.

`test_neither_kind_of_credential_reaches_the_platform` drives the real exporter with a fake `JobsV1Api`, replays `JobExecutor`'s `exc_info=e` log, and asserts on every posted log and error plus the formatted stdout record. I confirmed it fails with the exporter change reverted. The misleading comment at the old line 288 is corrected.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: All three fixed in 333ab237a, and every input from your table is now a test:

1. **Authorization rule:** the value stops at whitespace. The scheme word is kept only when a credential-shaped value follows it, so `Authorization: Bearer abc12345 while GET …` becomes `Authorization: Bearer [redacted] while GET …`, and a scheme-less `Authorization: abc12345 while …` still redacts the credential rather than the next word.
2. **`<scheme> <value>` rule:** the value has to look like a credential, meaning it contains a digit or is 24+ characters. `splunk enterprise`, `Basic configuration` and `token refresh_endpoint` pass through.
3. **Query-parameter rule:** it matches only when the name *ends* in a whole `_`/`-` credential segment (`access_token`, `X-Amz-Signature`, `api_key`), so `design`, `assignee`, `monkey`, `keyword`, `author` and `signature_version` are left alone.

Every rule also skips its own placeholder, since the filter and the exporter both scrub and a second pass must be a no-op. `test_redacting_twice_changes_nothing` pins that.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Fixed in 333ab237a: the catch is now `except BaseException`, with the reason in the docstring. `test_a_killed_scan_is_not_reported_as_a_success` raises `KeyboardInterrupt` after a published batch and asserts `succeeded: false` with the records still counted.

ml-engine/src/ml_engine/job_executors/discover_agents_executor.py
TomLisankie: Dropped in 333ab237a. D-06 is the dispatcher and always sets the field, so the dispatched `lookback_hours` is the only window read and the config's `lookback_window_seconds` is never consulted. The one remaining check is a `None` guard, needed for typing and reachable only from a Python-built spec. It fails the job with a named reason and an outcome instead of guessing a window, and it's covered by `test_job_carrying_no_lookback_is_rejected_rather_than_guessed`. The fallback test is replaced by one asserting the dispatched window wins over the config's.

ml-engine/src/ml_engine/job_executors/discover_agents_executor.py
TomLisankie: Covered in 333ab237a by `test_missing_credentials_client_still_reports_an_outcome`, which supplies a sink and omits only the client, so it reaches this guard and asserts the named outcome.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Done in 333ab237a. `SOURCE_SCANNERS` now maps vendor → `DiscoveryScannerFactory` (a scanner class registers as-is), each scan builds its own scanner, and the executor holds `dict(...)` of the registry. The reasoning is in a comment on the registry. Two tests cover it: `test_every_scan_gets_its_own_scanner` and `test_an_executor_does_not_alias_the_global_registry`.

ml-engine/src/ml_engine/job_executors/discovery_scan.py
TomLisankie: Went with atomicity as the firm contract in 333ab237a. The sentence about a non-atomic sink owing an exception with the accepted count is gone: the docstring now says a sink that raises has committed nothing, and that a partial commit is a bug in the sink rather than a case this module accounts for. `RaisingSink` plus `test_a_failing_publish_counts_only_the_batches_that_landed` pin it: the earlier batches stay counted, the failed batch counts nothing, and the job fails with the error recorded.

ntatsumi:
Flagging a mess I made, before this confuses anyone.

**The content of this PR is already in `dev`, merged under #2372 with my name on it.** That was not intentional and it was not the right way to land it.

What happened: #2372 (the Jamf endpoint connector) is written against `DiscoverySourceScanner`, so I based my branch on this one to develop against the real protocol. When #2372 was squash-merged, it carried this branch's commits with it — and a squash collapses them into a single commit authored by whoever opened the PR. So:

- `ml-engine/src/ml_engine/job_executors/discovery_scan.py` and the rest of D-07 are in `dev` now
- `git diff origin/dev <this branch> -- ml-engine/src/ml_engine/job_executors/` is **empty** — the content matches exactly
- your commits are not in `dev` individually, and the commit that carries them says "UP-4991 (D-19): the Jamf endpoint connector"

This PR reads `CONFLICTING` for that reason: its merge base predates the squash, so GitHub is diffing against a stale ancestor rather than showing you nothing-to-do.

I also changed one thing in your code while working against it, which you should look at rather than inherit silently: `test_an_executor_does_not_alias_the_global_registry` asserted `"jamf_pro" not in SOURCE_SCANNERS`, using a real vendor name as a stand-in for one nobody registers. A Jamf connector registers it now, so the test used a name that is not a vendor instead — it asserts the copy rather than which connectors ship.

Two findings from review are open against your file and I deliberately did not fix them there, because both are decisions about the seam rather than about my connector:

- `exc_info=True` reaches `ScopeJobLogExporter`, which posts `traceback.format_exception(...)` and `str(exc_value)` unredacted, so `redact_secrets` is bypassed on that path. `test_neither_kind_of_credential_reaches_the_job_log` only checks `record.getMessage()`.
- `DiscoverySourceScanner.scan` takes no logger, so a scanner logs to a module logger and nothing it reports reaches the Platform job log. That undercuts the "reported, not suppressed" behaviour the connectors rely on. A `logger` argument on the protocol would fix it for every connector at once.

Sorry for the mess — say how you want it handled and I will do the work.

ntatsumi:
Closing as absorbed: this PR's content is already in `dev`, merged under #2372. See the comments above for what happened and for the two review findings I am picking up separately.

PR #2359, opened 2026-09-21
UP-4980 (D-08): GenAI Engine task resolution for discovery records

TomLisankie:
[UP-4980](https://arthurai.atlassian.net/browse/UP-4980) · D-08. Where a discovered record becomes a task, and therefore where the identity guarantee for agentic discovery lives.

## What this adds

`POST /api/v2/agent-tasks/resolve` — the ML Engine scan job hands it the records one connector returned, and every record comes back with a task ID.

Resolution mirrors `TraceIngestionService._resolve_task_id`:

1. the record names a task → use it (`task_id` is optional on this input only; a SIEM does not know Arthur's task IDs)
2. `external_id` already maps to a task → use it
3. a service name the sensor observed maps to a task → use it, and key `external_id` to it so the next scan resolves at rung 2
4. nothing matched → mint a task and key `external_id` to it

Tasks minted here are the shape `create_auto_task` already mints for an unregistered OTEL trace — agentic, auto-created, no default rules — plus the sensor that found them. Discovery and OTEL auto-creation are one event seen from two sides; a scan-minted task that looked different would read as a second kind of agent in every downstream view.

**The key space is `service_name_task_mappings`, the table OTEL resolution already reads.** That is the whole mechanism behind "a discovered agent someone later instruments resolves to the same task": its traces arrive under a service name and the row is already there. Nothing reconciles the two afterwards because nothing has to. Two requests racing for the same identity are the one case that needs more than the table: see the review follow-ups below. Tested in both directions.

## Two decisions worth a look

**Two sources reporting one agent** — the open question D-03's upsert key assumes an answer to. They converge only if they agree on `external_id` or their observed service names overlap; otherwise two tasks. That is v1's intent (`external_id` is canonical, no identity resolution across sensors), and corroboration is provenance's to express in D-09 by holding two evidence records against one task. Both branches are tested so the behavior is pinned rather than incidental.

**The input takes only the three discovery categories**, not the whole `AgentCreationSource` union with a validator rejecting the rest. The schema then states what it accepts, and — less obviously — the wide union carries computed fields, so putting it in a request body made FastAPI split it into `-Input`/`-Output`, renaming the component `/api/v2/agent-tasks` has always returned for no change in shape. The narrow union leaves that response alone; the changelog is one additive line. Cost: a fourth discovery category is one more member here rather than arriving free with a repin.

## Acceptance criteria

| | |
|---|---|
| 500 distinct findings → 500 distinct tasks | tested at 500, not two rows |
| matching `external_id` routes, no duplicate | ✓ |
| no match creates a task | ✓ |
| re-running a scan creates nothing | the 500-record scan, re-run |
| later-instrumented agent keeps its task | ✓, and the reverse direction too |
| missing `external_id` fails validation, never routes to unmapped | rejected by the model and at the boundary |
| cluster-level tasks: an HA install shares one | address is not part of the identity |
| two sources: verify and confirm the behavior | above, both branches tested |

## Deliberately not here

Renaming a task from a later scan, or rewriting its creation source. Identity is stable; the per-sensor record of what was seen when is evidence, which D-09 persists. Compound service/`external_id` keying (the ticket's open question) is not built. No merge logic, so the accepted risk stands: unstable `external_id` values churn duplicates.

Also unchanged: the ML Engine side that calls this. D-13 owns the connector and its output contract.

## Review follow-ups

Three findings from CodeRabbit, all taken:

**A lost identity claim no longer strands a task.** `create_mapping` returns the mapping
that won on a conflict rather than raising; rung 4 discarded that return, so a request
that lost the race told its caller a task ID no key pointed at and left that task in the
inventory forever. It now claims `external_id` before any service name, yields to the
winner when it loses, deletes the task it minted on the way, and reports `EXTERNAL_ID`
rather than `CREATED` — the rung that actually answered. The same race exists in
`TraceIngestionService` step 6, which discards the return the same way; not touched here.

**Whitespace-only `external_id` and `name` are refused.** `min_length=1` let a space
through, and a space is not an identity: two agents whose sources both reported one would
collapse onto a single task, through the backstop that exists to prevent exactly that.
Values are rejected, not stripped — rewriting a key would make identity depend on this
engine's idea of trailing space.

**`created_count` is gone.** It was a plain `@property`, so it never reached the JSON or
the schema, and the route test recomputed it client-side from `resolved` and asserted on
its own arithmetic. Deleted rather than promoted to `@computed_field`: it had no callers,
and every client can derive it from the list it was handed. Those assertions now check
`resolved_by` over the wire, which is what the re-scan case is actually about.

Three new tests (blank values, whitespace preserved, the lost race). `uv run
generate_changelog` reports no schema change, so the changelog entry is unchanged.

## Testing

20 new tests; full `unit_tests` suite green (2033 passed). `routes_security_check` and the route-scope fuzz test pass — the endpoint is admin-only, like OTEL auto-creation, so tasks land in the default org and there is no tenant scope to enforce against.

## Summary by CodeRabbit

- **New Features**
  - Added an API endpoint to resolve discovered agents to existing tasks or create new tasks when no match is found.
  - Resolution can use explicit task IDs, external IDs, or service names, and supports batch requests.
  - Access is available to organization and task administrators with the required permission.
- **API Updates**
  - Discovery records now require a `last_seen` timestamp. Source `instance` and `resource_id` values must be non-empty.
- **Bug Fixes**
  - Concurrent identity claims resolve to the winning task instead of leaving an unnecessary duplicate.


genai-engine/src/services/task/discovery_task_resolution_service.py
```diff
@@ -0,0 +1,290 @@
...
+        rest: the service names would otherwise be keyed to a task rung 4 is about to
+        delete, and the request that won the identity maps them to the winner anyway.
+
+        Returns:
+            The task that owns `record.external_id` once the writes are done, which is
+            `task_id` unless another request claimed the identity first. Only rung 4
+            acts on this: the lower rungs resolved to a task that already existed, so
+            losing the race leaves nothing behind to clean up.
+        """
+        owner_task_id = self._claim_key(record.external_id, task_id, known_mappings)
```
ntatsumi: **`external_id` gets reported back as a service name.**

This writes `external_id` into `service_name_task_mappings` — but that table is also the *reverse* lookup. `get_service_names_by_task_id` returns every row for a task, `_enrich_tasks_with_service_names` puts them on `Task.service_names`, and `_get_task_creation_source` then overwrites `observations.service_names` with that list. So `GET /api/v2/agent-tasks` hands clients a Splunk record ID as a service name the agent emits telemetry under.

Ran it on this branch — one record, no observed service names:

```
external_id:                 3a117138-splunk-1
observations.service_names:  ['3a117138-splunk-1']
```

It collides the other way too: `TraceIngestionService._resolve_task_id` reads the same keys, so a trace whose `service.name` happens to equal some record's `external_id` resolves to that task.

Using this table as the key space is fine — the rows just need to be tellable apart. A prefix on identity keys (`external_id:<id>`), a `key_kind` column, or filtering the reverse lookup would each do it.

genai-engine/src/services/task/discovery_task_resolution_service.py
```diff
@@ -0,0 +1,290 @@
...
+        self,
+        record: DiscoveredAgentRecord,
+        known_mappings: dict[str, str],
+        org_id: Optional[UUID],
+    ) -> ResolvedAgentTask:
+        """Walk one record down the ladder. See the class docstring for the rungs."""
+        # Rung 1: the caller already knows the task.
+        if record.task_id:
+            task = self._require_task(record.task_id, record.external_id)
+            self._map_keys_to_task(record, task.id, known_mappings)
```
ntatsumi: **An explicit `task_id` is silently ignored once the identity is already mapped.**

`_map_keys_to_task` returns early when `external_id` already points at a different task, and rung 1 drops that return value. The response then says the record went to the requested task while the table still points at the old one. No race needed — this is any re-scan where the caller tries to re-route a known agent.

Ran it on this branch:

```
first scan minted      T1
requested              T2
response               T2   (resolved_by=explicit_task_id)
mapping[external_id]   T1
record's new svc name  never mapped at all
next scan resolves to  T1
```

So findings split across two tasks, and the agent flips between them scan to scan depending on whether the caller sends `task_id`. Either honour the claim or return the task that actually owns the identity — but the response shouldn't disagree with the table. Worth a test: the existing one uses a fresh `external_id`, which is the case that works.

genai-engine/src/services/task/discovery_task_resolution_service.py
```diff
@@ -0,0 +1,290 @@
...
+        # emitted SQL is the same run to run, which makes the query log readable.
+        lookup_keys = dict.fromkeys(
+            key
+            for record in records
+            for key in (record.external_id, *record.service_names)
+        )
+        known_mappings = self.mapping_repo.get_task_ids_by_service_names(lookup_keys)
+
+        return [
+            self._resolve_record(record, known_mappings, org_id) for record in records
```
ntatsumi: **One bad `task_id` kills the whole batch, and the retry can't recover it.**

`_require_task`'s 404 propagates out of this comprehension, so a batch of up to 1000 records stops at the bad record. The records before it are already committed (each mint and mapping commits on its own), but the caller gets a 404 with no body saying which ones landed — and re-submitting fails identically, so the scan job never gets a 200 for that batch.

Ran it on this branch, 3 records with an unknown `task_id` on the middle one:

```
raised: 404 Discovered record '...-bad' resolved to task '...', which does not exist
mapping[rec 1] -> <task>   # committed
mapping[rec 2] -> None
mapping[rec 3] -> None     # never processed
retry          -> 404
```

The docstring's "a batch that fails partway therefore leaves the records it already resolved durably resolved, which is what makes the retry safe" holds for a transient failure, but not for bad input. Reporting per-record failure in the response would keep the rest of the scan moving.

ntatsumi (commented):
Reviewed the resolution ladder and the key space. Three things I'd fix before merge — all verified by running them against this branch, details inline. The first two are silent: they produce wrong data rather than errors, and nothing in the current tests catches either.

genai-engine/src/services/task/discovery_task_resolution_service.py
TomLisankie: Fixed in b1e985446. I went with a `key_kind` column (`service_name` | `external_id`) and made it part of the primary key, so one string can be both kinds without either answering for the other. Existing rows backfill to `service_name`.

Every repository lookup is now scoped to a kind (defaulting to `service_name`), so `TraceIngestionService._resolve_task_id` never sees external-ID rows. `get_service_names_by_task_id` only returns service names, so `observations.service_names` no longer carries the record ID. I added tests for both directions: the reverse lookup, and a trace whose `service.name` equals an `external_id`. I also ran the migration up, down and up on Postgres with data in the table.

genai-engine/src/services/task/discovery_task_resolution_service.py
TomLisankie: Fixed in b15b08f5f. I went with returning the task that owns the identity: mappings are immutable, so the claim can't move it. When `external_id` already maps elsewhere, rung 1 now falls through to rung 2. The record resolves to the owner (`resolved_by=external_id`), and its new service names get mapped there too, so the response agrees with the table.

I added the test you suggested with an already-mapped `external_id` (fails on the old code): requested T2, response T1, mapping T1, new service name T1, next scan T1.

genai-engine/src/services/task/discovery_task_resolution_service.py
TomLisankie: Fixed in bc399d170. Every explicit `task_id` in the batch is checked in one query before anything is written. A record that fails the check goes into a new `failed` list on the response (`external_id`, `task_id`, `reason: task_not_found`, `detail`), and the rest of the batch resolves with a 200. `resolved` keeps its meaning: every entry is a success with a real `task_id`.

A task deleted after that check, which the upfront query can't see, is caught from the mapping's FK violation and reported the same way instead of 500ing. Your 3-record repro now gives records 1 and 3 resolved, record 2 in `failed`, and the same answer on retry. I also updated the docstring's retry-safety claim, and regenerated OpenAPI and the changelog.

genai-engine/src/services/task/discovery_task_resolution_service.py
```diff
@@ -0,0 +1,432 @@
...
+                (MappingKeyKind.SERVICE_NAME, service_name),
+            )
+            if not mapped_task_id:
+                continue
+            task = self._require_task(mapped_task_id)
+            logger.info(
+                f"Discovered record '{record.external_id}' matched task "
+                f"'{task.name}' ({task.id}) via service name '{service_name}'",
+            )
+            self._map_keys_to_task(record, task.id, known_mappings)
```
kc961163: Could we use the owner returned by _map_keys_to_task in this branch too? For example, we can find Task A through the service name, but another request can claim the external ID for Task B before our mapping write. This call then returns B, while the response still says A; retrying resolves to B. Could we return the winning owner and add a regression test for this service-name path?

genai-engine/alembic/versions/2026_09_23_1200-91bac5fc80b6_add_key_kind_to_service_name_task_mappings.py
```diff
@@ -0,0 +1,54 @@
...
+PRIMARY_KEY = "service_name_task_mappings_pkey"
+
+
+def upgrade() -> None:
+    op.add_column(
+        TABLE,
+        sa.Column(
+            "key_kind",
+            sa.String(),
+            nullable=False,
```
fryz: since this isn't nullable, don't you need to also define what the default value is and update the existing records w/ that new default value?

fryz: ah - see the note above - all good

genai-engine/src/schemas/agent_discovery_schemas.py
```diff
@@ -16,3 +22,86 @@ class DiscoverAndPollResponse(BaseModel):
...
+    the scan job can tell which records landed without re-submitting them.
+    """
+
+    resolved: list[ResolvedAgentTask] = Field(
+        description="Records that resolved to a task, in request order",
+    )
+    failed: list[FailedDiscoveredRecord] = Field(
+        default_factory=list,
+        description="Records that could not be resolved, in request order. "
+        "Re-submitting one unchanged fails the same way.",
```
kc961163: How should the ML Engine sink handle a response containing both resolved and failed? For example, if A and C resolve but B has an invalid task ID, we should preserve the two accepted records and also report B’s failure. The current sink contract only returns a count and says a raised batch committed nothing. Could we agree where that contract is updated so the worker and Run Store preserve both outcomes?

genai-engine/src/services/task/discovery_task_resolution_service.py
TomLisankie: Good catch — fixed in 1000f9f06. Rung 3 now checks the owner `_map_keys_to_task` returns and, if another request claimed the `external_id` first, resolves to that owner with `resolved_by=EXTERNAL_ID` (nothing is discarded, since the matched task predates the call). The owner lookup is shared with rung 4 via a new `_resolve_to_owner` helper. Added `test_losing_the_identity_claim_after_a_service_name_match_yields_to_the_winner`, which fails before the fix and checks the response names the winner, a retry resolves to the same task, and the service name stays with the task that already owned it.

genai-engine/src/schemas/agent_discovery_schemas.py
TomLisankie: Agreed, and it went further than `failed`: the "a raised batch committed nothing" rule was already untrue once each mint commits on its own. Since this PR is what changed the endpoint's guarantees, I updated the sink contract here in 88f6a0501 rather than leaving the ML Engine docstring promising something the endpoint doesn't do:

- `DiscoveryRecordSink.publish` now returns `DiscoveryPublishResult(accepted, failed)`, where each `FailedDiscoveryRecord` carries `external_id`, `reason`, `detail`. It's declared in ML Engine rather than taken from the generated client, so the persisted shape doesn't move when `genai_client` is regenerated.
- `DiscoveryScanOutcome` accumulates `failed_records` across batches and emits `records_failed` + `failed_records` in its payload. So for your example, A and C count toward `records_published` and B is reported with its reason.
- Record failures don't fail the scan (`succeeded` stays true). They're final for that input (re-submitting B unchanged fails the same way), so nothing retries them; they're for whoever configured the source to fix.
- The atomicity rule is replaced with what the endpoint actually guarantees: a sink that *raises* may have resolved part of the batch, but re-publishing is safe because resolved records come back via rung 2. The run counts none of a raising batch, so its contribution is a lower bound.

What's left for the other tickets: the HTTP sink that maps the resolve response into `DiscoveryPublishResult` is D-13's handoff, and the Discovery Run model (D-11) will need a field for `failed_records`. Does that shape work for the Run Store on your side?

kc961163: That shape works as the input for the Run Store. I’ll account for record-level failures separately from execution failure, and preserve the distinction between confirmed counts and potentially incomplete counts after an interrupted request. The HTTP sink integration remains with the connector-framework work.

kc961163 (dismissed):

genai-engine/src/services/task/discovery_task_resolution_service.py
```diff
@@ -0,0 +1,461 @@
...
+                f"'{task.id}', but its identity already maps to task "
+                f"'{identity_owner_id}'; resolving to the owner",
+            )
+
+        # Rung 2: a previous scan already minted a task for this identity.
+        existing_task_id = known_mappings.get(
+            (MappingKeyKind.EXTERNAL_ID, record.external_id),
+        )
+        if existing_task_id:
+            task = self._require_task(existing_task_id)
```
ntatsumi: **The bulk lookup's saving is given straight back here — the steady-state re-scan still costs one query per record.**

`get_task_ids_by_service_names` says a 500-record re-scan that changes nothing "costs a single round trip rather than 500". Every record that resolves at rung 2 or 3 then does its own `get_db_task_by_id`, so it costs 501.

Measured on this branch, counting statements at the DBAPI cursor:

```
STEADY-STATE RE-SCAN (500 known records, nothing changed)
  501 statements, 0 commits
      500  SELECT tasks                     <- one per record
        1  SELECT service_name_task_mappings

FIRST SCAN (500 new records)
  2001 statements, 1000 commits
      501  SELECT service_name_task_mappings
      500  INSERT tasks
      500  SELECT tasks
      500  INSERT service_name_task_mappings
```

Two separate things in that second block:

1. **The N+1.** `_load_explicit_tasks` already does exactly the right thing for rung 1 — one `query_tasks(ids=...)` for the whole batch. The same call would cover rungs 2 and 3: the task IDs are all sitting in `known_mappings` once it's seeded, so they can be fetched in one query before the loop instead of one at a time inside it.

2. **A commit per write.** `create_mapping` commits each row on its own, so a first scan of the maximum 1000-record batch is ~2000 commits in one request. The per-mint commit is deliberate and I'd leave it — it's what makes a partial batch durable — but the mapping writes for one record could share the mint's transaction rather than taking two more of their own.

Worth doing because this endpoint takes up to 1000 records and `permission_checker`'s wrapper is `async def`, so FastAPI treats the sync handler as a coroutine and runs it on the event loop rather than the threadpool — the whole process is blocked for the duration. That part is pre-existing across this router, not this PR's doing, but this is the endpoint that makes it expensive.

Not a correctness problem, and everything else I raised is fixed — happy for this to be a follow-up if you'd rather not grow the PR.

ntatsumi (changes requested):
One performance item left. The three correctness findings from my earlier review are all fixed and I re-ran the probes against this head to confirm — external IDs no longer surface as service names, an explicit `task_id` can no longer disagree with the table, and a bad `task_id` no longer kills its batch.

genai-engine/src/services/task/discovery_task_resolution_service.py
ntatsumi: Pushed this as `f1215d7eb` rather than leave it with you — shout if you'd rather it came out and went in a follow-up.

The tasks the mapping view names are now read alongside the view itself, folded into the query that was already fetching the tasks records name outright, so it's one query for both instead of two plus one per record. `_require_task` serves from that view and falls back to a query only for a task the view could not have held — one minted earlier in the same batch, or one a concurrent request claimed an identity for — so a task deleted mid-batch still fails its own record instead of the batch.

Same measurement as above, after:

```
STEADY-STATE RE-SCAN (500 known records, nothing changed)
  3 statements, 0 commits      (was 501)
```

There's a regression test with it, `test_a_rescan_costs_the_same_for_fifty_records_as_for_five_hundred`, written as two batch sizes costing the same rather than a fixed number so it doesn't break every time resolution does more or less work per call. It fails on the previous commit with `assert 501 == 51`.

I left the commit-per-write alone deliberately — batching the mapping writes into the mint's transaction would undo the property that a partially failed batch stays durably resolved, which is what makes the retry safe and what `failed` depends on.

Full genai-engine unit suite passes (2041), and isort/autoflake/black/mypy are clean.

PR #2380, opened 2026-09-23
UP-4991 (D-19): the payload identifies itself, so its attribute name is not an input

ntatsumi:
## What this fixes, and how it was found

Run against a real Jamf tenant for the first time. The connector's default attribute name was `"AI Inventory"`; the deployed fleet reports under **`"Arthur AI Inventory"`**.

`SOURCE_SCANNERS` holds `JamfScanner` as a zero-arg factory, so that default is what runs in production. Every device would have resolved to `NEVER_REPORTED`, `records_for` would have returned `None` for all of them, and the scan would have published **nothing** while logging "19 unreadable" — a fleet demonstrably covered in AI agents reported as empty. Plausible output, exit code zero, no error anywhere.

## The fix is not a config field

The obvious repair — make the name configurable — was wrong. **The name was never needed.**

`arthur1.` is a magic prefix. Its entire purpose is to let a reader recognise the value without being told where to look. Requiring the display name as well means the format identifies itself and we ask an admin anyway. The attribute is now found by that prefix, and by `ERROR:oversize:`, which is unframed for exactly the same reason.

So: no field added to the vendor schema in arthur-scope, nothing for an admin to get wrong, and a fleet that renames its attributes keeps working.

## What the live data ruled out

`no-cache` cannot identify the payload. On the reference fleet one device carries that sentinel on **both** the payload and the status attribute:

```
Arthur AI Inventory          arthur1.×9, <empty>×9, no-cache×1
Arthur AI Inventory Status   <empty>×9, 'ok …'×6, 'degraded …'×3, no-cache×1
```

An unpopulated attribute carries nothing to match at all. Both are read as a *reason of last resort* rather than as the payload — which keeps a deployment fault distinguishable from a Mac that has never reported. Neither changes what a scan publishes: every non-`OK` outcome was already "a device we cannot speak for".

Two framed values are two collectors, or one attribute duplicated. Reading the first is a guess about which is current, so it is logged rather than picked quietly.

## Verification against the tenant

19 devices, page-size forced to 2 so both cursors had to advance across ~10 pages:

- Token path, the five `section` params, `page-size` and `sort` — all accepted.
- `extensionAttributes` arrive in **two** sections (`<root>` and `general`), confirming the merge in `_to_device` was necessary, not defensive.
- Keyset on `id`: 19 read across 10 pages, zero duplicates.
- RSQL `general.reportDate=gt="…"` with the `,`/`;` cursor and the two-key sort — accepted, cursor advanced across pages carrying real millisecond timestamps.
- End to end with **no attribute name configured**: 9/9 payloads decoded, 48 findings — Claude Code, Claude Desktop, Claude in Chrome, ChatGPT Desktop, Codex CLI, Cursor, OpenClaw, Perplexity Comet, OpenHands, SWE-agent, aider, goose.

## Tests

The fixture now names the attribute something no runbook suggests, so **a test that passes is a test that did not depend on the name** — the whole existing suite went green under the rename, which is the actual proof. Six new tests cover the cases the live data exposed: any attribute name, the status attribute never mistaken for the payload, `no-cache` on two attributes, all-empty attributes, two framed values, and oversize found by prefix.

`tests/unit` 479 passed / 1 skipped · `mypy src/ml_engine` clean · `integrations/endpoint-ai-discovery/test/lint.sh` PASS.

## Docs

The runbook implied the name mattered and spelled the three attributes inconsistently, prefixing only `Arthur AI Osquery Prereq`. They now follow the reference deployment and say plainly that the name is the admin's to choose.

Jira: [UP-4991](https://arthurai.atlassian.net/browse/UP-4991)

## Summary by CodeRabbit

* **New Features**
  * Jamf inventory discovery now recognizes payloads by their content marker, regardless of the Extension Attribute’s display name.
  * Discovery handles multiple matching payloads, detects conflicting values, and distinguishes no-cache data from missing or unusable inventory.
  * Added a read-only check for Jamf tenant connectivity, inventory access, paging, and payload decoding.
* **Bug Fixes**
  * Scans now warn when devices are read but no inventory payloads can be decoded.
* **Documentation**
  * Updated Jamf setup and architecture guides with the “Arthur AI Inventory” and “Arthur AI Inventory Status” names, clarified how attributes are identified, and added guidance for finding Macs with stale inventory updates.


ntatsumi:
Addressed — and you were right that taking the first framed value was a guess resolved by dictionary order.

The rule is narrower than "reject when there is more than one candidate", because one case is benign: two attributes running the same script read the same file and carry the **same bytes**, so identical candidates are one payload seen twice. Those are accepted.

Distinct candidates are refused outright, including a framed payload beside an `ERROR:oversize:` marker — the marker carries neither content nor a date, so preferring the payload would still be a guess about which scan is current. The device reads as one we cannot speak for, which is what `None` already means in this module, and the warning names the competing attributes.

Three tests cover it: disagreeing payloads, framed-beside-oversize, and identical duplicates.

Two other things added in the same push, both argued for by the live tenant run:

- A scan that read devices and decoded **none** of them now says so. Publishing nothing otherwise makes "no agents anywhere" and "collection is broken" identical in the only output anyone looks at. Warning rather than exception — a fleet whose collector was deployed an hour ago is legitimately in that state.
- `ml-engine/scripts/jamf_smoke.py`, the read-only tool that found this PR's original defect against a real tenant.

`tests/unit` 467 passed / 1 skipped, mypy clean, and the smoke test re-run green against the live tenant.

ntatsumi:
Worked through all four reviews. Four fixed, two declined, one already fixed in an earlier push.

## Fixed

**`absent` was overclaimed in both docs.** You're right, and this is the caveat I'd already identified and then failed to carry into the prose. `architecture.md` said "no container runtime on this Mac at all"; `jamf-pro.md` said "No Docker socket", which predates the other-runtime probe entirely. The scan probes for the runtimes it knows, so `absent` means none of *those* were found. Both now say that — and it's stated as the same bound every branch has (`apps=ok` reports what matched where it globbed) rather than as a weakness.

**The summary sentence swept in `ok`.** "Only `absent` says nothing is wrong. Everything else in the table is a branch that could not look" — `ok` is in that table. Now: `ok` and `absent` are the two that say nothing is wrong.

**The blank-attribute guidance contradicted the paragraph beneath it.** That one is mine and it's the worst of the four: the pre-existing text says find blanks with operator `is` and an empty value; the paragraph I added above it said no criterion on the attribute can see them. Two procedures for one case in one section. My claim was about Jamf's matching engine and I never tested it, so it's gone. What's left is one attribute-level procedure and `Last Inventory Update` for the fleet-level question — which is also the field the collector reads as `general.reportDate`, so scoping on it matches what findings are judged by.

**`jamf_smoke.py`, both points.**
- https was enforced only via the connector, which **step 0 does not use** — it posts the token request directly, so `client_secret` would have gone out in cleartext on a URL the script never checked. Now checked before anything sends it; `http://` exits 2.
- The two cursor assertions failed whenever a walk fit in one page, which on a 19-device tenant is every run at the production page size. An assertion that fails where the code is right is worse than none. They now SKIP with the flag to pass, and the default page size is small so the default run genuinely exercises the cursor. Verified live both ways.

## Declined

**`vendor/osquery-ai-discovery/bin/container-scan:566` and `bin/discover:308`.** Both are edits to the vendored tree, which this repo forbids — `integrations/endpoint-ai-discovery/CLAUDE.md`: *"the only osquery implementation. Do not edit — run tools/vendor-queries.sh"*. A local edit there is overwritten by the next bump and, worse, makes the vendored copy stop being the reference it exists to be. A query or marker change is an upstream change.

On the substance, both were decided upstream rather than skipped:
- `discover:308` — "exclude `absent` only when Docker is not installed" is exactly what `unreadable:<runtime>` now provides. A runtime that is installed and unread reports that instead of `absent`, and it degrades. arthur-ai/osquery-ai-discovery#66.
- `container-scan:566` — when the Docker socket is healthy the scan succeeded, so `ok` is the honest marker; emitting `unreadable:` there would report a successful scan as a failed one. The other-runtime probe belongs on the no-socket path, which is where #65 put it.

## Already fixed

`_inventory_value` returning the first framed value — addressed in 921eac3, which refuses distinct candidates outright including framed-beside-oversize. Replied above.

`tests/unit` 470 passed / 1 skipped, mypy clean, `lint.sh` PASS, and the live tenant re-verified on default flags.

PR #2381, opened 2026-09-23
send engine version header to platform for datasets backwards comp

notthattal:
## Description

Adds the ml-engine version to the job agent and executor so the platform can decipher engine versions and make the dataset consolidation work backwards compatible with any version of the engine

## Jira Ticket

-  https://arthurai.atlassian.net/browse/UP-5022

## Partner MR

- https://gitlab.com/ArthurAI/arthur-scope/-/merge_requests/1926

## Summary by CodeRabbit

* **Improvements**
  * Job-related API requests now include the installed engine version, or an “unknown” value when version information is unavailable.
  * This version information is included consistently in requests used to create jobs and users, and in requests used during job execution.
  * Existing session handling, SSL verification settings, and API client behavior remain unchanged.


ntatsumi:
Took a close look at this one. There's one blocking issue, and the fix is a mechanism we already have in the repo.

## The header ships as `unknown` from every real deployment

`importlib.metadata.version("ml-engine")` only resolves if the `ml-engine` **distribution** is installed, and the image never installs it:

- `ml-engine/Dockerfile:77-81` exports dependencies with `--no-emit-project` and installs only that requirements file.
- `ml-engine/Dockerfile:84` then just `COPY`s the source to `/app/src/ml_engine`, and the entrypoint runs it straight off `PYTHONPATH` (`Dockerfile:235,243`).

So there's no `ml_engine-*.dist-info` on any `sys.path` entry, `PackageNotFoundError` fires on every call, and every containerized engine announces `X-Arthur-Engine-Version: unknown`. It resolves only on dev machines and in CI, where `uv sync` happens to install the project editable (`run-ml-engine-unit-tests` → `setup-uv` → `uv sync --directory ml-engine --frozen`, workflow lines 486-489) — which is also why `test_engine_version_is_resolved` passes. Green test, broken artifact, which is the combination that hurts most.

## Suggested fix: the mechanism behind the health endpoint

genai-engine already solved exactly this, and its solution survives into a distroless image:

- `genai-engine/src/routers/health_routes.py:17` → `utils.get_genai_engine_version()`
- `genai-engine/src/utils/utils.py:255` reads a plain `version` file and memoizes it in a module global
- `genai-engine/Dockerfile:131` — `COPY version /app/src/` puts the file in the image
- `.github/workflows/version-workflow.yml:157-158,181` writes and commits it on every bump

For ml-engine this is nearly free, and needs **no Dockerfile change** if the file lives inside the package:

1. Commit `ml-engine/src/ml_engine/version`. `Dockerfile:84`'s `COPY src/ml_engine /app/src/ml_engine` already carries it, and `Dockerfile:214` copies `/app/` wholesale into the final stage.
2. Add one `echo $NEW_ARTHUR_ENGINE_VERSION > ml-engine/src/ml_engine/version` next to the existing genai-engine ones in `version-workflow.yml` (~line 157) and add the path to the `git add` on line 181. That same commit already seds `ml-engine/pyproject.toml` on line 159, so the two stay in lockstep.
3. Read it with `Path(__file__).parent / "version"` and memoize. That path is identical under an editable dev install and in the image, so the test stops being environment-dependent and starts asserting something true of the shipped artifact.

## While you're in there

**Don't send `unknown`** (`engine_version.py:17`). The comment says the control plane treats an *absent* header as legacy, but the fallback sends a present, unparseable value — a third state the contract doesn't define. If the other side does a semver compare it throws or takes an undefined branch, and per the above that's the only path any real deployment takes. Skip `set_default_header` entirely when the version can't be resolved so the engine lands in the defined legacy bucket. (genai-engine's reader raises instead of falling back, which is also defensible — an engine that can't identify itself is a packaging bug, not a runtime condition.)

**Nothing is logged** (`engine_version.py:16`). The `except` swallows silently, so the packaging failure above would be invisible in engine logs; the only evidence would be in control-plane request headers. A `logger.warning` on that path would have caught it on the first deploy.

**The header name is a private literal on one side of a two-repo contract** (`engine_version.py:10`). The consumer lives in arthur-scope (MR 1926) — grepping that checkout for `Engine-Version` finds nothing, so the two spellings aren't verifiable from this repo, and a rename or typo on either side silently reverts every engine to "legacy" with no test failure and no error. We already have the pattern for this: `discovery_record_sink.py:61-63` puts `MAX_DISCOVERED_RECORDS_PER_REQUEST` in `arthur_common` "so both ends read one number". Same argument, and `arthur_common` is a dependency of both sides.

**The wrapper is applied at two call sites rather than in one factory.** `job_agent.py:49-69` and `job_executor.py:119-138` are the same ~20-line block (ssl_verify lookup → session → configuration → `ApiClient`), now wrapped identically. A platform client built in a third module later gets no compile-time or review-time reminder, and the failure mode is silent — the control plane just narrows what it returns. A `build_platform_api_client()` in `tools/` that both call would make the header unforgettable and delete the duplicated block.

**Two smaller ones.** `engine_version()` runs on every `JobExecutor()` construction, i.e. once per job process (`job_runner.py:82-84`), and in the container that's the full failing metadata scan every time — memoizing (as genai-engine does) makes it once. And `ApiClient()` with no configuration in the test caches a process-global `Configuration._default` for the rest of the pytest session and leaks an unclosed `PoolManager`; passing an explicit `Configuration()` keeps it hermetic.

One design question worth a sentence in the description: the platform already models this kind of fact as data-plane capabilities (`patch_data_plane` with `PatchDataPlaneCapabilities`, and `DataPlaneCapabilities` already carries `gen_ai_enabled`), and `JobExecutor` already builds `self.data_planes_client`. Declaring the version once at agent startup through that endpoint would put it in the OpenAPI contract and make it introspectable outside a live request, instead of riding along on every call invisibly to the generated clients on both sides. Since the control plane is changing anyway in the partner MR, it may be the same cost — but if there's a reason the header is the better fit here, worth writing down.

